PR #2, opened 2020-04-14
Add service create, delete, bind and unbind functions

grs:


client/van_serviceinterface_update.go
```diff
@@ -0,0 +1,355 @@
...
+			_, err = cli.KubeClient.CoreV1().ConfigMaps(cli.Namespace).Update(current)
+			if err != nil {
+				return fmt.Errorf("Failed to update skupper-services config map: %s", err)
+			} else {
+				return nil
+			}
+		} else {
+			return fmt.Errorf("Service %s already defined", service.Address)
+		}
+	} else if errors.IsNotFound(err) {
```
ajssmith, line 123: Would it work to assume skupper-services cm is create during van_router_create and treat this as exceptional?

client/van_serviceinterface_update.go
```diff
@@ -0,0 +1,355 @@
...
+			return err
+		}
+	} else if targetType == "pods" {
+		return fmt.Errorf("Target type for service interface not yet implemented")
+	} else {
+		return fmt.Errorf("Unsupported target type for service itnerface", targetType)
+	}
+}
+
+func (cli *VanClient) VanServiceInterfaceExpose(ctx context.Context, targetType string, targetName string, options types.VanServiceInterfaceCreateOptions) error {
```
ajssmith, line 323: Would prefer not to have two ways to accomplish the same operation in the interface. Is there a functional difference between expose/unexpose and bind/unbind? Could we keep expose/unexpose but somehow indicate recommended use is bind/unbind and then deprecate expose down the road?

client/van_serviceinterface_create.go
```diff
@@ -5,133 +5,22 @@ import (
...
-	"github.com/skupperproject/skupper/pkg/kube"
 )
 
-// TODO: Move to utils
-func stringifySelector(labels map[string]string) string {
-	result := ""
-	for k, v := range labels {
-		if result != "" {
-			result += ","
+func (cli *VanClient) VanServiceInterfaceCreate(ctx context.Context, service *types.ServiceInterface) error {
```
ajssmith, line 12: An initial thought was to have types that are for api "interfaces" that may have a relation to but differ from internal types such that here would be the use of VanServiceInterfaceCreateOptions. The question is whether it is worth maintaining two separate types or consolidate, we can always come back and address later on.

grs: VanServiceInterfaceList already uses the ServiceInterface in its signature. I think using that same type as the return for VanServiceInterfaceInspect makes sense (but returning a single instance rather than a list of them). It then feels reasonably natural to use that as the input to VanServiceInterfaceCreate or VanServiceInterfaceUpdate.

client/van_serviceinterface_update.go
grs: Possibly, but I'd quite like to be able to handle e.g. that getting deleted and recreated.

client/van_serviceinterface_update.go
grs: Agree on not having two so similar methods in the API. I have removed the expose method and now that is merely a convenience offered by the cli tool.

It might even be worth removing Bind/Unbind and just having Update.

I do still have the somewhat ugly GetHeadlessServiceConfiguration in the api. Not sure it is in quite the right shape or place yet but perhaps we can live with it for now and discuss some more to see if we can find a nicer way to handle that.

PR #4, opened 2020-04-16
Initial implementation of an end-to-end test

fgiorgetti:
This PR demonstrates an approach for writing tests (using shipshape) in the skupper
repository, without messing around with to main go.mod and go.sum files.

The test directory is a module itself, and it holds its own go module files.

If you guys like the approach, we can add integration with Travis or Circle-CI.


test/README.md
```diff
@@ -0,0 +1,7 @@
+# Running tests
+
+1. First you need to be authenticated into a cluster (with cluster-admin permissions)
```
grs, line 3: Out of curiousity, why does it need cluster-admin?

grs:
I like this, great work @fgiorgetti !

test/README.md
fgiorgetti: Shipshape creates namespaces in the cluster and (eventually) some cluster level resources, like the CRDs.

grs: Of course, makes sense!

test/README.md
```diff
@@ -0,0 +1,8 @@
+# Running tests
```
nicob87, line 1: comment not related to this line:
with this PR the "test" folder is owned by shipshape, I would rename to test_shipshape or create a subdirectory /test/shipshape.
So if for any reason we want to create anything "test" related but not shipshape specific, we can do it consistently.

fgiorgetti: @nicob87 thanks for bringing up this point.

The `test` directory was defined as a separate go module, to avoid clashes between test dependencies and skupper dependencies, it is not "owned" by shipshape, but is simply has shipshape as dependency. And it was done this way because shipshape currently depends on an older version of the k8s APIs.

IMHO we should use a standard way for writing our tests, in order to avoid code duplication and promote reuse, and maybe shipshape is not the best way to achieve that (which can be a true and fair statement).

Particularly, I do not like naming a test folder as test_shipshape. Instead I think we can organize tests by level/type inside test directory. Just for the records, inside test directory we currently have `examples/tcpecho` as a separate and individual test suite and we can have other individual suites with different test levels or purposes.

One thing I am also not 100% sure is about using a separate go module for testing (that is one of the reasons for having this PR as a proposal for a testing approach).

If shipshape is not a good fit for writing tests for skupper, or if having a separate go module for testing is a bad practice, then we must discuss other approaches, like creating common base types (like an internal framework) inside this repository to share common needs for all tests we are going to create.

@grs and @ajssmith please share your thoughts as well.

test/examples/tcpecho/setup.go
```diff
@@ -0,0 +1,58 @@
...
+	ENV_CLUSTER_LOCAL = "CLUSTER_LOCAL"
+)
+
+var _ = ginkgo.BeforeEach(func() {
+
+	// Loading configuration (ini or environment)
+	Config = testcommon.LoadConfig("skupper/tcpecho")
+
+	// Initializes using only Skupper Operator
+	skupperOperator := operators.SupportedOperators[operators.OperatorTypeSkupper].(*operators.SkupperOperatorBuilder)
```
nicob87, line 34: The operator that must be used in the test is the operator this PR not the master or verisoned one. Is it possible to do that?

test/README.md
fgiorgetti: @grs @ajssmith @nicob87 I am closing this PR, as per our discussion, it might be better to keep the e2e tests separate from this repo for now, as we are going to start with integration tests here first.
Thanks for the review and suggestions.

fgiorgetti:
Closing as this kind of test will be kept in a separate repo.

PR #12, opened 2020-05-01
Setup for mock client and tests

ajssmith:
Add support for mock client for use in tests. Initial example set of example tests to refine.

### DIFF
--- a/client/van_connector_create_test.go
+++ b/client/van_connector_create_test.go
@@ -1,0 +1,115 @@
+package client
+
+import (
+	"context"
+	"os"
+	"sort"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/skupperproject/skupper/api/types"
+	"gotest.tools/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+)
+
+func TestConnectorCreateError(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Create the client
+	cli, err := newMockClient("skupper", "", "")
+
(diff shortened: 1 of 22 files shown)


client/van_connector_create_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/skupperproject/skupper/api/types"
+	"gotest.tools/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+)
+
+func TestConnectorCreateError(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
nicob87, line 21: for now van_connector_create is not using ctx, this "cancel", call is for a future change? or just in case ... right?

client/van_connector_remove_test.go
```diff
@@ -0,0 +1,114 @@
...
+			})
+			assert.Check(t, err, "Unable to create connector for "+c.connName)
+		}
+
+		//TODO: remove should distinguish found, not found
+		err = cli.VanConnectorRemove(ctx, c.connName)
+		assert.Check(t, err, "Unable to remove connector for "+c.connName)
+
+		if c.createConn {
+			time.Sleep(time.Second * 1)
```
nicob87, line 104: is this needed? since the client is a mock...

client/van_connector_create_test.go
ajssmith: Yes, use of ctx is intended for a future change. The client calls should be utilizing but does not presently.

client/van_connector_remove_test.go
ajssmith: This is an area to discuss re. what consistent client behavior should be. The question is whether VanConnectorRemove should distinguish whether connName is found or not. Should it be an error if the connector does not exist or simply indicate success because it is "removed". Will plan to update unit test accordingly in the future.

PR #16, opened 2020-05-07
whitespace fixes for controller code

grs:


grs:
Just the result of running `go fmt` against these files.

PR #48, opened 2020-05-15
Using channel and select instead of time.Sleep

nicob87:
Test duration reduced notably:
--- PASS: TestConnectorCreateInterior (0.97s)
--- PASS: TestConnectorCreateInterior (4.60s)

Probably the same approach should be used in the remaining tests, I think almost all cases are similar.

### DIFF
--- a/client/van_connector_create_test.go
+++ b/client/van_connector_create_test.go
@@ -3,12 +3,10 @@ package client
 import (
 	"context"
 	"os"
-	"sort"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/skupperproject/skupper/api/types"
 	"gotest.tools/assert"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


client/van_connector_create_test.go
```diff
@@ -105,11 +102,14 @@ func TestConnectorCreateInterior(t *testing.T) {
 		})
 		assert.Check(t, err, "Unable to create connector")
 
-		// TODO: make more deterministic
-		time.Sleep(time.Second * 1)
-		assert.Assert(t, cmp.Equal(c.secretsExpected, secretsFound, trans), c.doc)
-	}
+		select {
```
ajssmith, line 105: Thanks you for the proposal. While I am not a fan of the time.sleep, the reason I did not use a channel was to have a general pattern for detecting negative results (e.g. to detect unintential creation of resources). While that may not apply so much to the connector tests, it gets uses in other tests where multiple resources are expected. It would be good to discuss whether this is a reasonable pattern for the unit tests or if there is a better way to detect negative results.

nicob87: Aha!, thanks for the feedback, will take a look to those other tests, and think about this, basically I think we should try to reduce testing time, somehow, ideally taking 1s to run all unit-tests, not ideally ... no more than 10 ... ?

nicob87:
This PR has conflicts now and it will be not merged as it is probably, closing for now, in any case it can be reopened if needed.

PR #52, opened 2020-05-22
Updates for console server

ajssmith:
This patch adds missing console server setup:
- adds skupper-site
- updates owner refs
- adds metrics resources
- adds oauth proxy 
- test updates
- etc.

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -25,7 +25,6 @@ type VanServiceInterfaceCreateOptions struct {
 	Port       int
 	TargetPort int
 	Headless   bool
-	Aggregate  string
 }
 
 type VanRouterInspectResponse struct {
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -6,16 +6,17 @@ type VanConnectorCreateOptions struct {
 }
 
 type VanRouterCreateOptions struct {
-	SkupperName       string
-	IsEdge            bool
-	EnableController  bool
-	EnableServiceSync bool
-	EnableConsole     bool
-	AuthMode          string
-	User              string
-	Password          string
-	ClusterLocal      bool
-	Replicas          int32
+	SkupperName         string
+	IsEdge              bool
+	EnableController    bool
+	EnableServiceSync   bool
+	EnableRouterConsole bool
+	EnableConsole       bool
+	AuthMode            string
+	User                string
+	Password            string
+	ClusterLocal        bool
+	Replicas            int32
 }
 
 type VanServiceInterfaceCreateOptions struct {
@@ -24,6 +25,7 @@ type VanServiceInterfaceCreateOptions struct {
 	Port       int
 	TargetPort int
 	Headless   bool
+	Aggregate  string
 }
 
 type VanRouterInspectResponse struct {
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -24,13 +24,9 @@ const (
 	// NamespaceDefault means the VAN is in the  skupper namespace which is applied when not specified by clients
 	NamespaceDefault    string = "skupper"
 	DefaultVanName      string = "skupper"
+	DefaultSiteName     string = "skupper-site"
 	ClusterLocalPostfix string = ".svc.cluster.local"
 )
-
-// KeyValue holds a key/value pair
-type KeyValue struct {
-	Key, Value string
-}
 
 // TransportMode describes how a qdr is intended to be deployed, either interior or edge
 type TransportMode string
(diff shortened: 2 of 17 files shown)


api/types/client.go
```diff
@@ -24,6 +25,7 @@ type VanServiceInterfaceCreateOptions struct {
 	Port       int
 	TargetPort int
 	Headless   bool
+	Aggregate  string
```
grs, line 28: This feels logically distinct from the console server. If including it I would also include the EventChannel attribute?

api/types/types.go
```diff
@@ -291,10 +287,13 @@ type Credential struct {
...
 	Post        bool
+	SiteOwner   bool
+	Data        map[string][]byte
 }
 
 type CertAuthority struct {
-	Name string
+	Name      string
+	SiteOwner bool
 }
```
grs, line 297: What is the SiteOwner for?

client/van_router_create.go
```diff
@@ -47,16 +71,40 @@ func GetVanControllerSpec(options types.VanRouterCreateOptions, van *types.VanRo
...
-	mounts := &[]corev1.VolumeMount{}
+	sidecars := []*corev1.Container{}
+	volumes := []corev1.Volume{}
+	mounts := make([][]corev1.VolumeMount, 0)
+	mounts = append(mounts, []corev1.VolumeMount{})
+
+	if options.AuthMode == string(types.ConsoleAuthModeOpenshift) {
+		sidecars = append(sidecars, OauthProxyContainer("skupper-proxy-controller"))
+		envVars = append(envVars, corev1.EnvVar{Name: "METRICS_PORT", Value: "8888"})
+		envVars = append(envVars, corev1.EnvVar{Name: "METRICS_HOST", Value: "localhost"})
```
grs, line 82: I wonder if it makes more sense to separate out the env vars by container?

client/van_router_create.go
```diff
@@ -19,7 +17,33 @@ import (
...
-func GetVanControllerSpec(options types.VanRouterCreateOptions, van *types.VanRouterSpec, transport *appsv1.Deployment) {
+func OauthProxyContainer(serviceAccount string) *corev1.Container {
+	return &corev1.Container{
+		Image: "openshift/oauth-proxy:latest",
+		Name:  "oauth-proxy",
+		Args: []string{
+			"--https-address=:8443",
+			"--provider=openshift",
+			"--openshift-service-account=" + serviceAccount,
+			"--upstream=http://localhost:8888",
```
grs, line 28: Use types.ConsoleOpenShiftServicePort here?

client/van_router_create.go
```diff
@@ -78,18 +126,68 @@ func GetVanControllerSpec(options types.VanRouterCreateOptions, van *types.VanRo
...
 	})
 	van.Controller.RoleBindings = roleBindings
+
+	svctype := "ClusterIP"
+	metricsPort := []corev1.ServicePort{
+		corev1.ServicePort{
+			Name:       "metrics",
+			Protocol:   "TCP",
+			Port:       8080,
+			TargetPort: intstr.FromInt(8080),
```
grs, line 136: use types.ConsoleDefaultServicePort here?

client/van_router_create.go
grs: Actually, this is used for both router console and skupper controller console, so unless those always use the same ports, might need to pass ports in?

api/types/client.go
ajssmith: Agreed that it doesn't need to be in this patch set. Relates to the question of VanServiceInterfaceCreateOptions and ServiceInterface types. I will remove for the purposes of this PR.

api/types/types.go
ajssmith: I added this to distinguish owner ref for all resources being either the site or the router deployment as creating the "spec" van_router_create is separate from the resource instantiations. There may be a better name to use or alternative approach.

client/van_router_create.go
ajssmith: Good suggestion, I will revise.

client/van_router_create.go
ajssmith: Good suggestion. I will update. In reviewing skupper-cli, it looked like router console and skupper controller console are using the same ports. Agree there may be cases to separate.

api/types/types.go
grs: I think all that matters is that the root owner is the configmap (so that deleting that delets everything). If some resources are 'owned' by the router deployment, which is itself 'owned' by the configmap, that is fine. I don't think there is a firm reason to choose one over the other (its more of a pragmatic choice as things evolve) so I would argue it shouldn't muddy the data structures. Wdyt?

ajssmith: Totally agree, will simplify with single root owner.

client/van_router_create.go
ajssmith: It turns out the env vars are already separated by container. If this needs to be done more generally, will address in a future patch set.

grs: Ah, looking at it again, the env vars are all for the main container. The oauthproxy container doesn't have any env vars. Confused myself when reading. Ignore this comment.

PR #55, opened 2020-05-29
Initial commit: van service interface create test.

mgoulish:
Test 4 intentionally fails -- maybe we can discuss....
Also would like to discuss style, etc.

### DIFF
--- a/client/van_serviceinterface_create_test.go
+++ b/client/van_serviceinterface_create_test.go
@@ -1,0 +1,121 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"github.com/skupperproject/skupper/pkg/utils"
+
+	"gotest.tools/assert"
+)
+
+var fp = fmt.Fprintf
+
+func TestVanServiceInterfaceCreate(t *testing.T) {
+	testcases := []struct {
+		name  string
+		doc   string
+		init  bool
+		err   string
+		addr  string
+		proto string
+		port  int
+	}{
+		{
+			name:  "vsic_1",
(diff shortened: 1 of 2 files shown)


client/van_serviceinterface_create_test.go
```diff
@@ -0,0 +1,121 @@
...
+			port:  0,
+			err:   "Skupper not initialised in skupper",
+		},
+		{
+			name:  "vsic_2",
+			doc:   "Normal initialization.",
+			init:  true,
+			addr:  "half-addr",
+			proto: "tcp",
+			port:  666,
```
nicob87, line 43: Devil!

client/van_serviceinterface_create_test.go
```diff
@@ -0,0 +1,121 @@
...
+			_, err = kube.NewTransportDeployment(van, &siteOwnerRef, cli.KubeClient)
+			assert.Assert(t, err == nil)
+
+			service = types.ServiceInterface{
+				Address:  c.addr,
+				Protocol: c.proto,
+				Port:     c.port,
+			}
+		} else {
+			service = types.ServiceInterface{}
```
nicob87, line 98: I think... this "else" is not needed. service variable is created (and initialized empty) before the "if c.init"

client/van_serviceinterface_create_test.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		err = cli.VanServiceInterfaceCreate(ctx, &service)
+
+		switch c.err {
+		case "":
+			if err != nil {
+				fp(os.Stdout, "Test %s failure: %s  An error was reported where none was expected.\n", c.name, c.doc)
+			}
+			assert.Assert(t, err == nil)
```
nicob87, line 108: assert.Check(t, err, fmt.Sprintf("Test %s failure: %s  An error was reported where none was expected.\n", c.name, c.doc))

PR #56, opened 2020-05-29
Check port number when validating interface.

mgoulish:
Do a basic reality check on the port number when validating the interface.

### DIFF
--- a/client/van_serviceinterface_update.go
+++ b/client/van_serviceinterface_update.go
@@ -150,6 +150,18 @@ func updateServiceInterface(service *types.ServiceInterface, overwriteIfExists b
 }
 
 func validateServiceInterface(service *types.ServiceInterface) error {
+	if service.Headless != nil {
+		if service.Headless.TargetPort < 0 || 65535 < service.Headless.TargetPort {
+			return fmt.Errorf("Bad headless target port number: %d", service.Headless.TargetPort)
+		}
+	}
+
+	for _, target := range service.Targets {
+		if target.TargetPort < 0 || 65535 < target.TargetPort {
+			return fmt.Errorf("Bad target port number. Target: %s  Port: %d", target.Name, target.TargetPort)
+		}
+	}
+
 	//TODO: change service.Protocol to service.Mapping
 	if service.Port < 0 || 65535 < service.Port {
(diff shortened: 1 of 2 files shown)


mgoulish:
I added checks for the targetPorts on the target list and on the headless thingy.

client/van_serviceinterface_update.go
```diff
@@ -150,8 +150,22 @@ func updateServiceInterface(service *types.ServiceInterface, overwriteIfExists b
 }
 
 func validateServiceInterface(service *types.ServiceInterface) error {
+	if service.Headless != nil {
+		if service.Headless.TargetPort < 0 || 65535 < service.Headless.TargetPort {
```
nicob87, line 154: Nice! just as a comment, this kind of atomic functions are the perfect target to use the unit-testing support that the language provides, built-in, i.e. for testing the first "if", just doing this:

[nbrignon@localhost skupper (check_port)]$ cat client/van_serviceinterface_update_test.go 
```
package client

import (
        "testing"

        "github.com/skupperproject/skupper/api/types"
        "gotest.tools/assert"
)

func TestValidateServiceInterface(t *testing.T) {
        service := &types.ServiceInterface{
                Address:      "",
                Protocol:     "",
                Port:         0,
                EventChannel: false,
                Aggregate:    "",
                Headless: &types.Headless{
                        Name:       "",
                        Size:       0,
                        TargetPort: 65536, //invalid port
                },
                Targets: nil,
                Origin:  "",
        }
        err := validateServiceInterface(service)
        assert.Error(t, err, "Bad headless target port number: 65536")

}
```
And running like this:

```
[nbrignon@localhost skupper (check_port)]$ go test -v -count=1 ./client -run=TestValidateServiceInterface
=== RUN   TestValidateServiceInterface
--- PASS: TestValidateServiceInterface (0.00s)
PASS
ok      github.com/skupperproject/skupper/client        0.008s
```

nicob87: Not related to this pr, but, the best practice, is, when you introduce a change, you introduce in the same pr the unit-test that proves that your change works, that way, if you test does not pass, you prevent  breaking master because you are detecting the test failure in the pr.

PR #62, opened 2020-06-04
site config updates

ajssmith:
This patch:
- Sets up site cm defaults 
- Adds setting to enable router console
- Corrects site controlled setting when site-controller is not used

### DIFF
--- a/client/van_site_config_create.go
+++ b/client/van_site_config_create.go
@@ -18,7 +18,18 @@ func (cli *VanClient) VanSiteConfigCreate(ctx context.Context, spec types.VanSit
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "skupper-site",
 		},
-		Data: map[string]string{},
+		Data: map[string]string{
+			"name":                   types.DefaultVanName,
+			"edge":                   "false",
+			"service-controller":     "true",
+			"service-sync":           "true",
+			"console":                "false",
+			"router-console":         "false",
+			"console-authentication": "",
+			"console-user":           "",
+			"console-password":       "",
+			"cluster-local":          "false",
+		},
 	}
 	if spec.SkupperName != "" {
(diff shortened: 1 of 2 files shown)


client/van_site_config_create.go
```diff
@@ -18,7 +18,18 @@ func (cli *VanClient) VanSiteConfigCreate(ctx context.Context, spec types.VanSit
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "skupper-site",
 		},
-		Data: map[string]string{},
+		Data: map[string]string{
+			"name":                   types.DefaultVanName,
```
grs, line 22: With skupper-cli, if the site name is not specified explicitly it defaults to the name of the namespace the site is in. I suspect that is not much better than a fixed name in many cases, but I *think* I have a very mild preference for it.

client/van_site_config_create.go
```diff
@@ -18,7 +18,18 @@ func (cli *VanClient) VanSiteConfigCreate(ctx context.Context, spec types.VanSit
...
+			"edge":                   "false",
+			"service-controller":     "true",
+			"service-sync":           "true",
+			"console":                "false",
```
grs, line 26: I think the console should be on by default (though I am fine if we have to temporarily default to false for some reason).

client/van_site_config_create.go
```diff
@@ -18,7 +18,18 @@ func (cli *VanClient) VanSiteConfigCreate(ctx context.Context, spec types.VanSit
...
+			"router-console":         "false",
+			"console-authentication": "",
```
grs, line 28: console-authentication could default to 'internal' (that might not be a great name, but it is the current default behaviour, so would be consistent with the aim of this patch to show that in the config I think). (Again, if there is a reason to temporarily prefer a different default that is also ok).

PR #64, opened 2020-06-09
Van Service Interface test -- take 2.

mgoulish:
Andy please take a look -- now I am checking expected entities -- but I think I may have gotten into one set of things that are not relevant to the Service Interface creation.


nicob87:
Michael -- nice to meet you!

PR #74, opened 2020-06-18
Nb build un images

nicob87:
Closes skupperproject/skupper#35
This pr pushes a new tag (the githash) of the service-controller image, and at the end, it deletes the image.
For now I am using muy own quay.io/nicob87 registry, to use quay.io/skupper we need authorized user credentials, not sure if this user exists yet... i.e. I am not authorized to remove images from quay.io/skupper.

Since the SKUPPER_SERVICE_CONTROLLER_IMAGE now points to this temporary image:tag, the integration test runs using this pr image, instead of the last released one.

`client/van_router_create.go:    if os.Getenv("SKUPPER_SERVICE_CONTROLLER_IMAGE") != "" {`

TODO: do something similar for site_controller image, for now we are not testing nothing about it. (in the integration tests).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,8 @@ commands:
             echo 'export GOROOT=/usr/local/go' >> $BASH_ENV
             echo 'export GOPATH=$HOME/go' >> $BASH_ENV
             echo 'export PATH=$GOPATH/bin:$GOROOT/bin:$PATH' >> $BASH_ENV
+            echo 'export SKUPPER_SERVICE_CONTROLLER_IMAGE=${CI_SERVICE_CONTROLLER_IMAGE}:${CIRCLE_SHA1}' >> $BASH_ENV
+            echo 'export SKUPPER_SITE_CONTROLLER_IMAGE=${CI_SITE_CONTROLLER_IMAGE}:${CIRCLE_SHA1}' >> $BASH_ENV
             source $BASH_ENV
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


grs:
Could we push the images to the minikube repo (get repo details from minikube docker-env)?

.circleci/config.yml
```diff
@@ -97,6 +96,10 @@ yaml-templates:
       paths:
         - dist
 
+  images_environment: &images_environment
+    CI_SERVICE_CONTROLLER_IMAGE: quay.io/nicob87/service-controller
+    CI_SITE_CONTROLLER_IMAGE: quay.io/nicob87/site-controller
```
fgiorgetti, line 101: @nicob87 can we use some quay.io organization instead of a private repo?

nicob87: your question is related to grs question ... I am using my private just for showing how it works ... my initial intention was to the, change it to quay.io/skupper ...

nicob87:
> Could we push the images to the minikube repo (get repo details from minikube docker-env)?
@grs  Thanks for the review!
if I understand your question, you are suggesting "not pushing anywhere" and hack the minikube cluster registry, so it contains the image. Probably this could work, but, only works in that particular minikube cluster, in that particular job in that particular vm. If you want to run any other job anywhere moreover that minikube cluster, you have to do again the hacking process ... 

I preferred this solution, since, it decouples the problem, it creates a new image, the same as could be the released image, and the test runs in the cluster without needing any magic.

If we add another kind of cluster to the testing, like an aws, or gce or any ... we still can use our temporal image.

And .. if the image is no present in a real registry I think we may have also "imagePullPolicy" issues, since the cluster will fail if it tryes to pull it (for any reason).

grs:
@nicob87 I'm ok with an external registry. This is certainly an improvement, so +1 from me.

ajssmith:
Thanks for this, it is great improvement.

PR #79, opened 2020-06-22
Enable site-controller for multiple namespaces

ajssmith:
This patch enables the site controller to concurrently deploy VANs to separate namespaces.

This patch:
- Updates site-controller deploy.yaml to default namespace and cluster role/binding
- Distinguishes cli namespace from target resource deployment namespace (via opts)

Note: no changes to command line in this patch but could support separate ns in a future patch.

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -1,17 +1,31 @@
 package types
 
 type VanConnectorCreateOptions struct {
-	Name string
-	Cost int32
+	SkupperNamespace string
+	Name             string
+	Cost             int32
+}
+
+type VanConnectorRemoveOptions struct {
+	SkupperNamespace string
+	Name             string
+	ForceCurrent     bool
+}
+
+type VanConnectorInspectResponse struct {
+	SkupperNamespace string
+	Connector        *Connector
+	Connected        bool
 }
 
 type VanSiteConfig struct {
-	Spec             VanSiteConfigSpec
-	Reference        VanSiteConfigReference
+	Spec      VanSiteConfigSpec
+	Reference VanSiteConfigReference
 }
 
 type VanSiteConfigSpec struct {
 	SkupperName         string
+	SkupperNamespace    string
 	IsEdge              bool
 	EnableController    bool
 	EnableServiceSync   bool
@@ -45,9 +59,4 @@ type VanRouterInspectResponse struct {
 	TransportVersion  string
 	ControllerVersion string
 	ExposedServices   int
-}
-
-type VanConnectorInspectResponse struct {
-	Connector *Connector
-	Connected bool
 }
--- a/client/van_router_create.go
+++ b/client/van_router_create.go
@@ -181,8 +181,12 @@ func (cli *VanClient) GetVanRouterSpecFromOpts(options types.VanSiteConfigSpec, 
 	} else {
 		van.Name = options.SkupperName
 	}
+	if options.SkupperNamespace == "" {
+		van.Namespace = cli.Namespace
+	} else {
+		van.Namespace = options.SkupperNamespace
+	}
 
-	van.Namespace = cli.Namespace
 	van.AuthMode = types.ConsoleAuthMode(options.AuthMode)
 	van.Transport.LivenessPort = types.TransportLivenessPort
 
@@ -414,7 +418,7 @@ func (cli *VanClient) GetVanRouterSpecFromOpts(options types.VanSiteConfigSpec, 
 		CA:          "skupper-ca",
 		Name:        "skupper-amqps",
 		Subject:     "skupper-messaging",
-		Hosts:       "skupper-messaging,skupper-messaging." + cli.Namespace + ".svc.cluster.local",
+		Hosts:       "skupper-messaging,skupper-messaging." + van.Namespace + ".svc.cluster.local",
 		ConnectJson: false,
 		Post:        false,
 	})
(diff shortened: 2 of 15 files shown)


cmd/site-controller/controller.go
```diff
@@ -9,47 +9,47 @@ import (
...
+	workqueue            workqueue.RateLimitingInterface
+	siteId               string
 }
 
 func NewSiteController(cli *client.VanClient) (*SiteController, error) {
-	siteInformer:= corev1informer.NewFilteredConfigMapInformer(
+	siteInformer := corev1informer.NewFilteredConfigMapInformer(
 		cli.KubeClient,
-		cli.Namespace,
+		metav1.NamespaceAll,
```
grs, line 35: If we had an env var that could specify the namespace, and would default to all if not present, that would let the site controller be used even without the need to grant it cluster level privileges.

cmd/site-controller/deploy.yaml
```diff
@@ -86,15 +86,16 @@ rules:
...
-kind: RoleBinding
+kind: ClusterRoleBinding
 metadata:
   labels:
     application: skupper-site-controller
   name: skupper-site-controller
 subjects:
 - kind: ServiceAccount
   name: skupper-site-controller
+  namespace: default
```
grs, line 97: Should probably add an explicit namespace to the serviceaccount definition above if we need to specify one here. As above, if possible, I'd like to keep the ability to run without cluster level rights, so if we had the env var for the site controller as described above, we could then have two separate deploy yamls, one to watch all namespaces, one to watch the current namespace.

ajssmith: Do you think it would be helpful to have two deploy .yamls? One to watch "all" and one to watch current namespace?

grs: I think it would be great if we could, yes. It may end up that no one needs the version that watches the current namespace, but it looks like a relatively small effort is required to keep that option open until we know for sure.

client/van_router_create.go
```diff
@@ -181,8 +181,12 @@ func (cli *VanClient) GetVanRouterSpecFromOpts(options types.VanSiteConfigSpec,
 	} else {
 		van.Name = options.SkupperName
 	}
+	if options.SkupperNamespace == "" {
+		van.Namespace = cli.Namespace
```
nicob87, line 185: So now cli.Namespace, is a kind of default? is it possible to just remove it?

client/van_site_config_inspect.go
```diff
@@ -25,11 +25,17 @@ func (cli *VanClient) VanSiteConfigInspect(ctx context.Context, input *corev1.Co
...
+	// TODO: what should the defaults be for name, namespace
 	if skupperName, ok := siteConfig.Data["name"]; ok {
 		result.Spec.SkupperName = skupperName
 	} else {
 		result.Spec.SkupperName = cli.Namespace
 	}
+	if skupperNamespace, ok := siteConfig.Data["namespace"]; ok {
+		result.Spec.SkupperNamespace = skupperNamespace
+	} else {
+		result.Spec.SkupperNamespace = cli.Namespace
```
nicob87, line 37: here you "Get" a resource, if a field is empty, the client is assuming that the content of the field is his own configured Namespace attribute ... is it possible to  just always set the siteConfig.Data["namespace"] when creating the config map? Same for skupperName ...

client/van_router_create.go
ajssmith: With this patch, the cli namespace and the namespace for site deployment can be different. Previously, it assumed they were the same namespace (e.g. default). The cli has not been updated to enable deployment to a different namespace but we could make that change in a future patch.

cmd/site-controller/deploy.yaml
grs: Nice work on making this configurable, thanks!

client/van_router_create.go
grs: The cli already allows the namespace to be specified, and the cli namespace is then set accordingly. I don't see any reason for wanting to then ignore that and specify a namespace via some alternative option.

client/van_site_config_inspect.go
grs: I think 'namespace' probably should not be an entry in the data section of the configmap, but should be implied by the namespace that configmap is in.

cmd/site-controller/deploy-watch-all-ns.yaml
```diff
@@ -0,0 +1,103 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: skupper-site-controller
+  namespace: default
```
grs, line 6: I wonder if we should create and use a specific namespace, e.g. 'skupper-site-controller' perhaps, instead of default? Obviously this is only an example, so it is not hugely important, but I think it might be a bit nicer than polluting default?

ajssmith:
@grs thanks for the helpful review comments, will refine and update...

client/van_router_create.go
ajssmith: The cmd/skupper/skupper.go ensures that the cli namespace setting is defined as originally intended in the create opts. This was just a check to ensure that the options.SkupperNamespace is defined as it is now required and was defaulting to the cli.Namespace.  In the case of the site-controller, it is set to the site namespace.

Alternatively, we could treat an empty SkupperNamespace here as an error/exception if it is not spec'ed in the options.

client/van_site_config_inspect.go
ajssmith: +1, will change

nicob87:
+1!

api/types/client.go
```diff
@@ -1,17 +1,31 @@
 package types
 
 type VanConnectorCreateOptions struct {
-	Name string
-	Cost int32
+	SkupperNamespace string
```
grs, line 4: I was sure I had already made this comment, but can't see it anywhere. Perhaps I forgot to click the submit or something and it got lost.

My instinct is that it would be nicer to have the namespace as another parameter to each of the client calls (i.e. along side the Context. That is more uniform and to me the namespace is a distinct option. If left as "" it would then imply that the cli Namespace would be used. Even nicer might by to have two variants of each client method, e.g. van_router_create() and van_router_create_ns(), where the second option would have the namespace argument.

That said, this comment is really more of a comment on the API not the functionality. I would be fine leaving it as something to discuss later.

ajssmith: Definitely agree with your observations. It would be useful to discuss API conventions generally (e.g. such as multiple variants) and then work towards revising (maybe right after 0.3).

PR #80, opened 2020-06-23
cli readme

pwright:


cmd/skupper/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The [Skupper](https://skupper.io/) command line enables you to:
+
+* Create a Virtual Application Network (VAN) site
+* Connect to other VAN sites
+* Connect services
+
+## Getting `skupper`
+
+To install `skupper`, download the [latest release](https://github.com/skupperproject/skupper/releases).
```
pwright, line 11: @ajssmith not sure if this is how users will get cli?

cmd/skupper/README.md
```diff
@@ -0,0 +1,41 @@
...
+skupper expose
+```
+
+To connect to the first site from a second site:
+
+```
+skupper connect --secret /path/to/mysecret.yaml
+```
+
+For a complete list of `skupper` commands:
```
pwright, line 35: @ajssmith  is it reasonable to only mention the 'bones' of the commands here, and 'promise' to document them more completely later?

cmd/skupper/README.md
ajssmith: You could point them here: https://skupper.io/start/

As I believe most of the examples have been updated to reference this narrative.

cmd/skupper/README.md
ajssmith: That is reasonable to me as we will definitely be iterating/refining over the release horizon.

cmd/skupper/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+To install `skupper`, download the [latest release](https://github.com/skupperproject/skupper/releases).
+
+
+## Using `skupper`
+
+To create your first VAN site:
+
+```
+skupper init
```
grs, line 19: My inclination would be to separate `skupper init` from the creation of the connection token. E.g.

To initialise a namespace to use skupper, run `skupper init` against it. This can be undone by `skupper delete`.

To connect one namespace using skupper to another, you first need a connection token from the target namespace which you can obtain using `skupper connection-token /apth/to/mysecret.yaml`. The token written into the path specified can then be used at the other namespace with `skupper connect /path/to/mysecret.yaml`. After some time you can verify these are connected using `skupper status`.

cmd/skupper/README.md
```diff
@@ -0,0 +1,41 @@
...
+skupper connection-token /path/to/mysecret.yaml
+```
+
+To expose a service from the first site:
```
grs, line 23: Doesn't need to be in the first site. Could be in either or indeed both.

cmd/skupper/README.md
pwright: @grs  is there a difference between 'site' and 'namespace'? ie could I take your suggestion but use 'site'?

grs: Yes, I think that would be fine.

cmd/skupper/README.md
pwright: I've edited to include this (and your other comment)

PR #81, opened 2020-06-25
create site controller readme (fixes #77)

### LINKED ISSUE #77: Add skupper/cmd/site-controller README
This is the site-controller application so README should provide overview of operation and details/content for:

- site config maps
- token requests
- tokens for connect, etc.

pwright:


cmd/site-controller/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+`data:console-user` -  username for 'unsecured' option
+
+`data:console-password` - password for 'unsecured' option
+
+`data:edge` -  (true/false) Set up an edge skupper site
+
+`data:router-console` - (true/false) Set up a Dispatch Router console (not recommended)
+
+`data:service-controller` -
```
pwright, line 31: @ajssmith is this an optional service controller equivalent to the 'proxy-controller'?

What effect does this have for users?

ajssmith: Yes, it is equivalent to --enable-proxy-controller in the cli. 

It would be for a scenario where you want to deploy the router but do not want/need the service controller to be running.

cmd/site-controller/README.md
```diff
@@ -0,0 +1,54 @@
...
+  console-password: "barney"
+  console-user: "rubble"
+  edge: "false"
+  name: skupp3r
+  router-console: "true"
+  service-controller: "true"
+  service-sync: "true"
+kind: ConfigMap
+metadata:
+  name: skupper-site
```
ajssmith, line 53: It would be helpful to document that the presence of this label is what causes the site-controller to take action on this config map. Without the label, the site-controller will filter out evaluation of the cm.

pwright: @ajssmith added note

PR #82, opened 2020-06-25
Update Skupper readme (fixes #75)

### LINKED ISSUE #75: Add skupper README
Add README for https://github.com/skupperproject/skupper 

Project overview
Narrative/links to project components
Licensing

pwright:


README.md
```diff
@@ -1,41 +1,30 @@
-# [WIP] skupper [WIP]
+# Skupper
+Skupper enables cloud communication by enabling you to create a Virtual Application Network.
 
-Note: this repository is a work in progress targeted for the consolidation
-and refactoring of the skupper implementation.
+This upper layer network decouples addressing from the underlying network infrastructure, providing a secure and simple alternative to VPNs.
 
-Command line tool for setting up and managing skupper installations
+You can use Skupper to create simple integrations between clusters as described in the [Getting Started](https://skupper.io/start/index.html) or create a topology that supports:
```
grs, line 6: I don't really like the word 'integrations' here. What about something like:

You can use Skupper to create a network from namespaces on one or more Kubernetes clusters as described in .... There are no restrictions on the topology created which can include redundant paths but does not need to be a full mesh. Skupper then ensures anycast and multicast (for http) communication over that.

Or something along those lines...?

README.md
```diff
@@ -1,41 +1,30 @@
...
 
+Developing Skupper
 
+* [Community](https://skupper.io/community/index.html)
+* [Site controller](cmd/site-controller/README.md)
+* [CLI](cmd/skupper/README.md)
+* [Console](/skupperproject/gilligan)
 
+# Licensing
+Skupper is related to the [Apache QPID Dispatch Router](https://github.com/apache/qpid-dispatch) project and uses the same [Apache License 2.0](https://github.com/skupperproject/skupper/blob/master/LICENSE).
```
grs, line 30: Rather than the vague 'is related to', perhaps 'uses' (and then say 'and is released under the same AL2 license' to avoid repetition of 'uses')?

README.md
```diff
@@ -1,41 +1,29 @@
...
-
-Note: if using minikube, you can get the LoadBalancer service to setup
-an external ip by running minikube tunnel. (Or else use
---cluster-local as described above).
 
+Developing Skupper
 
+* [Community](https://skupper.io/community/index.html)
+* [Site controller](cmd/site-controller/README.md)
+* [CLI](cmd/skupper/README.md) (This replaces the [Skupper CLI repo](https://github.com/skupperproject/skupper-cli))
```
pwright, line 25: @ajssmith does this help alert users to the 'old' cli?

README.md
pwright: @grs updated, lemme know if it works now?

README.md
```diff
@@ -1,41 +1,29 @@
-# [WIP] skupper [WIP]
+# Skupper
 
-Note: this repository is a work in progress targeted for the consolidation
-and refactoring of the skupper implementation.
+Skupper enables cloud communication by enabling you to create a Virtual Application Network.
 
-Command line tool for setting up and managing skupper installations
+This upper layer network decouples addressing from the underlying network infrastructure, providing a secure and simple alternative to VPNs.
```
grs, line 5: I would use 'application layer' rather than 'upper layer' personally.

grs: I feel that describing skupper as a 'simple alternative to VPNs' is too narrow.

README.md
```diff
@@ -1,41 +1,29 @@
...
 
-## Usage
+You can use Skupper to create a network from namespaces inn one or more Kubernetes clusters as described in the [Getting Started](https://skupper.io/start/index.html).
+This guide describes a simple network, however there are no restrictions on the topology created which can include redundant paths. 
+You can also create more complex (full mesh) networks, Skupper then supports anycast and multicast communication.
```
grs, line 9: This last sentence seems to imply that you need a more complex network (perhaps a full mesh) to support anycast and multicast. (Also a full mesh is not what I would mean by a complex network).

README.md
```diff
@@ -2,11 +2,15 @@
 
 Skupper enables cloud communication by enabling you to create a Virtual Application Network.
 
-This upper layer network decouples addressing from the underlying network infrastructure, providing a secure and simple alternative to VPNs.
+This application layer network decouples addressing from the underlying network infrastructure.
+One example usage is to provide a secure and simple alternative to a VPN.
```
grs, line 6: Personally I prefer the notion that skupper enables secure communication across clusters 'without the need for a VPN' rather than being an 'alternative to a VPN'. I appreciate the difference there may be subtle (or only in my own head!), but I don't think we want to be seen as in the same category as a VPN. We want to allow people to communicate securely without needing VPNs.

README.md
```diff
@@ -2,11 +2,15 @@
...
 
 You can use Skupper to create a network from namespaces inn one or more Kubernetes clusters as described in the [Getting Started](https://skupper.io/start/index.html).
-This guide describes a simple network, however there are no restrictions on the topology created which can include redundant paths. 
-You can also create more complex (full mesh) networks, Skupper then supports anycast and multicast communication.
+This guide describes a simple network, however there are no restrictions on the topology created which can include redundant paths.
+
+You can also create more complex networks, for example, a full mesh network.
```
grs, line 11: I don't think a full-mesh is a good example of a complex network. The previous sentence I think is perhaps enough. Maybe we could also explicitly state that sites don't have to be directly connected to be able to communicate, as long as they are connected to the network in some way.

README.md
```diff
@@ -1,41 +1,34 @@
-# [WIP] skupper [WIP]
+# Skupper
 
-Note: this repository is a work in progress targeted for the consolidation
-and refactoring of the skupper implementation.
+Skupper enables cloud communication by enabling you to create a Virtual Application Network.
 
-Command line tool for setting up and managing skupper installations
+This application layer network decouples addressing from the underlying network infrastructure.
+This enables communication without a VPN.
```
pwright, line 6: @grs Maybe 'This enable communication in scenarios where you would typically use a VPN'?

README.md
```diff
@@ -1,41 +1,34 @@
...
 
-## Usage
+You can use Skupper to create a network from namespaces in one or more Kubernetes clusters as described in the [Getting Started](https://skupper.io/start/index.html).
+This guide describes a simple network, however there are no restrictions on the topology created which can include redundant paths.
 
-See `skupper help` or `skupper <command> --help` for details.
+Connecting one Skupper site to another site enables communication both ways.
+Communication can occur using any path available on the network, that is, direct connections are not required to enable communication.
```
pwright, line 12: @grs  Another way to say this would be mention that a topology is a directed graph, would that help?

README.md
grs: Let's add 'secure' in there, then I am delighted! :-)

README.md
grs: Yes, that is better.

README.md
ajssmith: That works. As part of the release, we will need to update the website accordingly and maybe update the README in the skupper-cli to reference here, etc.

PR #84, opened 2020-06-26
update test to run on local cluster

ajssmith:
Update to van router create tests to run against mock and real cluster and improve coverage.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -3,10 +3,18 @@ package client
 import (
 	"flag"
 	"os"
+	"sort"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"gotest.tools/assert"
 )
+
+var trans = cmp.Transformer("Sort", func(in []string) []string {
+	out := append([]string(nil), in...)
+	sort.Strings(out)
+	return out
+})
 
 func TestNewClient(t *testing.T) {
 	testcases := []struct {
--- a/client/van_router_create_test.go
+++ b/client/van_router_create_test.go
@@ -2,8 +2,6 @@ package client
 
 import (
 	"context"
-	//	"fmt"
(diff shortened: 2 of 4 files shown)


client/client_test.go
```diff
@@ -3,11 +3,19 @@ package client
...
+	"github.com/google/go-cmp/cmp"
 	"gotest.tools/assert"
 )
 
+var trans = cmp.Transformer("Sort", func(in []string) []string {
+	out := append([]string(nil), in...)
+	sort.Strings(out)
+	return out
+})
+
```
nicob87, line 18: nice deduplication!

client/van_router_create_test.go
```diff
@@ -193,10 +208,21 @@ func TestVanRouterCreateDefaults(t *testing.T) {
...
+		var err error
+		if !isCluster {
+			cli, err = newMockClient(c.namespace, "", "")
+		} else {
+			cli, err = NewClient(c.namespace, "", "")
+		}
 		assert.Check(t, err, c.doc)
 
-		informers := informers.NewSharedInformerFactory(cli.KubeClient, 0)
+		if isCluster {
```
nicob87, line 220: I think you can create the namespace also in the mock and you just remove this if (the same with when you delete the namespace)

client/van_router_create_test.go
```diff
@@ -193,10 +208,21 @@ func TestVanRouterCreateDefaults(t *testing.T) {
...
+			_, err = kube.NewNamespace(c.namespace, cli.KubeClient)
+			assert.Check(t, err, c.doc)
```
nicob87, line 222: maybe the "delete namespace" can be deferred here, so you are sure it is removed, no matter what kind of error you have in the test

PR #87, opened 2020-06-29
Proposal for combining logging and error reporting, while increasing info returned to user.

mgoulish:
This is not a pull request -- 
Just something to discuss to see what you think of this technique.
Let's discuss...

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -1,13 +1,20 @@
 package types
 
+import (
+         "fmt"
+         "os"
+         "runtime"
+         "time"
+       )
+
 type VanConnectorCreateOptions struct {
 	Name string
 	Cost int32
 }
 
 type VanSiteConfig struct {
-	Spec             VanSiteConfigSpec
-	Reference        VanSiteConfigReference
+	Spec      VanSiteConfigSpec
+	Reference VanSiteConfigReference
 }
 
 type VanSiteConfigSpec struct {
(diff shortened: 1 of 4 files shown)


client/van_serviceinterface_update.go
```diff
@@ -96,10 +96,11 @@ func getServiceInterfaceTarget(targetType string, targetName string, deducePort
 	}
 }
 
-func updateServiceInterface(service *types.ServiceInterface, overwriteIfExists bool, owner *metav1.OwnerReference, cli *VanClient) error {
+func updateServiceInterface(service *types.ServiceInterface, overwriteIfExists bool, owner *metav1.OwnerReference, cli *VanClient, rs *types.Results) {
```
grs, line 99: Why change from a return value to an 'out' parameter?

api/types/client.go
```diff
@@ -51,3 +58,91 @@ type VanConnectorInspectResponse struct {
 	Connector *Connector
 	Connected bool
 }
+
+const (
+	EventCodeNone = iota
+	EventCodeInfo
+	EventCodeWarning
+	EventCodeError
+)
```
nicob87, line 67: Very nice work!

Just my two cents:
- For supporting different verbosity and logging levels I would investigate existing libraries/options...
    from a quick read: 
    https://www.honeybadger.io/blog/golang-logging/ <- this short blog suggests using "logrus" (https://github.com/sirupsen/logrus) I took a look to the project Readme (and examples in there, and looks easy and interesting)

- Moreover the previous comment I like the idea of having our own return type. 
Instead of creating  a custom error new type, we could just implement the error interface which is  very simple (just adding 1 metod to our error_type), and probably it is almost done in the current Print method.

 ```
[nbrignon@localhost skupper (grand_unified_theory)]$ tail -n 5 /usr/lib/golang/src/builtin/builtin.go 
// The error built-in interface type is the conventional interface for
// representing an error condition, with the nil value representing no error.
type error interface {
        Error() string
}
```

examples of custom errors:
https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go
https://golangbot.com/custom-errors/

And following with this logic, if we do not use the error type for logging, and just for reporting errors, the error codes should be different than "info, warning, debug" and could be "somehing like":

```
const (
	ErrorCodeNone = iota
	ErrorCodeNotFound
        ErrorCodeNoSpace
        ErrorCodeNotImplemented
        ErrorCodeYouAreCrazy
        ErrorCodeInvalidArgument
        ErrorCode666
)
```
These are just my thoughts today.... probably missing something haha! but hope this helps!

mgoulish:
Closing this for now. Possible (but not probable) discussion in future.
An approach like this might have some role in future, but I understand better now (I think) what Andy and Gordon are looking for in an improved error/status reporting capability -- and this ain't it.

api/types/client.go
fgiorgetti: Just adding something related to logrus (as I was going through the openshift-installer source code)...
As I mentioned above, it is being used by the openshift-installer as well, but on the other hand, in its
README file they mention a few other alternatives, as logrus is currently in maintenance mode.
But I believe it might be worth evaluating those other libraries as well: [Zerolog](https://github.com/rs/zerolog), [Zap](https://github.com/uber-go/zap), and [Apex](https://github.com/apex/log).

PR #89, opened 2020-06-30
building the binary inside a docker host instead of depending on the host os/arch

nicob87:
Basically I have stolen Andy's idea about using multi stage docker build.
https://docs.docker.com/develop/develop-images/multistage-build/

I think what we need here is exactly the use case for multistage builds, and as it is expected, the size of the base image is only increased by the size of the binary added to it (no overhead):
```

registry.access.redhat.com/ubi8-minimal   latest              c58d4cf08dc2        8 days ago          142MB
quay.io/nicob87/service-controller        multi               1ae6453d76df        36 seconds ago      183MB

$ docker run quay.io/nicob87/service-controller:multi ls -lh service-controller
-rwxr-xr-x. 1 root root 39M Jun 30 17:47 service-controller
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ build-site-controller:
 
 build-controllers: build-site-controller build-service-controller
 
-docker-build: build-controllers
+docker-build:
 	${DOCKER} build -t ${SERVICE_CONTROLLER_IMAGE} -f Dockerfile.service-controller .
 	${DOCKER} build -t ${SITE_CONTROLLER_IMAGE} -f Dockerfile.site-controller .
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -16,28 +16,16 @@ build-site-controller:
...
 	${DOCKER} build -t ${SITE_CONTROLLER_IMAGE} -f Dockerfile.site-controller .
 
 docker-push:
 	${DOCKER} push ${SERVICE_CONTROLLER_IMAGE}
 	${DOCKER} push ${SITE_CONTROLLER_IMAGE}
 
-format:
-	go fmt ./...
-
-client-mock-test:
```
grs, line 30: Did you intentionally remove these makefile targets? Seems unrelated to the purpose of the PR.

nicob87: No!
Fixing it now

nicob87: The faster reviewer in the far west!

PR #93, opened 2020-07-01
Failing ci if code not formatted.

### LINKED ISSUE #94: verifying source format in ci (go fmt ./...)

nicob87:
Also applying format to the remaining unformatted files.
Closes #94

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,16 @@ jobs:
       name: go/default
       tag: "1.13"
     steps:
+      - checkout
+      - run:
+          name: "Enforce Go Formatted Code"
+          command: >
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -144,14 +144,23 @@ jobs:
       name: go/default
       tag: "1.13"
     steps:
+      - checkout
+      - run:
+          name: "Enforce Go Formatted Code"
+          command: >
```
ajssmith, line 150: Thanks for this, it will be great to have the check against PRs.

Would it workto use the 'make format', e.g. in general have the same command setup via make that gets performed in CI

nicob87: Good point. Done.

PR #96, opened 2020-07-01
updates to connector tests

ajssmith:
Revise tests, increase coverage for client connector routines.

### DIFF
--- a/client/van_connector_create.go
+++ b/client/van_connector_create.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"regexp"
 	"strconv"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


client/van_connector_create.go
```diff
@@ -143,7 +125,12 @@ func (cli *VanClient) createConnector(ctx context.Context, secret *corev1.Secret
 		connector.Role = string(types.ConnectorRoleEdge)
 	}
 	qdr.AddConnector(&connector, current)
+
 	_, err := cli.KubeClient.AppsV1().Deployments(options.SkupperNamespace).Update(current)
```
nicob87, line 129: I think this works, and saves us from implementing the "retryOnConflic" logic every time we want to update a resource.
```
        "k8s.io/client-go/util/retry"
....
        retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {

                _, err = cli.KubeClient.AppsV1().Deployments(options.SkupperNamespace).Update(current)
                return err
        })
        if retryErr != nil {
                return fmt.Errorf("Failed to update qdr deployment: %w", err)

```
and maybe on every retry you can retry using the "latest" version of the deployment "something like" :

```
        retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {

                current, err := kube.GetDeployment(types.TransportDeploymentName, options.SkupperNamespace, cli.KubeClient)
                if err != nil {
                        panic(fmt.Errorf("Failed to get latest version of redis: %s", err))
                }
                _, err = cli.KubeClient.AppsV1().Deployments(options.SkupperNamespace).Update(current)
                return err
        })
        if retryErr != nil {
                return fmt.Errorf("Failed to update qdr deployment: %w", retryErr)
```

ajssmith: This is a great suggestion. Will address in a separate patch.

PR #106, opened 2020-07-15
Handle changes to services

ajssmith:
This patch enables updates to existing services to be made. Note, that for now, changes to address or mapping (e.g. protocol) require delete/add.

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -128,6 +128,17 @@ func deleteServiceArgs() func(*cobra.Command, []string) error {
 	}
 }
 
+func updateServiceArgs() func(*cobra.Command, []string) error {
+	return func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			return fmt.Errorf("name of service to update must be specified")
+		} else if len(args) > 1 {
+			return fmt.Errorf("illegal argument: %s", args[1])
+		}
+		return nil
+	}
+}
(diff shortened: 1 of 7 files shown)


cmd/skupper/skupper.go
```diff
@@ -542,6 +553,57 @@ func main() {
 	}
 	cmdService.AddCommand(cmdDeleteService)
 
+	var serviceToUpdate types.ServiceInterface
```
grs, line 556: I was not anticipating a cli option to update services (not to say that I am opposed). To the comment in the PR at the top, is it the case that protocol changes can be made but won't work? Or is it that it will trigger and add/delete?

ajssmith: I can remove the cli option if it is not something we would look to support. 

Initially, the protocol change was included but observed that the bridge-server would get a port in use exception (e.g. the transition of an existing listener to a different protocol). I plan to create an issue to follow up and wanted to get the base service change support as a start.

ajssmith: Thinking about it, I should address protocol change as part of this PR. I can research the issue further.

grs: I have no objection to the cli option. I think it is also fine to commit this without the protocol, but in that case I would be inclined to disallow that option in the cli (i.e. don't let the protocol be changed, or print an error saying not supported yet).

Issue #21, opened 2020-05-09
Ensure check for invalid syntax across all commands


ajssmith:
from skupper-cli #35

nicob87:
Return value must be not zero for error

```
[nbrignon@localhost skupper (grand_unified_theory)]$ ./skupper im broken
Error: unknown command "im" for "skupper"
Run 'skupper --help' for usage.
[nbrignon@localhost skupper (grand_unified_theory)]$ echo $?
0
```

PR #107, opened 2020-07-15
Ensure return value is non-zero if cli command fails with error

grs:
Fixes issue #21

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -197,6 +197,7 @@ func main() {
 			}
 			if err != nil {
 				fmt.Println(err.Error())
+				os.Exit(1)
 			} else {
 				fmt.Println("Skupper is now removed from '" + cli.Namespace + "'.")
 			}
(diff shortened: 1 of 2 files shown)


cmd/skupper/skupper.go
```diff
@@ -197,6 +197,7 @@ func main() {
 			}
 			if err != nil {
 				fmt.Println(err.Error())
+				os.Exit(1)
```
nicob87, line 200: just as a comment, the "log" module has the log.FatalX family that does what you are doing manually here (printing and exiting with error code 1, with the difference that as it is the log module, it also prints the logging header, i.e.:

```
func main() { log.Fatalf("Bye.") }
```
produces:
image

grs: Yes, however at present we aren't using the log module in the output for the cli. I felt an explicit exit was the cleanest solution for now.

nicob87: Correct!

PR #110, opened 2020-07-16
Testing connection being stablished between two skupper sites for different configurations.

nicob87:
Test is table driven, probably we can add more entries to the table.

For now testing clusterLocal=true/false, and edge=true/false

### DIFF
--- a/test/integration/basic/basic.go
+++ b/test/integration/basic/basic.go
@@ -1,0 +1,180 @@
+package basic
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/test/cluster"
+	"gotest.tools/assert"
+)
+
+type BasicTestRunner struct {
+	cluster.ClusterTestRunnerBase
+}
+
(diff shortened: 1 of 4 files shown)


test/integration/basic/basic.go
```diff
@@ -0,0 +1,180 @@
...
+func (r *BasicTestRunner) Setup(ctx context.Context, createOptsPublic types.VanSiteConfig, createOptsPrivate types.VanSiteConfig) {
+	var err error
+	err = r.Pub1Cluster.CreateNamespace()
+	assert.Assert(r.T, err)
+
+	err = r.Priv1Cluster.CreateNamespace()
+	assert.Assert(r.T, err)
+
+	createOptsPublic.Spec.SkupperNamespace = r.Pub1Cluster.CurrentNamespace
+	r.Pub1Cluster.VanClient.VanRouterCreate(ctx, createOptsPublic)
```
ajssmith, line 51: Suggest adding assert on err returned from VanRouterCreate

(and below...)

nicob87: good catch!

PR #111, opened 2020-07-16
VAN Service Interface test now works with a real cluster as well as in

mgoulish:
the mock environment.

### DIFF
--- a/client/van_serviceinterface_create_test.go
+++ b/client/van_serviceinterface_create_test.go
@@ -38,8 +38,8 @@ func check_result(t *testing.T, name string, timeoutSeconds float64, resultType 
 		if len(*found) >= len(expected) {
 			break
 		}
-		time.Sleep(100 * time.Millisecond)
-		timeoutSeconds -= 0.1
+		time.Sleep(time.Second)
+		timeoutSeconds -= 1.0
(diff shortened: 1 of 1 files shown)


client/van_serviceinterface_create_test.go
```diff
@@ -193,43 +205,47 @@ func TestVanServiceInterfaceCreate(t *testing.T) {
...
+		}
 
 		// Create the VAN Service Interface.
 		service := types.ServiceInterface{
 			Address:  testcase.addr,
 			Protocol: testcase.proto,
 			Port:     testcase.port,
 		}
-		err = cli.VanServiceInterfaceCreate(ctx, &service)
+		observedError := cli.VanServiceInterfaceCreate(ctx, &service)
```
ajssmith, line 231: I would have expected that when you run against a real cluster, there will be a service created for each service interface created named testcase.addr. It may be that we need a delay to account for the informer latency to discover the services before the next loop on test cases. Delay maybe a second or two and see what happens. Longer term, we will need something more deterministic.

mgoulish:
You were right.  It takes 10 or 12 seconds for that addr service to show up when using the real cluster.

PR #113, opened 2020-07-17
support for annotated deployments and services

grs:


cmd/service-controller/definition_monitor.go
```diff
@@ -70,6 +95,148 @@ func (m *DefinitionMonitor) runDefinitionMonitor() {
...
+
+func (m *DefinitionMonitor) deleteServiceDefinitionForAddress(address string) error {
+	svc, ok := m.annotated[address]
+	if ok {
+		// delete the svc definition
+		changed := []types.ServiceInterface{}
+		deleted := []string{
+			svc.Address,
+		}
+		return kube.UpdateSkupperServices(changed, deleted, "annotation", m.vanClient.Namespace, m.vanClient.KubeClient)
```
ajssmith, line 214: Should address be deleted from m.annotated prior to returning through kube.UpdateSkupperServiice?

I tried the sequence:
$kubectl annotate deployment mongo-a skupper.io/proxy=tcp
$kubectl annotate deployment mongo-a skupper.io/port=27017
$kubectl annotate deployment mongo-a skupper.io/port-
$kubectl annotate deployment mongo-a skupper.io/port=27017

And it did not return as an exposed service.

grs: The m.annotated map should be updated when the service definitions get updated. There was a bug here however, where the last service was deleted it was not updating the annotated map. I've pushed a fix for that. Good catch, thanks!

PR #122, opened 2020-07-20
Adding build targets to enable remote debug

fgiorgetti:
I have been able to enable remote debugging using dlv (Delve).
There are several references to it and it seems to work with several IDEs.


nicob87 (approved):
This may be very useful for developing/debugging!
Maybe, we can create a dedicated folder like "docker-debug"

And... maybe a Readme explaining how to use them also could be good.

fgiorgetti:
It might be a good idea to move things under docker-debug/.

Do you guys think we'd also need a separate Makefile or is it
ok to use the existing one?

I will prepare a README as well and place it under a docker-debug
or something similar.

fgiorgetti:
Closing as this PR is outdated. Once this is needed by others, I can create a new one.

PR #125, opened 2020-07-21
Support for a kubernetes service name as a target 

grs:


grs:
@ajssmith sorry andy, I was a bit hasty with that push. Am actually testing out a better change now that attempts to deduce the port from the target service instead of this. Will push again asap. (I'm trying to stop taking shortcuts!)

Issue #90, opened 2020-06-30
skupper status reports incorrect connect info with edge mode


ajssmith:
Observed while deploying skupper-example-mongodb-replica-set

Three router setup, 2 public (interior mode) and 1 private (edge mode)

Public2 connects to Public1
Private1 connects to Public1 and Public 2

From Public interiors 'skupper status' generates:
VanRouter is enabled for namespace '"skupper" in interior mode'. It is connected to 3 other sites (1 indirectly). It has no exposed services.

From Private edge 'skupper status' generates:
VanRouter is enabled for namespace '"skupper" in edge mode'. It is connected to 3 other sites (2 indirectly). It has no exposed services.

Expected result was that status returns: 'It is connected to 2 other sites'.

Changing Private deployment from edge to interior produced expected result.

QDR version 1.12.0

ajssmith:
Thinking about it further, would correct status be: 'It is connected to 2 other sites (1 indirectly)' as there is only one active edge uplink?

Are there any rules adding connectors when in edge mode?

grs:
There can only be one active connector from an edge at one time.

The skupper-cli repo had this same issue and fixed it, possibly that was after work had begun on this one?

grs:
Comment above from me is inaccurate. Should learn to read the issue :-(

mgoulish:
Gordon - thanks for noticing redundant code -- all removed now -- same testing results as before.
Tests 1 through 4 good -- number 5 fails the same way.

PR #142, opened 2020-07-27
skupper status reports incorrect connect info with edge mode (#90)

mgoulish:
New function just for finding directly and indirectly connected nodes for edges.

### DIFF
--- a/pkg/qdr/mgmt.go
+++ b/pkg/qdr/mgmt.go
@@ -46,7 +46,7 @@ type Connection struct {
 	Dir        string `json:"dir"`
 }
 
-func getConnectedSitesFromNodesEdge(nodes []RouterNode, namespace string, clientset kubernetes.Interface, config *restclient.Config) (types.TransportConnectedSites, error) {
+func getConnectedSitesFromNodesEdge(namespace string, clientset kubernetes.Interface, config *restclient.Config) (types.TransportConnectedSites, error) {
 	result := types.TransportConnectedSites{}
 	direct := make(map[string]bool)
 	indirect := make(map[string]bool)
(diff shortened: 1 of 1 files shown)


pkg/qdr/mgmt.go
```diff
@@ -46,19 +46,56 @@ type Connection struct {
 	Dir        string `json:"dir"`
 }
 
-func getConnectedSitesFromNodes(nodes []RouterNode, direct bool, namespace string, clientset kubernetes.Interface, config *restclient.Config) (types.TransportConnectedSites, error) {
+func getConnectedSitesFromNodesEdge(nodes []RouterNode, namespace string, clientset kubernetes.Interface, config *restclient.Config) (types.TransportConnectedSites, error) {
```
grs, line 49: The `nodes` argument does not get used here and should be removed for clarity.

pkg/qdr/mgmt.go
```diff
@@ -80,9 +117,10 @@ func GetConnectedSites(edge bool, namespace string, clientset kubernetes.Interfa
 			if uplink == nil {
 				return result, nil
 			} else {
+
 				nodes, err := getNodesForRouter(uplink.Container, namespace, clientset, config)
 				if err == nil {
-					result, err := getConnectedSitesFromNodes(nodes, false, namespace, clientset, config)
+					result, err := getConnectedSitesFromNodesEdge(nodes, namespace, clientset, config)
```
grs, line 123: As above, the `nodes` passed in here are not needed or used in `getConnectedSitesFromEdge()`, so the `if edge` part of this should be completely replaced by the call to the new `getConnectedSitesFromEdge()`.

We should also get rid of the `getEdgeUplink()` method, which is now redundant (and was based on the incorrect assu,ption that there would be only one uplink).

Issue #95, opened 2020-07-01
Flaky CI pipeline, needs tuning.


nicob87:
https://app.circleci.com/pipelines/github/skupperproject/skupper/150/workflows/b39d9140-e1eb-4559-b13b-0f416b88713e/jobs/866
sometimes, probably due to variation in the VM bandwidth, The test fails, moreover the internet issue, probably the test can be fixed, to retry, or wait or something, in order the make it work always.

nicob87:
To improve this, at least, the port forwarding must be implemented using the "client-go" api instead of calling the cli "kubectl port-forward" command:
Some doc about that:
https://github.com/kubernetes/client-go/issues/51
https://gianarb.it/blog/programmatically-kube-port-forward-in-go
https://github.com/gianarb/kube-port-forward/blob/master/main.go

nicob87:
Without the port-forwarding here we have had the same error:
https://app.circleci.com/pipelines/github/skupperproject/skupper/266/workflows/15f31f26-1863-4c00-95b1-f0dff1228e6b/jobs/1319

Specifically doing:

```
conn, err := net.DialTCP("tcp", nil, tcpAddr)                                                                   
if err != nil {                                                                                                 
return fmt.Errorf("Dial failed: %s\n", err.Error())                                                     
}
```

With this information some resear may be done, it it makes sense probably just retrying the dialTCP attempt may be enough to fix this.



```

grs:
It will take some time after the service is exposed for the router (or currently bridge-server) to actually configure the listener and start accepting connections. Retrying seem the right thing for now, we have talked about a way of determining when the service is actually ready (https://github.com/skupperproject/skupper/issues/163)

PR #143, opened 2020-07-27
Basic http test using http://httpbin.org/ server and https://github.com/tsenart/vegeta client.

### LINKED ISSUE #40: http integration test

nicob87:
Basic http test using http://httpbin.org/ server and https://github.com/tsenart/vegeta client.

The test itself is really simple, and in subsequent PRs I think we can improve this a lot, basically adding more tests and asserting some more element. 

vegeta collected Metrics are printed in the test CI log. 

Closes #40

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,6 @@ commands:
       - docker/install-docker
       - go/install
       - kube-orb/install-kubectl
-      - run: go get -u github.com/tsenart/vegeta
       - run: make
 
   run_cluster_tests:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,7 @@ commands:
       - docker/install-docker
       - go/install
       - kube-orb/install-kubectl
+      - run: go get -u github.com/tsenart/vegeta
       - run: make
 
   run_cluster_tests:
--- a/test/integration/http/http.go
+++ b/test/integration/http/http.go
@@ -1,0 +1,177 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/test/cluster"
+	"gotest.tools/assert"
+
+	appsv1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/davecgh/go-spew/spew"
+	vegeta "github.com/tsenart/vegeta/v12/lib"
+)
+
+type HttpClusterTestRunner struct {
+	cluster.ClusterTestRunnerBase
+}
+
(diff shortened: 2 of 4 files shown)


test/integration/http/http.go
```diff
@@ -0,0 +1,177 @@
...
+	var metrics vegeta.Metrics
+	for res := range attacker.Attack(targeter, rate, duration, "Big Bang!") {
+		metrics.Add(res)
+	}
+	metrics.Close()
+
+	spew.Dump(metrics)
+
+	// Success is the percentage of non-error responses.
+	assert.Assert(r.T, metrics.Success > 0.95, "too many errors! see the log for details.")
```
fgiorgetti, line 105: Maybe we should also include a small matrix with other HTTP methods and URLs,
to validate different results, like 404, 500, etc.

.circleci/config.yml
```diff
@@ -66,6 +66,7 @@ commands:
       - docker/install-docker
       - go/install
       - kube-orb/install-kubectl
+      - run: go get -u github.com/tsenart/vegeta
```
fgiorgetti, line 69: Are you getting the test dependency explicitly here to avoid
adding dependencies to go.mod?
If yes, can't we use a generic instruction like:
```go get -t -u ./test/...```

File: test/integration/http/http.go (hunk 1 above)
Comment (nicob87) [reply test/integration/http/http.go:105]: agree, I think we can do that in subsequent PRs

File: .circleci/config.yml (hunk 2 above)
Comment (nicob87) [reply .circleci/config.yml:69]: Actually this is a mistake, should I add the dependency in the go.mod?, also the spew is a new dependency

File: test/integration/http/http.go [hunk 3]
```diff
@@ -0,0 +1,177 @@
...
+	//TODO https://github.com/skupperproject/skupper/issues/95
+	//all this hardcoded sleeps must be fixed, probably along with #95
+	//for now I am just keeping them in the same values that we are using
+	//for tcp_echo test, since in case of reducing test may fail
+	//intermitently
+
+	r.Pub1Cluster.GetService("httpbin", 3*minute)
+	time.Sleep(20 * time.Second) //TODO XXX What is the right condition to wait for?
+
+	r.Pub1Cluster.KubectlExecAsync(fmt.Sprintf("port-forward service/httpbin 8080:80"))
```
ajssmith, line 82: As you commented in #95 would it be good to start with using client-go/tools/portforward from the get go?

nicob87: Correct, that is the ideal, emmm although it is not that straightforward, I am taking a look to that, if I find how to do it, will modify this part

fgiorgetti: One small comment is that it might be interesting to use a dynamic local port, instead of hard coding it to use 8080.
Something like: `port-forward service/httpbin :80` (then we must parse kubectl output to get the dynamic port).

For upstream CI only this seems like not needed now (and eventually we can simply log an enhancement for that at
this point), but thinking of people trying to run it locally, this might be helpful as 8080 is used by default by several apps.

Issue #145, opened 2020-07-27
edge connectivity failure


mgoulish:
With the following topology (see below) one node consistently fails to report.
 
     edge        -->  internal_1
     edge        -->  internal_2
     internal_1 --> internal_3
     internal_3 --> internal_5
     internal_2 --> internal_4
     internal_4 --> internal_6

mgoulish:
What's happening in our code is that the call to exec.Stream() in our fn router_exec() is failing because the command terminates with return code of 1.

I can reproduce this manually by going to namespace 'e' and doing this:

   kubectl exec -it skupper-router-7dbd9d8b55-v4qcg -- qdmanage query --type node
It fails, returning an empty list.
Then look at router logs:
      kubectl logs skupper-router-7dbd9d8b55-v4qcg router

and see a **framing error**!

2020-07-28 18:25:14.106846 +0000 SERVER (info) [C252] Connection from 127.0.0.1:34474 (to localhost:5672) failed: amqp:connection:framing-error connection aborted

mgoulish:
A couple other notes:

  * The framing error in the above comment happened several hours after I had created the namespaces and made all the connections. So it was not because of some effect that only happens before things are settled down.

  * I was able to successfully do other queries, such as:    
       kubectl exec -it skupper-router-7dbd9d8b55-v4qcg -- qdmanage query --type connection

  The 'node' query was the only one that blew up. But that's the one that we use in getNodesForRouter().

grs:
Which router is in namespace 'e'? Do you get an error from qdmanage or it just exits with an error code? If not can you enable PN_TRACE_FRM=1 when running to see if that sheds some light?

grs:
One tangential comment on the topology is that it does not represent a single network, but two distinct networks. The edge will only actively be communicating through one of them at any time. As such it is probably an anti-pattern in topology design (some connection between the interior sets 1,3,5 and 2,4,6 should be made).

However the issue as described still needs investigated as it likely highlights some bug somewhere.

mgoulish:
Namespace 'e' has the edge router -- the i* namespaces are interiors.
In our code we only get an error code (1) back from qdmanage -- but if I run it by hand and then look at edge-router log, I can see the framing error.

Definitely true about the anti-pattern here -- personally I do not think the edge router should allow two simultaneous connections to actually be formed.

Thanks -- I will try to find more info with PN_TRACE_FRM.

mgoulish:
OK, I know some more.
[...]

PR #150, opened 2020-07-31
Fixes to reported connected sites

grs:


grs:
This handles the issue described in https://github.com/skupperproject/skupper/issues/145#issuecomment-666575672 by detecting and warning the user if there are multiple edge uplinks to distinct networks.

It also handles the case where there are more edge sites connected and makes similar fixes to interior status as were made already to edge site status.

cmd/skupper/skupper.go
```diff
@@ -416,6 +416,12 @@ func main() {
 				if vir.Status.TransportReadyReplicas == 0 {
 					fmt.Printf("Skupper is enabled for namespace %q%s. Status pending...", cli.Namespace, modedesc)
 				} else {
+					if len(vir.Status.ConnectedSites.Warnings) > 0 {
+						for _, w := range vir.Status.ConnectedSites.Warnings {
+							fmt.Printf("Warning: %s", w)
+							fmt.Println()
```
nicob87, line 422: isn't this Println equivalent to add a "\n" at the end of the "Printf"?

cmd/skupper/skupper.go
```diff
@@ -416,6 +416,12 @@ func main() {
 				if vir.Status.TransportReadyReplicas == 0 {
 					fmt.Printf("Skupper is enabled for namespace %q%s. Status pending...", cli.Namespace, modedesc)
 				} else {
+					if len(vir.Status.ConnectedSites.Warnings) > 0 {
```
nicob87, line 419: I think this if is not needed, if "len=0" it just do not iterate

grs: The len check will also return 0 if the slice is nil. Arguably you could just check if it is nil or not here, but to me the len is a nice way of checking whether iteration is needed in either case.

nicob87: aha!

cmd/skupper/skupper.go
nicob87: no need to change, I was not sure it this was intentional, Thanks!

PR #161, opened 2020-08-05
Test in kubernetes job

nicob87:
This PR attempts to create a new way of running an integration test.
The main issue we were having is that we need to execute the skupper setup code in the host, because we need to support to interact with several clusters in the same test.
But for executing the test itself it is useful to have access to the cluster network. To solve this previously we were using the "kubectl port-forwarding" hack, with this we just run the code that we want inside the cluster.

Taking advantage  of the go feature that allows to compile go tests into a binary, and execute the binary later...
The idea is to maintain only one single "TEST_IMAGE" that contains all needed test binaries.

TODO:
- migrate http, and basic test to this new format and remove the "port-forwarding" black magic.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,10 @@ parameters:
     default: quay.io/nicob87/site-controller:${CIRCLE_SHA1}
     type: string
 
+  ci_test_image:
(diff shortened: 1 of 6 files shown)


nicob87:
Closing in favour of #162

nicob87:
Closing in favor of #162

PR #162, opened 2020-08-05
Test in Kubernetes Job

nicob87:
This PR attempts to create a new way of running an integration test.
The main issue we were having is that we need to execute the skupper setup code in the host, because we need to support to interact with several clusters in the same test.
But for executing the test itself it is useful to have access to the cluster network. To solve this previously we were using the "kubectl port-forwarding" hack, with this we just run the code that we want inside the cluster.

Taking advantage of the go feature that allows to compile go tests into a binary, and execute the binary later...
The idea is to maintain only one single "TEST_IMAGE" that contains all needed test binaries.

Closes: #172 

TODO:
    migrate http, and basic test to this new format and remove the "port-forwarding" black magic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ VERSION := $(shell git describe --tags --dirty=-modified)
 SERVICE_CONTROLLER_IMAGE := quay.io/skupper/service-controller
 SITE_CONTROLLER_IMAGE := quay.io/skupper/site-controller
 TEST_IMAGE := quay.io/skupper/skupper-tests
-TEST_BINARIES_FOLDER := ./test/integration/bin
+TEST_BINARIES_FOLDER := ${PWD}/test/integration/bin
 DOCKER := docker
 
 
@@ -50,7 +50,7 @@ vet:
 	go vet ./...
 
 clean:
-	rm -rf skupper service-controller site-controller release
+	rm -rf skupper service-controller site-controller release ${TEST_BINARIES_FOLDER}
 
 package: release/windows.zip release/darwin.zip release/linux.tgz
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,9 +1,13 @@
 VERSION := $(shell git describe --tags --dirty=-modified)
 SERVICE_CONTROLLER_IMAGE := quay.io/skupper/service-controller
 SITE_CONTROLLER_IMAGE := quay.io/skupper/site-controller
+TEST_IMAGE := quay.io/skupper/skupper-tests
 DOCKER := docker
 
-all: build-cmd build-controllers
+all: build-cmd build-controllers build-tests
```
fgiorgetti, line 7: Should we include the new target here (as part of the all target)? Once we have dozens
of test apps, we are going to pollute the base directory with test binaries. Maybe we can
put test binaries in a separate place.

test/cluster/cluster_test_runner.go
```diff
@@ -167,3 +171,90 @@ func (cc *ClusterContext) GetService(name string, timeout_S time.Duration) *apiv
...
+				cc.KubectlExec(cmd)
+			} else {
+				if job.Status.Succeeded > 0 {
+					fmt.Println("Job Successful!!")
+					cc.KubectlExec(cmd)
+					return job, nil
+				}
+				fmt.Printf("Job failed!!!??, status = %v\n", job.Status)
+				cc.KubectlExec(cmd)
+				return job, nil
```
fgiorgetti, line 255: Guess it should return an error in this case?

nicob87: I am returning an error if function timeouts waiting for the job, but if the job ends normally with status failed, maybe that just means that the command ended with exit code != 0, I am not returning error intentionally.
Probably I should modify the print message, it is there since I was debugging this.

Makefile
nicob87: I agree, the same applies to skupper, skupper-service-controller and skupper-site-controller binaries?

fgiorgetti: skupper binary should be there (imho), as users might be interested in it.

nicob87: Thanks!, I will modify makefile to create a specific folder for containing integration tests binaries

nicob87: Done

PR #167, opened 2020-08-08
Added unit tests to code related with annotated resources

fgiorgetti:
As I am working on a new integration test (blackbox) that
will validate skupper handling annotated resources, I ended
up creating unit tests for most of the functions used by
the related code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := $(shell git describe --tags --dirty=-modified)
+VERSION := $(shell git describe --tags --dirty=-modified --always)
 SERVICE_CONTROLLER_IMAGE := quay.io/skupper/service-controller
 SITE_CONTROLLER_IMAGE := quay.io/skupper/site-controller
 DOCKER := docker
@@ -36,6 +36,16 @@ client-cluster-test:
 vet:
 	go vet ./...
 
+cmd-test:
+	go test -v -count=1 ./cmd/...
+
+pkg-test:
+	go test -v -count=1 ./pkg/...
+
+.PHONY: test
+test:
+	go test -v -count=1 ./pkg/... ./cmd/... ./client/...
+
 clean:
 	rm -rf skupper service-controller site-controller release
 
@@ -58,6 +68,4 @@ release/darwin/skupper: cmd/skupper/skupper.go
 
 release/darwin.zip: release/darwin/skupper
 	zip -j release/darwin.zip release/darwin/skupper
-
-
 
--- a/cmd/service-controller/definition_monitor_test.go
+++ b/cmd/service-controller/definition_monitor_test.go
@@ -1,0 +1,565 @@
+package main
+
+import (
+	"fmt"
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/client"
+	"gotest.tools/assert"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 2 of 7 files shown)


cmd/service-controller/definition_monitor_test.go
```diff
@@ -0,0 +1,565 @@
...
+	type result struct {
+		service types.ServiceInterface
+		success bool
+	}
+
+	type test struct {
+		name     string
+		service  *corev1.Service
+		expected result
+	}
```
nicob87, line 196: Another minor comment:
In all the test cases it seems to be a
```type test struct {
		name     string
		service  *corev1.Service
		expected result
	}
```
or a:
```
type test struct {
		name       string
		deployment *v1.Deployment
		expected   result
	}
```

maybe, you can just create two globals:
testDeployment and testService

pkg/kube/configmaps_test.go
```diff
@@ -0,0 +1,330 @@
...
+			Namespace: NS,
+			OwnerReferences: []metav1.OwnerReference{
+				{Name: "TestNewConfigMap"},
+			},
+		},
+	}
+	kubeClient.CoreV1().ConfigMaps(NS).Create(existingCm)
+
+	// Add a fake reaction when trying to get or create "error-cm"
+	kubeClient.Fake.PrependReactor("*", "configmaps", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
```
nicob87, line 50: Nice!

nicob87:
I understand CI had not run for this PR because of the "ci not running for fork with ci enabled in the fork also", would be nice to trigger ci before merging.
Unfortunately, the only way of doing that is "pushing a commit" (after unfollowing skupper project in the fork), you can reset the commit and force a push, as a way of triggering the ci, in the same pr, without adding additional junk commits.

cmd/service-controller/definition_monitor_test.go
fgiorgetti: My idea was to keep these inner types contained and private to the
test method, as it is mostly suited to those particular unit tests.

fgiorgetti:
> I understand CI had not run for this PR because of the "ci not running for fork with ci enabled in the fork also", would be nice to trigger ci before merging.
> Unfortunately, the only way of doing that is "pushing a commit" (after unfollowing skupper project in the fork), you can reset the commit and force a push, as a way of triggering the ci, in the same pr, without adding additional junk commits.

Correct. I have some small changes to push and now that I have unfollowed skupper on my fork, I suppose it is going to run.

fgiorgetti:
> Nice test additions.
> 
> Suggest updating Makefile so that these can be run in ci, or we can do it in a separate pr
> 
> pkg-test:
> go test -v -count=1 ./pkg
> 
> service-controller-mock-test:
> go test -v count=1 ./cmd/service-controller
> 
> service-controller-cluster-test:
> go test -v count=1 ./cmd/service-controller -use-cluster

will do

cmd/service-controller/definition_monitor_test.go
nicob87: I am ok with that!

Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION := $(shell git describe --tags --dirty=-modified)
+VERSION := $(shell git describe --tags --dirty=-modified --always)
```
fgiorgetti, line 1: @nicob87 I have added this extra flag here, as I was constantly getting
errors in this particular line. Please see if you're ok with it.

pkg/utils/utils.go
```diff
@@ -42,3 +43,18 @@ func StringifySelector(labels map[string]string) string {
 	}
 	return result
 }
+
+// LabelToMap expects label string to be a comma separated
+// list of key and value pairs delimited by equals.
+func LabelToMap(label string) map[string]string {
```
fgiorgetti, line 49: @ajssmith this is the function that I added to pkg/utils.

PR #169, opened 2020-08-10
add service interface update tests

ajssmith:


client/van_serviceinterface_update_test.go
```diff
@@ -0,0 +1,394 @@
...
+			AuthMode:          "",
+			User:              "",
+			Password:          "",
+			ClusterLocal:      true,
+		},
+	})
+	assert.Assert(t, err)
+
+	// create three service definitions
+	time.Sleep(time.Second * 1)
```
nicob87, line 284: why is this sleep need?

nicob87: I think, we may run this same tests with clusters others that a cluster in localhost, probably we should at least unify this sleeps as a global configurable parameter, so we can reduce or increase then modifying a single line in the code... 
probably this comment is not only for this pr, and we can create an issue for that

client/van_serviceinterface_update_test.go
```diff
@@ -0,0 +1,394 @@
...
+	err = cli.VanServiceInterfaceBind(ctx, si, "deployment", "nginx", "http", 8080)
+	assert.Assert(t, err)
+
+	items, err := cli.VanServiceInterfaceList(ctx)
+	assert.Assert(t, err)
+	assert.Equal(t, len(items), 3)
+
+	if *clusterRun {
+		// this delay is for service-controller to update
+		time.Sleep(time.Second * 30)
```
nicob87, line 341: similar to previous comment, this sleep may be a global defined variable (or environment variable), "serviceControllerUpdateDelayS" or similar

client/van_serviceinterface_update_test.go
```diff
@@ -0,0 +1,394 @@
...
+		if c.port != 0 {
+			si.Port = c.port
+		}
+		if c.eventChannel != si.EventChannel {
+			si.EventChannel = c.eventChannel
+		}
+		if c.aggregate != si.Aggregate {
+			si.Aggregate = c.aggregate
+		}
+		err = cli.VanServiceInterfaceUpdate(ctx, si)
```
nicob87, line 363: After updating, isn't it missing to test for the particular attribut updated? i.e: port modified from 9090 to 9091?

client/van_serviceinterface_update_test.go
```diff
@@ -0,0 +1,394 @@
...
+				Containers: []corev1.Container{
+					{
+						Name:            "tcp-go-echo",
+						Image:           "quay.io/skupper/tcp-go-echo",
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						Ports: []corev1.ContainerPort{
+							{
+								Name:          "http",
+								Protocol:      corev1.ProtocolTCP,
+								ContainerPort: 80,
```
nicob87, line 49: shouldn't this be tcp 9090?

client/van_serviceinterface_update_test.go
nicob87: andd would be nice to add some debugging message on every hardcoded sleep, so we can easily track testing hard delays

PR #170, opened 2020-08-11
First five topology tests for edge connectivity.

mgoulish:
I want to see what everybody thinks of this.  It's all working, but it takes almost 5 freaking minutes to run!
( Leave the verbose variable true, to see the pretty color printing...)

Create the topology, make all the connections, then ask the edge node its status, and see if it has the correct connectivity count.

Table-driven topology creation.
Reminds me of .... Mercury!

### DIFF
--- a/client/edge_connectivity_test.go
+++ b/client/edge_connectivity_test.go
@@ -1,0 +1,414 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"gotest.tools/assert"
+)
+
+/*================================================================
+  Multiple tests of different topologies, in each case with
+  a single edge.
+  Build the topology, then ask the edge about its connectivity.
+  See if it has the right number of direct and indirect
+  connections.
+================================================================*/
+
+var fp = fmt.Fprintf
+
+var verbose bool = true
+var red string = "\033[1;31m"
+var green string = "\033[1;32m"
+var cyan string = "\033[1;36m"
+var resetColor string = "\033[0m"
+
+type SkupperNamespace struct {
+	name   string
+	isEdge bool
+
+	client         *VanClient
+	secretFileName string
+}
+
+type Connexion struct {
+	from string
+	to   string
+}
+
+type TopologyTest struct {
+	name             string
+	diagram          []string
+	namespaces       []*SkupperNamespace
+	connexions       []Connexion
+	direct, indirect int // These are the expected results.
+}
+
+func getNamespace(targetNamespace string, t *TopologyTest) *SkupperNamespace {
+	for _, ns := range t.namespaces {
+		if ns.name == targetNamespace {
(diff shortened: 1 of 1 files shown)


client/edge_connectivity_test.go
```diff
@@ -0,0 +1,414 @@
...
+				return response.Status.ConnectedSites.Direct,
+					response.Status.ConnectedSites.Indirect,
+					nil
+			}
+		}
+		if response != nil {
+			info(testName, fmt.Sprintf("%d : (%d,%d) != (%d,%d)", timeout, response.Status.ConnectedSites.Direct, response.Status.ConnectedSites.Indirect, expectedDirect, expectedIndirect))
+		}
+		time.Sleep(time.Second)
+		timeout--
```
fgiorgetti, line 106: Timeout here means like a number of attempts with a fixed 1sec delay. It's not a problem at all, but I would just rename the variable and, eventually, even use a time ticker (https://golang.org/pkg/time/#NewTicker).

mgoulish:
By the way -- this tests increases total coverage from 50% to 59%.

client/edge_connectivity_test.go
mgoulish: I agree I should rename that variable, but I'm not sure about using a ticker here. 

I wasn't sure what it would do in the case of a slow receiver, like this is.   ( Sometimes the call to VanRouterInspect() returns immediately but sometimes it takes forever-and-a-day. )

So I tried a little example, in which my tick-receiver slept 3 seconds after getting each tick -- and printed out the time of each tick reception.  Here's what it did:
Current time:  2020-08-11 12:56:06.771001102 -0400 EDT m=+1.000125956
Current time:  2020-08-11 12:56:07.771110428 -0400 EDT m=+2.000235270
Current time:  2020-08-11 12:56:10.771080978 -0400 EDT m=+5.000205839
Current time:  2020-08-11 12:56:13.771073602 -0400 EDT m=+8.000198461
Current time:  2020-08-11 12:56:16.771024949 -0400 EDT m=+11.000149785
Current time:  2020-08-11 12:56:**19**.771052147 -0400 EDT m=+14.000176962
Current time:  2020-08-11 12:56:**22**.77106865 -0400 EDT m=+17.000193509
Current time:  2020-08-11 12:56:**25**.771076399 -0400 EDT m=+20.000201297
Current time:  2020-08-11 12:56:**28**.7710596 -0400 EDT m=+23.000184462
Current time:  2020-08-11 12:56:**31**.771085751 -0400 EDT m=+26.000210585
Current time:  2020-08-11 12:56:**34**.771038603 -0400 EDT m=+29.000163376

I think the tick is coming into the channel while my goroutine is 'working' -- and then the moment it's done 'working it receives that tick goes again.  I don't think that's what we want.  If it took 5 or 10 seconds calling VanRouterInspect -- and then came back with no answer -- I don't think we should fire again immediately, but rather just pause for a second (or two) before trying again. 

Otherwise I feel like we're just hammering on something that is not yet ready to answer...

mgoulish:
I am closing this one in favor of another PR I am about to send up with the same test altered to work in the Integration Test environment.

Issue #147, opened 2020-07-29
make it easier/more obvious how to use console


grs:
E.g. include reference and instructions in output from skupper init , change the name of the service / route etc

grs:
Closing as skupper init now contains information. Will tackle renaming of service/route in separate issue.

PR #194, opened 2020-08-13
add console url to inspect response

ajssmith:
In the spirit of #147, adding url to inspect makes it simpler to access the console.

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -59,4 +59,5 @@ type RouterInspectResponse struct {
 	TransportVersion  string
 	ControllerVersion string
 	ExposedServices   int
+	ConsoleUrl        string
 }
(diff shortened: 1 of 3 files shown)


ajssmith:
Yes, 'skupper status' will trigger

pwright:
just cli? eg `kubectl get configmaps skupper-site -o yaml`?

grs:
> just cli? eg `kubectl get configmaps skupper-site -o yaml`?

I think just cli is good enough for now. We need to think through a general approach for conveying information from the skupper system to the user on the command line when the skupper cli is not in use. That is a bigger task and should not hold up improvement on the cli though.

pwright:
@grs that makes sense, but created an issue to track the differences twixt cli and yaml

Issue #175, opened 2020-08-11
Refactor types.go


ajssmith:
Distribute to where used, keep local and private as appropriate

ajssmith:
track as task

PR #201, opened 2020-08-17
use native resource types

ajssmith:
The re-factoring and consolidation of api/types.go (#175 ) will occur over a series of patches. 

In this patch the types that 'wrapped' native resources types in the deployment spec, etc. are removed to simplify and improve maintainability.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -10,6 +10,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


fgiorgetti:
Looks much cleaner.

client/router_create.go
```diff
@@ -109,34 +110,59 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
...
-		Annotations:    annotation,
+	serviceAccounts = append(serviceAccounts, &corev1.ServiceAccount{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ServiceAccount",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "skupper-proxy-controller",
+			Annotations: annotation,
+		},
```
grs, line 128: Just a thought, I wonder if it would be worth having a method in the respective kube packages to create a basic object of each type? It might reduce some of the boiler plate, especially for commonly used types. Not sure though, it is just a suggestion.

Issue #196, opened 2020-08-13
Enable tests to, optionally, run against multiple clusters


fgiorgetti:
The cluster tests are currently being executed against a single cluster.
We need a mechanism to have them running, optionally, against a single
or multiple clusters.

fgiorgetti:
@ajssmith I believe this one can be closed.

PR #206, opened 2020-08-19
Annotated resources integration test

### LINKED ISSUE #130: Test to validate skupper reaction to annotated resources
We need an automated test that validates how Skupper reacts to annotated resources:
- Resources already annotated when Skupper is deployed
- Resources annotated after Skupper is deployed
- Annotation removed from resources

And it must validate all supported resources:
- Services
- Deployments

fgiorgetti:
* TestAnnodatedResources new integration test added
* General flags to be used --kubeconfig and --edgekubeconfig
* Interruption handler to enforce TearDown on CTRL C
* Curl wrapper (to run curl on existing pods and avoid extra images and deployments)

Resolves #130 #196

### DIFF
--- a/test/integration/annotation/annotated_resource_setup.go
+++ b/test/integration/annotation/annotated_resource_setup.go
@@ -21,7 +21,7 @@ const (
 	nsCluster1 = "annotation-1"
 	nsCluster2 = "annotation-2"
 	timeout    = 120 * time.Second
-	interval   = 10 * time.Second
+	interval   = 20 * time.Second
 )
 
 var (
--- a/test/integration/annotation/annotated_resource_setup.go
+++ b/test/integration/annotation/annotated_resource_setup.go
@@ -1,0 +1,318 @@
+package annotation
+
+import (
+	"context"
+	"fmt"
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/client"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"github.com/skupperproject/skupper/pkg/utils"
+	"github.com/skupperproject/skupper/test/utils/k8s"
+	"gotest.tools/assert"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"testing"
+	"time"
+)
+
+const (
+	nsCluster1 = "annotation-1"
+	nsCluster2 = "annotation-2"
+	timeout    = 120 * time.Second
+	interval   = 10 * time.Second
+)
+
+var (
+	clusters           int = 1
+	ctx                    = context.Background()
+	kubeConfigCluster1 string
+	kubeConfigCluster2 string
+	cluster1           *client.VanClient
+	cluster2           *client.VanClient
+	// teardownClusters prevents deleting an existing namespace
+	teardownClusters []*client.VanClient
+)
+
+// test allows defining the matrix to run the test table
+type test struct {
+	name                  string
+	doc                   string
+	modification          func(*testing.T)
+	expectedSites         int
+	expectedServicesProto map[string]string
+}
+
+// Setup simply creates the namespaces using a single
+// cluster by default, or using two clusters if at
+// least two kubeconfig files have been provided. It
+// also initializes the cluster1 and cluster2 variables.
+func Setup(t *testing.T) {
+	var err error
+
+	// Need at least one public (or a default) cluster
+	pubClusters := k8s.KubeConfigFilesCount(t, false, true)
+	if pubClusters == 0 {
+		t.Skip("no public clusters available")
+	}
+	if k8s.KubeConfigFilesCount(t, true, true) > 1 {
+		clusters = 2
+	}
+	t.Logf("Setup - using %d clusters", clusters)
+
+	// Initializing clients
+	t.Log("Initializing clients")
+
+	// cluster1 will be used as public (connection token will be generated from it)
+	kubeConfigCluster1 = k8s.KubeConfigFiles(t, false, true)[0]
+	// cluster2 will connect to cluster1 (can be either an edge, a public or the default)
+	kubeConfigCluster2 = k8s.KubeConfigFiles(t, true, true)[0]
+	cluster1, err = client.NewClient(nsCluster1, "", kubeConfigCluster1)
+	assert.Assert(t, err)
+	cluster2, err = client.NewClient(nsCluster2, "", kubeConfigCluster2)
+	assert.Assert(t, err)
+
+	// Creating namespaces (or failing if already exists or other error occurs)
+	t.Log("Creating namespaces")
+	_, err = kube.NewNamespace(nsCluster1, cluster1.KubeClient)
+	assert.Assert(t, err)
+	teardownClusters = append(teardownClusters, cluster1)
+	_, err = kube.NewNamespace(nsCluster2, cluster2.KubeClient)
+	assert.Assert(t, err)
+	teardownClusters = append(teardownClusters, cluster2)
+}
+
+// Teardown ensures all resources are removed
+// when test completes
+func TearDown(t *testing.T) {
+	t.Logf("Tearding down - using %d clusters", clusters)
+	t.Log("Deleting namespaces")
+	for _, cluster := range teardownClusters {
+		err := kube.DeleteNamespace(cluster.Namespace, cluster.KubeClient)
+		assert.Assert(t, err)
+	}
+}
+
+// DeployResources provides common setup for kicking off all annotated
+// resource tests. It deploys a static set of resources:
+// - Deployments
+// - Services.
+//
+// And it also includes static annotations to the deployed resources.
+//
+// The default resources and annotations (if true) that will
+// be added to both cluster1 and cluster2 are:
+// deployment/nginx  ## to both clusters
+//   annotations:
+//     skupper.io/proxy: tcp
+// service/nginx-1-svc-exp-notarget  ## cluster1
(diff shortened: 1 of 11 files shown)


test/integration/annotation/annotated_resource_setup.go
```diff
@@ -0,0 +1,318 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"testing"
+	"time"
+)
+
+const (
+	nsCluster1 = "annotation-1"
+	nsCluster2 = "annotation-2"
+	timeout    = 120 * time.Second
```
nicob87, line 23: you can import "github.com/skupperproject/skupper/test/cluster"
and then just use: ImagePullingAndResourceCreationTimeout, in the spirit of centralizing the tuning

fgiorgetti: As per our early discussion, I have raised #209 so we can refactor all integration tests to use a standard
base type that provides these common behaviors.

fgiorgetti:
I have just pushed another commit that is stabilizing this new test, then
executed it for 50 times in a row and all attempts passed.

test/integration/annotation/annotated_resource_setup.go
fgiorgetti: like this you mean?
`timeout    = cluster.ImagePullingAndResourceCreationTimeout`

test/integration/annotation/annotated_resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+				return res, nil
+			})
+			assert.Assert(t, err, "timed out waiting for expected services/protocol list to match")
+
+			// 4.3. Validating all exposed services are reachable across clusters
+			for _, cluster := range []*vanClient.VanClient{cluster1, cluster2} {
+				for svc := range test.expectedServicesProto {
+					t.Logf("validating communication with service %s through %s", svc, cluster.Namespace)
+					// reaching service through proxy-controller's pod (with some attempts to make sure bridge is connected)
+					err = utils.Retry(interval, 20, func() (bool, error) {
```
nicob87, line 159: another hardcode 20. sorry for repeiting...

test/integration/annotation/annotated_resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+						endpoint := fmt.Sprintf("http://%s:8080", svc)
+						resp, err = tools.Curl(cluster.KubeClient, cluster.RestConfig, cluster.Namespace, "", endpoint, tools.CurlOpts{Timeout: 10})
```
nicob87, line 161: Just as a comment:
Other approach  here is just to use cluster.CreateTestJob, and you can write the test in pure go.
And the retry logic is supported by the kubernetes cluster itself (backoff limit).
And you save of this complex process of creating a pod-> executing remotely commands in the pod, then parsing the stdout, to finaly get the result (if I understando correctly)

test/integration/annotation/annotated_resource_setup.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// Need at least one public (or a default) cluster
+	pubClusters := k8s.KubeConfigFilesCount(t, false, true)
+	if pubClusters == 0 {
+		t.Skip("no public clusters available")
+	}
+	if k8s.KubeConfigFilesCount(t, true, true) > 1 {
+		clusters = 2
+	}
+	t.Logf("Setup - using %d clusters", clusters)
```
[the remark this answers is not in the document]
fgiorgetti: > * It seems to me you have created a new integration test from scratch, that does not integrate at all with the other existing integration tests.
> * It is using new cli options (to support multiple clusters), that are unknown to the other existing integration tests.

The reason why I have chosen not to use the ClusterTestRunnerBase is that is is currently statically using just one KUBECONFIG file, and I wanted to use it with more than one to be able to run this test across different cluster versions. 

Anyway I have already opened an issue to enhance the common (or base) test types used by the integration
tests in order to standardize the regular workflow #209 .


> * What I would do to avoid diverging and having to maintain two different testing trees is:
> * Strip this PR, removing al the logic related to supporting multiple clusters, and submit it as a separate pr that implements that for all existing integration tests.

Do you guys think it is better to do this now, instead of merging this one, which is supposed to work along with the
others (as long as we do not set the new flags) and working on #209 after that?


> * After the first item it would be not so difficult, to just use the existing ClusterTestRunner and ClusterContexts as the other tests are doing.
> * Finally and as separate PR, that closes the multicluster support issue, you can submit a PR that adds the multi-cluster support, modifying, removing, refactoring or anything you like to the "cluster" package, and that way that change will automatically applied to all integration tests.
> * What is in the cluster package right now if far from being a very beautiful piece of code I know ( I was learning Go while doing that) but also it is a tiny piece of code. I propose you to put your hands on it and modify, etc. As I have mentioned, instead of diverging, and creating a new integration test from scratch.

Anyway, I would be glad to work on #209 as my next assignment for Skupper.

test/integration/annotation/annotated_resource_setup.go
```diff
@@ -0,0 +1,318 @@
...
+	t.Log("Initializing clients")
+
+	// cluster1 will be used as public (connection token will be generated from it)
+	kubeConfigCluster1 = k8s.KubeConfigFiles(t, false, true)[0]
+	// cluster2 will connect to cluster1 (can be either an edge, a public or the default)
+	kubeConfigCluster2 = k8s.KubeConfigFiles(t, true, true)[0]
+	cluster1, err = client.NewClient(nsCluster1, "", kubeConfigCluster1)
+	assert.Assert(t, err)
+	cluster2, err = client.NewClient(nsCluster2, "", kubeConfigCluster2)
+	assert.Assert(t, err)
```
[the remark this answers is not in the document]
fgiorgetti: Correct, same like the other tests using --use-cluster are doing. But as I said above, the only reason for not reusing
was because of the possibility of using different clusters.

I totally agree we need to standardize and prioritize #209.

test/integration/annotation/annotated_resource_setup.go
```diff
@@ -0,0 +1,318 @@
...
+	err = cluster1.ConnectorTokenCreateFile(ctx, "cluster1", tokenFile)
+	assert.Assert(t, err, "unable to create token to cluster1")
+
+	// Connecting cluster2 to cluster1
+	secret, err := cluster2.ConnectorCreateFromFile(ctx, tokenFile, types.ConnectorCreateOptions{SkupperNamespace: cluster2.Namespace})
+	assert.Assert(t, err, "unable to create connection to cluster1")
+	assert.Assert(t, secret != nil)
+
+	// Waiting till skupper is running on both clusters
+	for _, c := range []*client.VanClient{cluster1, cluster2} {
```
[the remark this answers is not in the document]
fgiorgetti: :+1:

test/integration/annotation/annotated_resource_setup.go
```diff
@@ -0,0 +1,318 @@
...
+		t.Logf("waiting for Skupper sites to be connected at: %s", c.Namespace)
+		err = utils.Retry(interval, 20, func() (bool, error) {
```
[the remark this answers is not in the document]
fgiorgetti: Okay, I agree those values should be defined as constants to help with eventual tuning needs.

test/integration/annotation/annotated_resource_test.go
```diff
@@ -0,0 +1,190 @@
...
+				test.modification(t)
+			}
+
+			// 4.2 Validate services from DATA endpoint (with retries)
+			var resp *tools.CurlResponse
+			var consoleData ConsoleData
+			err = utils.Retry(interval, 20, func() (bool, error) {
+				resp, err = tools.Curl(cluster1.KubeClient, cluster1.RestConfig, cluster1.Namespace, "", dataEndpoint, curlOpts)
+				if err != nil {
+					return false, nil
```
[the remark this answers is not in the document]
fgiorgetti: :+1:

test/integration/annotation/annotated_resource_test.go
fgiorgetti: haha, that is fine.

test/integration/annotation/annotated_resource_test.go
fgiorgetti: By the way, this retry is not part of the PR, it is already part of the main `pkg/utils/retry.go`.

This is not what this Curl wrapper is doing actually. We do not need to run any pod, it basically uses
the "curl" binary that is available in most of the images we use (like the ubi image provides curl).
And in this particular initial implementation, we are running curl inside the controller pod.

If I write a pure "go" application, then I would need not just to deal with the code for the client but also
maintain an image, deploy as a job/pod (which would be much more time consuming).

In this case we simply allow users to run Curl on existing pods. If you try it separately, you can call
HTTP servers and get a response (properly parsed) in a few secs (or less).

test/integration/annotation/annotated_resource_setup.go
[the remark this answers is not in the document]
fgiorgetti: > t.Log prints all messages once the test have finished, is that what you want?

Actually it shows up immediately, as long as you run the test using "-test.v" and
unlike using fmt.Println which prints either way. See:

> func (c *B) Logf(format string, args ...interface{})
> Logf formats its arguments according to the format, analogous to Printf, and records the text in the error log. A final newline is added if not provided. For tests, the text will be printed only if the test fails or the -test.v flag is set. For benchmarks, the text is always printed to avoid having performance depend on the value of the -test.v flag.

But as we discussed earlier today, there is something blocking logs from streaming live to the stdout.

fgiorgetti:
I will refactor this PR to use the updated `ClusterTestRunner` and submit a new one.

PR #210, opened 2020-08-20
Prevent self-connection, and test for it.

mgoulish:
See what you think of this.  Storing secret file name in the client, so that we can later disallow self-connection.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -18,6 +18,7 @@ type VanClient struct {
 	KubeClient  kubernetes.Interface
 	RouteClient *routev1client.RouteV1Client
 	RestConfig  *restclient.Config
+	SecretFile  string
 }
 
 func NewClient(namespace string, context string, kubeConfigPath string) (*VanClient, error) {
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -43,6 +43,9 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
 }
 
 func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile string, options types.ConnectorCreateOptions) (*corev1.Secret, error) {
+	if secretFile == cli.SecretFile {
+		return nil, fmt.Errorf("Can't form connection to self: |%s|\n", secretFile)
+	}
 	secret, err := cli.ConnectorCreateSecretFromFile(ctx, secretFile, options)
 	if err != nil {
 		return nil, err
--- a/client/self_connect_test.go
+++ b/client/self_connect_test.go
@@ -1,0 +1,104 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"gotest.tools/assert"
+)
+
+type Test struct {
+	namespaces []string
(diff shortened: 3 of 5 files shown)


client/self_connect_test.go
```diff
@@ -0,0 +1,104 @@
...
+			AuthMode:          "",
+			User:              "",
+			Password:          "",
+			ClusterLocal:      true,
+		},
+	})
+	assert.Check(t, err, "Unable to create public router")
+
+	// Create Private Router: edge. ----------------------
+	err = private_client.RouterCreate(ctx, types.SiteConfig{
```
fgiorgetti, line 70: Does it need to create the private router for this test?

client/client.go
```diff
@@ -18,6 +18,7 @@ type VanClient struct {
 	KubeClient  kubernetes.Interface
 	RouteClient *routev1client.RouteV1Client
 	RestConfig  *restclient.Config
+	SecretFile  string
```
fgiorgetti, line 21: I see your point. But I have a few concerns. 
- we might have multiple tokens, so they would need to be stored in a slice
- if we store then, they will only be available if token were generated by the active pod (if pod was recycled, you'd lose that info)

Maybe instead of storing the secret file name, you could compare the value of the 
metadata.annotations.inter-router-host and inter-router-port from the provided secret
file. If those two match to the running host, then you proceed with rejecting the "self-connection".

mgoulish: Thanks -- that sounds more reasonable -- I will try it.

mgoulish:
Closing this one in favor of forthcoming.

mgoulish:
Sorry, I have not been able to find a way to implement Fernando's suggestion above ^^^.

It looks like the problem is that a secret is also generated on the 'private' namespace that has the same host and port annotations as the secret that my code created on the 'public' namespace. I have not been able to find any fields in the labels or annotations that will let me distinguish a secret that I created from a secret that was autogenerated after the call to ConnectorCreateSecretFromFile() in the function ConnectorCreateFromFile().

I ended up going in directions that amounted to the same thing I had done in this branch -- stash extra information somewhere -- which would have the same issues that Fernando raised with my initial implementation.

So!  I propose reinstating this solution to get an immediate fix -- and adding a new Issue to solve Fernando's concern about information loss on pod-recycle, and storing a slice in case of multiple secrets.

client/connector_create.go
```diff
@@ -43,6 +43,9 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
 }
 
 func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile string, options types.ConnectorCreateOptions) (*corev1.Secret, error) {
+	if secretFile == cli.SecretFile {
```
grs, line 46: I don't see how this can work in general. E.g. when used from the cli, the process exists when the token is created so the inmemory state is lost. Also there could be many connection tokens generated.

mgoulish: Ah!  You mean 'exits', right?
Cripes.  I didn't think of the CLI.
OK.... back to drawingboard.

mgoulish:
Closing this again.
Approach like this cannot work in CLI scenario.
:-(

client/connector_create.go
grs: Doh, yes I meant 'exits'!

PR #211, opened 2020-08-20
Increasing timeout for integration tests

fgiorgetti:


nicob87:
currently the Run Integration Tests step is taking between 10 and 13 minutes, why is that we need to increase it to more than 30?

fgiorgetti:
> currently the Run Integration Tests step is taking between 10 and 13 minutes, why is that we need to increase it to more than 30?

because the default timeout is 10 mins, so if it exceeds 10, it will panic.

nicob87:
For the record:
-timeout d
            If a test binary runs longer than duration d, panic.
            If d is 0, the timeout is disabled.
            The default is 10 minutes (10m).

nicob87:
For the record, an example of the failure also:
https://app.circleci.com/pipelines/github/skupperproject/skupper/310/workflows/b79c7e1f-3ffc-41b6-b8f1-2dbc4cab0989/jobs/1611

.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ commands:
     steps:
       - run:
           name: Run Integration Tests
-          command: go test -tags=integration -v ./test/integration/...
+          command: go test -tags=integration -timeout=30m -v ./test/integration/...
```
nicob87, line 85: Can we start changing it to 15?

fgiorgetti: We can try 15 of course, because the amount of time it would need in CircleCI won't
be the same we observe in our own environment. If it does not work, then we can tune
later.

fgiorgetti: Let me know if you prefer to try with 15 first and I will update the PR.

nicob87: yup, it is not a requirement, since we really do not know if 15 is ok but, 30 could be wrong also, In that spirit, I prefer to increase as less as possible.

fgiorgetti: done.

Issue #185, opened 2020-08-11
cmd/skupper (cli) unit test


ajssmith:
Implement and enhance cli unit test coverage

PR #214, opened 2020-08-24
Unittesting positional arguments verification.

nicob87:
This pr is part of #185, but it does not close it. More changes are needed.

Unittesting positional arguments verification.
related to that several minor changes:
rootCmd configuration moved to init() function. main only rootCmd.Execute() it.
This is the normal way to setup rootCmd, and it is needed, so when unit-testing the code inside init() is executed.
Some functions simplified to be just a function instead of a function that returns a function (exposeTargetArgs, bindArgs, etc).

IMO: there are some inconsistencies and a couple of things that may be improved, having the unittest it is very easy to just change the unit-test to what you want to happen and then refactor the code in order to make the test succeed (TDD!!)

i.e:
```
»·······assert.Error(t, b([]string{"one", "resource/name", "three"}), "extra argument: three")                                            
»·······assert.Error(t, b([]string{"one", "resource/name", "three", "four"}), "illegal argument: four")
```
if you call some commands with one extra argument you get an "extra argument" error, but if you call it with two, you get "illegal argument" error. This has no sense ... if I understand correctly.

moreover that, we support ":" and "/" notation but the "usage" message does not mention it.
 ```
[nbrignon@localhost skupper (nb-skupper-cli-testing-basic)]$ ./skupper bind
Error: Service name, target type and target name must all be specified (e.g. 'skupper bind <service-name> <target-type> <target-name>')
Usage:
  skupper bind <service-name> <target-type> <target-name> [flags]
```
I propose this and some other issues to be addressed in separate pr, linked to issue: #227

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -82,68 +82,62 @@ func requiredArg(name string) func(*cobra.Command, []string) error {
 	}
 }
 
-func exposeTarget() func(*cobra.Command, []string) error {
-	return func(cmd *cobra.Command, args []string) error {
-		if len(args) < 1 || (!strings.Contains(args[0], "/") && len(args) < 2) {
-			return fmt.Errorf("expose target and name must be specified (e.g. 'skupper expose deployment <name>'")
+func stringSliceContains(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
 		}
-		if len(args) > 2 {
-			return fmt.Errorf("illegal argument: %s", args[2])
-		}
-		if len(args) > 1 && strings.Contains(args[0], "/") {
-			return fmt.Errorf("extra argument: %s", args[1])
-		}
-		targetType := args[0]
-		if strings.Contains(args[0], "/") {
-			parts := strings.Split(args[0], "/")
-			targetType = parts[0]
-		}
-		if targetType != "deployment" && targetType != "statefulset" && targetType != "pods" && targetType != "service" {
-			return fmt.Errorf("expose target type must be one of 'deployment', 'statefulset', 'service' or 'pods'")
-		}
-		return nil
 	}
+	return false
 }
 
-func createServiceArgs() func(*cobra.Command, []string) error {
-	return func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/skupper/skupper.go [hunk 1]
```diff
@@ -82,68 +82,62 @@ func requiredArg(name string) func(*cobra.Command, []string) error {
 	}
 }
 
-func exposeTarget() func(*cobra.Command, []string) error {
```
nicob87, line 85: The diff in this file show a lot of changes, but actually there are just 4 or 5 lines modified moreover the new stringSliceContains method

cmd/skupper/skupper.go
```diff
@@ -82,68 +82,62 @@ func requiredArg(name string) func(*cobra.Command, []string) error {
...
+	if len(args) > 1 && strings.Contains(args[0], "/") {
+		return fmt.Errorf("extra argument: %s", args[1])
 	}
+	targetType := args[0]
+	if strings.Contains(args[0], "/") {
+		parts := strings.Split(args[0], "/")
+		targetType = parts[0]
+	}
+	if !stringSliceContains(validExposeTargets, targetType) {
+		return fmt.Errorf("expose target type must be one of 'deployment', 'statefulset', 'service' or 'pods'")
```
grs, line 112: Could use the validExposeTargets  to create the usage message.

grs (approved):
Good improvements!

cmd/skupper/skupper.go
nicob87: Correct! will do that.

cmd/skupper/skupper.go
```diff
@@ -109,7 +109,7 @@ func exposeTargetArgs(cmd *cobra.Command, args []string) error {
 		targetType = parts[0]
 	}
 	if !stringSliceContains(validExposeTargets, targetType) {
-		return fmt.Errorf("expose target type must be one of 'deployment', 'statefulset', 'service' or 'pods'")
+		return fmt.Errorf("expose target type must be one of: %s", validExposeTargets)
```
grs, line 112: Maybe strings.Join would give nicer looking output?

nicob87: yup, sorry for that, now using strings.join and commits squashed

PR #229, opened 2020-08-26
Prevent self connection 2

mgoulish:
Store information in the token file to disallow self-connection.
Tested from CLI as well as unit tests.

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -98,6 +98,10 @@ func (cli *VanClient) ConnectorCreateSecretFromFile(ctx context.Context, secretF
 }
 
 func (cli *VanClient) ConnectorCreate(ctx context.Context, secret *corev1.Secret, options types.ConnectorCreateOptions) error {
+        // Check this secret to make sure it's not one of ours.
+        if secret.Annotations["originating-namespace"] == cli.Namespace {
+          return fmt.Errorf("Can't connect to self. Namespace: |%s|", cli.Namespace)
+        }
 
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		current, err := kube.GetDeployment(types.TransportDeploymentName, options.SkupperNamespace, cli.KubeClient)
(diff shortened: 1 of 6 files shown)


client/connector_create.go
```diff
@@ -98,14 +98,17 @@ func (cli *VanClient) ConnectorCreateSecretFromFile(ctx context.Context, secretF
 }
 
 func (cli *VanClient) ConnectorCreate(ctx context.Context, secret *corev1.Secret, options types.ConnectorCreateOptions) error {
+        // Check this secret to make sure it's not one of ours.
+        if secret.Annotations["originating-namespace"] == cli.Namespace {
```
grs, line 102: Namespace is not sufficient; there could be many sites using the same namespace. It would have to be the site id (which is a uuid and should be unique).

[I do think that perhaps Fernando's suggestion is the simplest solution. I.e. test the host/port that the token is annotated to connect to. I think the issue you hit is something specific to the minikube tunnel, where a loadbalancer service with the same name may not get a unique ip(?). If so that won't work anyway. Using --cluster-local on on or other would fix that.]

pkg/certs/certs.go
```diff
@@ -20,7 +20,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
-	"fmt"
+        "fmt"
 	"io/ioutil"
```
grs, line 24: Looks like unintentional whitespace changes?

pkg/kube/secrets.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/skupperproject/skupper/pkg/utils/configs"
 )
 
+
```
grs, line 16: Looks like unintentional whitespace changes?

client/connector_create.go
grs: One other comment here: tokens generated by the site-controller have a 'skupper.io/generated-by' annotation on them, the value being the site id. It would make sense to have that annotation even for tokens generated using the cli. That would then be something you could test against.

mgoulish:
Version 3 coming up....

PR #230, opened 2020-08-31
Change console build to use yarn install so yarn.lock is honored

ErnieAllen:
The current console install uses 'npm install' to crate the node_modules directory. This ignores the yarn.lock file that is present and retrieves the latest version of all dependencies. This is not what we want. We want to honour the yarn.lock file to ensure the build is reproducible. 

This PR changes the 'npm install' into 'yarn install'. Yarn will then use the yarn.lock file to get the versions of the dependencies that have been tested.


grs:
@ErnieAllen from the description I think you swapped the wrong word? `npm install` -> `npm yarn` instead of `npm install` -> `yarn install`

PR #231, opened 2020-08-31
Skupper cli testing unexpose

nicob87:
Just picking any command to do the first POC,  I am testing here the "unexpose" command.
For each command, we can test the "flags" parsing, and also the "run".

Using RunE: allows to centralize the "exit" point.

Choosing weather you want to print the "usage" message on error (or not) is easy now just calling "slienceCobra".

A Cli interface was created, so we can create our mock cli, with simple implementations of the methods that the run command should use. So for now the mocked interface only has the "ServiceInterfaceUnbind" method.

For the same reason and for centralizing options. the global Options type is created, when all commands are migrated, this "Options" struct must contain all options.

Probably some details will show up when migrating other options but I think with this pattern we should be able to test all comands.

part of: #185

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -730,8 +730,7 @@ func init() {
 }
 
 func main() {
-	if err := rootCmd.Execute(); err != nil {
-		fmt.Println(err)
+	if rootCmd.Execute() != nil {
 		os.Exit(1)
 	}
 }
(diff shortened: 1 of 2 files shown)


nicob87:
Probably a good idea instead of reviewing this one is to review #239, it contains it, but!, taking into account the size of the PRs, I think swashing everything into a single one may be too big.

PR #232, opened 2020-09-01
Use site ID to prevent self-connection.

mgoulish:
Put site ID info in the connection token, and later use it to prevent self-connection.

This commit adds two new tests:
    * the edge-connectivity test that I modified from a previous PR, and
    * the self-connection test.

In both of these tests I had to modify the router creation step to not use the SiteConfigSpec directly, but rather call the SiteConfigCreate function and use what it returns. This is because both of these tests must run on a real cluster, and in that context I need to call SiteConfigCreate() or I will not be able to later access the information that it stores in the namespace.

Both of my tests pass, and all other unit tests pass, but I am seeing one failure in TestConnectorCreateInterior.
That is weird though: it is not exercising my new code, and it does not fail when run individually -- only when run together with all tests. (And my tests do not run before it.) 

So I will create a separate issue to figure out what is going on there.

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -42,6 +42,17 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
 	return "conn" + strconv.Itoa(max)
 }
 
+func (cli *VanClient) isOwnToken(ctx context.Context, secret *corev1.Secret) (bool, error) {
+	siteConfig, err := cli.SiteConfigInspect(ctx, nil)
+	if err != nil {
+		return false, err
+	}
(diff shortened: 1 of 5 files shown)


grs (approved):

nicob87:
@mgoulish  please rebase on master (ci is fixed now) again! so we can see the pipeline run. You should have no conflict.

mgoulish:
Nico -- 

OK, I will rebase.  BUT -- I forgot -- we wanted to make that edge connectivity test run as an integration test, right?
Not a unit test.
So I will go back and look at what I did to make that happen, and my rebased version will include those changes.

mgoulish:
Closing this one in favor of a rebased version, coming soon.

PR #235, opened 2020-09-01
add debug dump facility

ajssmith:
This patch adds debug command with a dump subcommand. The dump action will collect deployment, logs, stats, config maps, etc and store to archive file.

Over time, additional subcommands can be added to simplify common debug actions related to skupper use.


client/router_create.go
```diff
@@ -62,7 +62,7 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 	van.Controller.Replicas = 1
 	//TODO: change these to types constants
 	van.Controller.Labels = map[string]string{
-		"application":          "skupper",
+		"application":          "skupper-service-controller",
```
grs, line 65: The original idea with the 'application' label was to have one way to select all objects related to skupper. The skupper.io/component would then distinguish specific parts. We may need to change this but it would be good to retain some label common to all.

client/client_debug.go
```diff
@@ -0,0 +1,157 @@
...
+			return err
+		}
+		for _, pod := range podList {
+			for container := range pod.Spec.Containers {
+				if pod.Spec.Containers[container].Name == "router" {
+					// while we are here collect qdstats, logs will show these operations
+					for x := range flags {
+						qdr, err := kube.ExecCommandInContainer([]string{"qdstat", flags[x]}, pod.Name, "router", cli.Namespace, cli.KubeClient, cli.RestConfig)
+						if err == nil {
+							writeTar(pod.Name+"-qdstat"+flags[x]+".txt", qdr.Bytes(), time.Now(), tw)
```
grs, line 134: Just thinking aloud, one issue with the tar format is that it makes it harder if people want to edit out urls (or other data) that they consider private/sensitive. Though it is certainly convenient on the receiving side, I wonder if for the user it might be preferable to have a simple text file with some kind of delimiter? (As I say, just thinking aloud and this is already good enough to go with for now I think).

grs (approved):

client/client_debug.go
```diff
@@ -0,0 +1,157 @@
...
+		return err
+	}
+
+	return writeTar(name+"-configmap.yaml", b.Bytes(), time.Now(), tw)
+}
+
+func (cli *VanClient) SkupperDump(ctx context.Context, tarName string, version string) error {
+	configMaps := []string{"skupper-site", "skupper-services", "skupper-internal", "skupper-sasl-config"}
+	deployments := []string{"skupper-site-controller", "skupper-router", "skupper-service-controller"}
+	flags := []string{"-g", "-c", "-l", "-n", "-e", "-a", "-m", "-p"}
```
fgiorgetti, line 81: Maybe just drop a comment saying this flags are for qdstat

client/client_debug.go
fgiorgetti: Makes sense. Maybe (as you said it does not need to be done now) we could have like a "output-directory" writer as well, that just creates files in a given directory. But it is really good as is.

fgiorgetti:
Nice job @ajssmith

nicob87 (approved):
skupperDump works if executed inside the cluster?

client/client_debug.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	var cversions []string
+	vir, err := cli.RouterInspect(context.Background())
+	if err == nil {
+		cversions = append(cversions, fmt.Sprintf("%-30s %s", "client version", version))
+		cversions = append(cversions, fmt.Sprintf("%-30s %s", "transport version", vir.TransportVersion))
+		cversions = append(cversions, fmt.Sprintf("%-30s %s\n", "controller version", vir.ControllerVersion))
+	}
```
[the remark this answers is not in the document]
ajssmith: Changed so that empty cversions is not generated.

Generally, I made the assumption (that can be revisited) that this command is for dumping purposes and that other skupper commands are better suited for status and inspection. What gets dumped is what is readily available to gather which could be environment info. etc. and may not need the router to be running.

nicob87:
Just a note: commit history should at least be squashed before merge (and removed the "fix formatting" part of the message).

PR #236, opened 2020-09-01
Use site ID to prevent self-connection.

mgoulish:
No change to the edge connectivity test -- let it run as unit test for now, per Nico.

### DIFF
--- a/client/edge_connectivity_test.go
+++ b/client/edge_connectivity_test.go
@@ -1,0 +1,413 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"gotest.tools/assert"
+)
+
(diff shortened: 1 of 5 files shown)


nicob87 (approved):
Nice!

client/edge_connectivity_test.go
```diff
@@ -0,0 +1,413 @@
+package client
```
ajssmith, line 1: Per the discussion, i thought this was moving over to test/integration

ajssmith (changes requested):
This looks good. Wasn't the edge test moving locale?

Can you please also edit the pr title to its original?

client/edge_connectivity_test.go
mgoulish: I talked to Nicolas about that and he ended up deciding that it should stay where it is.
I'm afraid I don't recall the reasoning now...

ajssmith (approved):
lgtm

Next step will be to refactor for ClusterTestRunner

mgoulish:
closing in favor of 249 -- rebased.

PR #239, opened 2020-09-03
skupper cli testing expose

nicob87:
this pr contains: #231
Basically it does the required changes to support vanClient calls used during the "skupper expose" command.
Also a  little bit of de-duplication  and unit-testing of specific functions as they were detected during the "expose command" testing development itself.

Commits have been done as stripped as possible, with the intention of facilitating understanding if review is made (commit by commit):

```

84b8774a (origin/nb-skupper-cli-testing-expose) cli: testing "skupper expose"
b759294c cli: implemented vanClient mock required logic to unit-test cmdExpose. And fixed cmdUnexpose test according to this refactor.
05d007db cli: cmdExpose logic moved iniside exposeRun function (so it will be unittested)
a87995c2 cli: ExposeOptions now part of global Options. cli: all Flags and Parsing options must be part of *Args function (removed from Run functions) cli: testing unexposeTargetArgs and exposeTargetArgs
ea83218b vanClient: added GetNamespace getter, so vanClient interface has that method
6757d941 cli: deduplicating targetType and targetName parsing
d5ec1333 cli: cmdExpose returns error if SeriviceInterfaceInspect returns error.
cc5f9788 (origin/nb-skupper-cli-testing-unexpose) Testing unexpose flags parsing.
```

part of: #185

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -20,6 +20,10 @@ type VanClient struct {
 	RestConfig  *restclient.Config
 }
 
+func (cli *VanClient) GetNamespace() string {
+	return cli.Namespace
+}
+
 func NewClient(namespace string, context string, kubeConfigPath string) (*VanClient, error) {
 	c := &VanClient{}
 
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -30,6 +30,14 @@ type ExposeOptions struct {
 	Headless   bool
 }
 
+type Options struct {
+	//all subCommands options will be moved to this struct when refactor is
+	//complete.
+	unexposeAddress string
+}
+
+var options Options
+
 func expose(cli *client.VanClient, ctx context.Context, targetType string, targetName string, options ExposeOptions) error {
 	serviceName := options.Address
 	if serviceName == "" {
@@ -154,16 +162,50 @@ func check(err error) bool {
 	}
 }
 
-func NewClient(namespace string, context string, kubeConfigPath string) *client.VanClient {
+//this function remains since os.Exit(1), is still used, once the all commands
+//are migrated, this function is not needed anymore.
+func NewClient(namespace string, context string, kubeConfigPath string) (*client.VanClient, error) {
 	cli, err := client.NewClient(namespace, context, kubeConfigPath)
 	if err != nil {
 		fmt.Println(err.Error())
 		os.Exit(1)
 	}
-	return cli
+	return cli, nil
 }
 
 var rootCmd *cobra.Command
+
+func silenceCobra() {
+	rootCmd.SilenceErrors = true
+	rootCmd.SilenceUsage = true
+}
+
+type vanClientInterface interface {
+	//all required methods will be added here while unit-testing all
+	//subcommands (for now, for testing unexpose we only need this one)
+	ServiceInterfaceUnbind(ctx context.Context, targetType string, targetName string, address string, deleteIfNoTargets bool) error
+}
+
+func unexposeRun(cmd *cobra.Command, args []string, options Options, cli vanClientInterface) error {
+	targetType := args[0]
+	var targetName string
+	if len(args) == 2 {
+		targetName = args[1]
+	} else {
+		parts := strings.Split(args[0], "/")
+		targetType = parts[0]
+		targetName = parts[1]
+	}
+	err := cli.ServiceInterfaceUnbind(context.Background(), targetType, targetName, options.unexposeAddress, true)
+	if err != nil {
+		return fmt.Errorf("Error, unable to skupper service: %s", err.Error())
+	}
+
+	fmt.Printf("%s %s unexposed\n", targetType, targetName)
+	return nil
+}
+
+type clientCommandFunc func(cmd *cobra.Command, args []string, options Options, cli vanClientInterface) error
 
 func init() {
 	routev1.AddToScheme(scheme.Scheme)
@@ -172,6 +214,18 @@ func init() {
 	var namespace string
 	var kubeconfig string
 
+	ClientCommand := func(run clientCommandFunc) func(cmd *cobra.Command, args []string) error {
+		return func(cmd *cobra.Command, args []string) error {
+			silenceCobra() //if needed this may be optional
+			cli, err := NewClient(namespace, kubeContext, kubeconfig)
+			if err != nil {
+				return err
+			}
+			err = run(cmd, args, options, cli)
+			return nil
+		}
+	}
+
 	var routerCreateOpts types.SiteConfigSpec
 	var cmdInit = &cobra.Command{
 		Use:   "init",
@@ -179,7 +233,7 @@ func init() {
 		Long:  `init will setup a router and other supporting objects to provide a functional skupper installation that can then be connected to other skupper installations`,
 		Args:  cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
-			cli := NewClient(namespace, kubeContext, kubeconfig)
+			cli, _ := NewClient(namespace, kubeContext, kubeconfig)
 			//TODO: should cli allow init to diff ns?
 			routerCreateOpts.SkupperNamespace = cli.Namespace
 			siteConfig, err := cli.SiteConfigInspect(context.Background(), nil)
@@ -214,7 +268,7 @@ func init() {
 		Long:  `delete will delete any skupper related objects from the namespace`,
 		Args:  cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
-			cli := NewClient(namespace, kubeContext, kubeconfig)
+			cli, _ := NewClient(namespace, kubeContext, kubeconfig)
 			err := cli.SiteConfigRemove(context.Background())
 			if err != nil {
 				err = cli.RouterRemove(context.Background())
@@ -234,7 +288,7 @@ func init() {
 		Short: "Create a connection token.  The 'connect' command uses the token to establish a connection from a remote Skupper site.",
 		Args:  requiredArg("output-file"),
 		Run: func(cmd *cobra.Command, args []string) {
-			cli := NewClient(namespace, kubeContext, kubeconfig)
+			cli, _ := NewClient(namespace, kubeContext, kubeconfig)
 			err := cli.ConnectorTokenCreateFile(context.Background(), clientIdentity, args[0])
 			if err != nil {
 				fmt.Println("Failed to create connection token: ", err.Error())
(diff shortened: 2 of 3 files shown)


ajssmith (commented):
@nicob87 thanks for these pr's as there is a lot of work in them.

Wanted to start a conversation on the scope of the tests here. Is the scope the VanClient or to test cli commands with go and cobra?

The current VanClient tests focus on the core van operations and the additional scope here is the cli on top. Generally we want to be able to execute a command, check dependencies, change input and check what command execution "prints" out. 

The approach in this pr is a good one by creating local functions to emulate the command behaviors but changes a lot, will require maintenance to keep behaviors in synch and bypasses the go/cobra path itself.

For discussion:
There are likely many approaches to the test goal but came across the approach of using a constructor function to return a command that can  then be used to drive the go/cobra interactions without any changes to the command function structure.

[https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra](url)

and tried a small prototype:

[https://github.com/ajssmith/skupper/tree/cli-test](url)

With this approach we could run against both a k8s fake/real client and additionally use commands to create runtime context for certain paths all via the go/cobra path.

nicob87:
@ajssmith 
Hi! first of all the intention of these PR's is to "unit-test". It does not intends to be an integration test in any way.

> Wanted to start a conversation on the scope of the tests here. Is the scope the VanClient or to test cli commands with go and cobra?
> 
> The current VanClient tests focus on the core van operations and the additional scope here is the cli on top. Generally we want to be able to execute a command, check dependencies, change input and check what command execution "prints" out.
> 

Specifically what I want to test is (as you mention) "the additional scope here is the cli on top". Since the vanApi testing is done somewhere else I am specifically interested in isolating new changes, from the vanClient code.

To do that (and to reply to your comment) I have taken a look to unit-test documentation, definition, examples and golang specific unit-testing methods.


https://www.guru99.com/unit-testing-guide.html
"
UNIT TESTING is a type of software testing where individual units or components of a software are tested. The purpose is to validate that each unit of the software code performs as expected. Unit Testing is done during the development (coding phase) of an application by the developers. **Unit Tests isolate a section of code** and verify its correctness. A unit may be an individual function, method, procedure, module, or object. 
In SDLC, STLC, V Model, Unit testing is first level of testing done before integration testing. Unit testing is a WhiteBox testing technique that is usually performed by the developer. Though, in a practical world due to time crunch or reluctance of developers to tests, QA engineers also do unit testing. 
"
...
"
Unit Testing Example: Mock Objects

**Unit testing relies on mock objects** being created to test sections of code that are not yet part of a complete application. Mock objects fill in for the missing parts of the program.

For example, you might have a function that needs variables or objects that are not created yet. In unit testing, those will be accounted for in the form of mock objects created solely for the purpose of the unit testing done on that section of code. 
" 

https://stackify.com/unit-testing-basics-best-practices/
"
And, finally, unit tests don’t exercise multiple components of your system and how they act.  If you have a console application and you pipe input to it from the command line and test for output, you’re executing an end-to-end system test — not a unit test.
"

Related to this:
https://www.youtube.com/watch?v=b13zwMPLklE
I found this video specifically about golang unit-testing very interesting, and useful, explaining howto develop in "go" along with coding the unit-tests at the same time. (and mocking basics in "go")

So (except from the very basics functions that only use language primitives as arguments... a.k.a: c = sum(a,b)) from unit-test basics comes the need of a mock. 

So, what is the scop of this tests?
image

Usage message, help message etc may also be tested but lets let that for another discussion.

1) is testing for positional arguments parsing: this is a simple unit-test, since you do not need mocks, just calling the function with the desired positional args.
Simple and straightforward test that allows you to call unexposeTargetArgs with any possible value you want, and assert on the expected error. And actually this test is very usefull since it shows that sometime we are returning "illegal" or "extra" as argument error in a not consistent way (IMHO)
```

func Test_unexposeTargetArgs(t *testing.T) {
	genericError := "expose target and name must be specified (e.g. 'skupper expose deployment <name>'"
	targetError := "expose target type must be one of: [deployment, statefulset, pods, service]"

	u := func(args []string) error {
		return unexposeTargetArgs(nil, args)
	}

	assert.Error(t, u([]string{}), genericError)
	assert.Error(t, u([]string{"depl/name"}), targetError)

	assert.Error(t, u([]string{"depl/name", "two"}), "extra argument: two")
	assert.Error(t, u([]string{"depl/name", "two", "three"}), "illegal argument: three")
	assert.Error(t, u([]string{"depl/name", "two", "three", "four"}), "illegal argument: three")

	assert.Error(t, u([]string{"depl/name"}), targetError)
	assert.Error(t, u([]string{"anything", "name"}), targetError)

	assert.Error(t, u([]string{"deployment"}), genericError)

	assert.Assert(t, u([]string{"deployment", "name"}))

	assert.Error(t, u([]string{"deployment", "name", "three"}), "illegal argument: three")
	assert.Error(t, u([]string{"deployment", "name", "three", "four"}), "illegal argument: three")

	for _, target := range validExposeTargets {
		assert.Assert(t, u([]string{target, "name"}))
	}
}
```

3) is testing for flags arguments parsing: this is done by cobra... the only verification to do is that we are assigning to the right variable the cobra parsed flag.

```
func Test_cmdUnexposeParseArgs(t *testing.T) {
	cmd_args := []string{"unexpose", "deployment/name", "--address", "theAddress"}
	expected_subcmd_args := cmd_args[1:]
	command, subcommand_args, err := rootCmd.Find(cmd_args)
	assert.Assert(t, err)
	assert.Assert(t, cmp.Equal(expected_subcmd_args, subcommand_args))

	assert.Assert(t, command.ParseFlags([]string{}))
	assert.Equal(t, options.unexposeAddress, "")

	assert.Assert(t, command.ParseFlags(expected_subcmd_args))
	assert.Equal(t, options.unexposeAddress, "theAddress")

	//Probably this is excessive testing, as we are testing the cobra library
	//itself, but, it is free!
	assert.Error(t, command.ParseFlags([]string{"--address"}),
		"flag needs an argument: --address")
}
```

2) !!! is the most complicated case, we have logic there, luckly that logic only uses the vanClient object as a dependency, so to test it we only need 1 mock. As an example here will show only the unexpose command.

function that we need to mock:
image

execution path 1 (service interface unbind succeeds):

image

execution path 2 (service interface unbind return error):
image


So if I can execute those 2 execution paths, I will have 100% test coverage over that function, and also I can tell from a unittest perspective that I am 100% sure that, that "unit" of code works as It suppose to do. If then, the real vanClient method crashes, or does things that are not correct, it is not my responsability, that is due a bug in the vanClient implementation AND that bug must be exposed in the vanClient tests, not here.

Finally:
To execute those paths, I have done the only thing you can do in go to mock (and AFAIK: you must): creating an interface, and using it for unit-testing.
Again I recommend this 30 min tutorial:
https://www.youtube.com/watch?v=b13zwMPLklE

For now, I created the interface in the cmd/skupper package and actually the only one that knows about that interface is the cmd/skupper. Along with the interface I am creating the mock, although it looks like a complex change the mock implmementation is very very simple. 
It just saves the calling arguments for methods, and allows you to configure return values. That is all, and that is all we need.

```

func (v *vanClientMock) ServiceInterfaceUnbind(ctx context.Context, targetType string, targetName string, address string, deleteIfNoTargets bool) error {
	var calledWith = serviceInterfaceUnbindCallArgs{
		targetType:        targetType,
		targetName:        targetName,
		address:           address,
		deleteIfNoTargets: deleteIfNoTargets,
	}
	v.serviceInterfaceUnbindCalledWith = append(v.serviceInterfaceUnbindCalledWith, calledWith)

	return v.injectedReturns.serviceInterfaceUnbind
}
```

And here is a drawing of how the unit-test of the run methods works, imo:

image




> The approach in this pr is a good one by creating local functions to emulate the command behaviors but changes a lot, will require maintenance to keep behaviors in synch and bypasses the go/cobra path itself.

Changes a lot: I do not agree, the logic of the run functions remains exactly as it is, just moved into functions...

It will require maintenance: Yes, always unit-test need maintenance, and I think actually any kind of testing you do needs refactoring if you change methods signature, and or behaviour, anyway! I do not think it is such a big job, once the test are working, when you change something you just need to accompany the change with the corresponding test modification.

> 
> For discussion:
> without any changes to the command function structure

Why not changing it,  is a win?

> There are likely many approaches to the test goal but came across the approach of using a constructor function to return a command that can then be used to drive the go/cobra interactions without any changes to the command function structure.

It seems to me that you are proposing an integration test here, that overlaps with the already existing integration tests also, just to test 30 lines of code (per run method), we are re-executing the all function call stack, from the cli to the cluster! when we are already testing from client to cluster.

So IMO: we already have test that test from "client->cluster" and for testing the new code between cli->client, you are proposing to re-execute client->cluster, just calling client methods from the cli instead of directly from go code ... 

Appendix: And following with the same unit-test logic, we also lack of unittests for the vanClient, which would be to mock the kube package and test (vanClient->kube) (and pkg/certs, pkg/utils, etc). 

IMO our client/*_test.go tests even when using the kubernetes mock at the end, are more similar to integration tests, than to unit-tests.

This has several side efects (again IMHO):
    1) time consuming, even when using the kubernetes/cli mock... how many sleep(1) do we have there?
    2) you do not control the execution path, or it is very difficult to do it, to make vanClient api call return an error, you have to reverse the vanClient method code, verify what must be returned by the cluster in order to make it fail, (if you are lucky) and if you are not lucky, you will also have pkg/kube, or similar intermediate functions... i.e:
    to test "skupper init" error condition... you need exercise a all path from:
        cli->vanClient->kube.CreateThings()->kubeClientMock() 
     3) it is likely mathematically and physically  impossible to test all possible execution paths.
           - if you have a cli call ->  that executes a run method -> that executes 4 vanApi calls-> and each api call does 3 kube/something calls, and each kube/something call does 1 or more kubernetesClient calls, first it is almost impossible to cover the combination of all possible successful/error return codes, options of all of that functions, and second if you do it we will probably endup having hours of integration tests, that are actually doing almost the same (it does not scaleup too mutch ... ) I think...


> 
> https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra
> 
> and tried a small prototype:
> 
> https://github.com/ajssmith/skupper/tree/cli-test
> 
> With this approach we could run against both a k8s fake/real client and additionally use commands to create runtime context for certain paths all via the go/cobra path.

I think I have already replied this, but in this pr, doing this is specifically what I do not want to do. I think writing this kind of integration test you are proposing here is possible. And probably a different discussion...


And finally again some more diagram, thanks for your patience if you reached this point!:
- This is what I understand you are proposing as solution for cli testing:

image

- this is what I propose as unit-test:

image

- This is our current  "unit?" tests for the client package:
image

- And this is what I understand it would be a unit-test for the client package:

image

- And this is an even better design, and unit-test diagram for the client package IMO:

image

- And this is the ideal unit-test for example for the "kube" pkg ( have not precisely verified the current code there but... "ideal"):
image

If you reached this point, again, Thanks for reading!

nicob87:
Equivalent changes where added in other pr.

PR #244, opened 2020-09-04
cli: cmdExpose returns error if SeriviceInterfaceInspect returns error.

nicob87:
this bug was detected when developing/executing #239  since that is in review probably for some more days (or will be refactored), pushing the bugfix isolated.


ajssmith (approved):
thanks

PR #245, opened 2020-09-06
Test covers 13% of pkg/qdr mock environment only

mgoulish:
I believe this tests everything in pkg/qdr that can be tested without having a live cluster.
It gets 13.4% coverage.
I will do live-cluster tests in a separate file, in another PR.

### DIFF
--- a/pkg/qdr/amqp_mgmt_mock_test.go
+++ b/pkg/qdr/amqp_mgmt_mock_test.go
@@ -1,0 +1,160 @@
+package qdr
+
+import (
+	"strings"
+	"testing"
+        "flag"
+
+	"crypto/tls"
+	"gotest.tools/assert"
+)
+
+var clusterRun = flag.Bool("use-cluster", false, "run tests against a configured cluster")
+
+
+func TestQDR(t *testing.T) {
+
+	// AsInt -----------------------------------
+	goodAsInt := Record{}
+	goodAsInt["int8"] = int8(0)
+	goodAsInt["int16"] = int16(0)
+	goodAsInt["int32"] = int32(0)
+	goodAsInt["int64"] = int64(0)
+
+	goodAsInt["uint8"] = uint8(0)
+	goodAsInt["uint16"] = uint16(0)
+	goodAsInt["uint32"] = uint32(0)
+	goodAsInt["uint64"] = uint64(0)
+
+	for key, value := range goodAsInt {
+		_, ok := AsInt(value)
+		assert.Assert(t, ok, "AsInt failed on type |%s|.", key)
+	}
+
+	badAsInt := Record{}
+	badAsInt["float32"] = float32(0)
+	badAsInt["float64"] = float64(0)
+	badAsInt["string"] = "0"
+
+	for key, value := range badAsInt {
+		_, ok := AsInt(value)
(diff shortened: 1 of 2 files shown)


pkg/qdr/amqp_mgmt_mock_test.go
```diff
@@ -0,0 +1,159 @@
...
+	goodAsInt["int64"] = int64(0)
+
+	goodAsInt["uint8"] = uint8(0)
+	goodAsInt["uint16"] = uint16(0)
+	goodAsInt["uint32"] = uint32(0)
+	goodAsInt["uint64"] = uint64(0)
+
+	for key, value := range goodAsInt {
+		_, ok := AsInt(value)
+		assert.Assert(t, ok, "AsInt failed on type |%s|.", key)
```
nicob87, line 30: you could assert also that the returned value is what you expect, (cero in this case) and or, some other value... since 0  is usually the default

pkg/qdr/amqp_mgmt_mock_test.go
```diff
@@ -0,0 +1,159 @@
...
+	badAsInt["string"] = "0"
+
+	for key, value := range badAsInt {
+		_, ok := AsInt(value)
+		assert.Assert(t, !ok, "AsInt succeeded on type |%s|, but should fail.", key)
+	}
+
+	// I guess we can at least test that these next few don't crash...
+	badAsInt.AsInt("string")
+	asConnection(badAsInt)
```
nicob87, line 45: you can build a particular record, and assert that asConnection returns a Connection object equal to what you expect ... 
This is just an example of comparing structs, not sure if it is the best way in go or there is another one ... 
https://github.com/skupperproject/skupper/blob/056b6020170a0cca12551cc3c8fded0fb5fe5bf7/cmd/skupper/skupper_test.go#L250

pkg/qdr/amqp_mgmt_mock_test.go
```diff
@@ -0,0 +1,159 @@
...
+	badAsInt["float64"] = float64(0)
+	badAsInt["string"] = "0"
+
+	for key, value := range badAsInt {
+		_, ok := AsInt(value)
+		assert.Assert(t, !ok, "AsInt succeeded on type |%s|, but should fail.", key)
+	}
+
+	// I guess we can at least test that these next few don't crash...
+	badAsInt.AsInt("string")
```
nicob87, line 44: Actually you are exposing here that the "AsInt" method, is a method, but it does not use the object its belong ... you can assert it returns 0... which is what it must return in case of error I think ...

nicob87 (commented):
Other thing exposed in this pr, is that "bool" is used to return error by this "api" instead of the standard golang "error" type

pkg/qdr/amqp_mgmt_mock_test.go
mgoulish: Very good point about the use of bool rather than (more informative) error.

These fns are not used in very many places -- I guess I would vote to change them right now to use more golang-standard behavior.   (Which would also require changes to my tests.)

I will go ahead and make that change and then we can see if we like it.

ajssmith (approved):

PR #246, opened 2020-09-07
Token cost annotation

grs:


fgiorgetti:
Looks good.

ajssmith (approved):
lgtm

client/connector_token_create.go
```diff
@@ -61,14 +64,17 @@ func configureHostPortsFromRoutes(result *RouterHostPorts, cli *VanClient) (bool
 	}
 }
 
-func configureHostPorts(result *RouterHostPorts, cli *VanClient) bool {
-	ok, err := configureHostPortsFromRoutes(result, cli)
+func configureHostPorts(result *RouterHostPorts, cli *VanClient, namespace string) bool {
+	if namespace == "" {
+		namespace = cli.Namespace
+	}
+	ok, err := configureHostPortsFromRoutes(result, cli, namespace)
```
nicob87, line 71: Just a question: Is this an implicit way of asking if this is an ocp or a kubernetes cluster? (and configuring hostports in case of ocp).
Or it also may happen that being an ocpcluster, you want to continue the execution below?

grs: At present if routes are supported, we use them. Otherwise we fallback to defining the service with type LoadBalancer. (However the change above is really only ensuring that the right namespace is used, i.e. the namespace the token request was created in).

nicob87: > (However the change above is really only ensuring that the right namespace is used, i.e. the namespace the token request was created in).
correct!
ack!

Issue #227, opened 2020-08-25
Consistency on skupper cli positional arguments error messages, and improvements.


nicob87:
IMO: there are some inconsistencies and a couple of things that may be improved, having the unittest it is very easy to just change the unit-test to what you want to happen and then refactor the code in order to make the test succeed (TDD!!)

i.e:
```
»·······assert.Error(t, b([]string{"one", "resource/name", "three"}), "extra argument: three")                                            
»·······assert.Error(t, b([]string{"one", "resource/name", "three", "four"}), "illegal argument: four")
```
if you call some commands with one extra argument you get an "extra argument" error, but if you call it with two, you get "illegal argument" error. This has no sense ... if I understand correctly.

moreover that, we support ":" and "/" notation but the "usage" message does not mention it.
 ```
[nbrignon@localhost skupper (nb-skupper-cli-testing-basic)]$ ./skupper bind
Error: Service name, target type and target name must all be specified (e.g. 'skupper bind <service-name> <target-type> <target-name>')
Usage:
  skupper bind <service-name> <target-type> <target-name> [flags]
```
TODOS:
1) In orther to improve this, the cobra command already has some helpers:
```
  func OnlyValidArgs(cmd *Command, args []string) error {                                                                                   
  func ArbitraryArgs(cmd *Command, args []string) error {                                                                                   
  func MinimumNArgs(n int) PositionalArgs {                                                                                                 
  func MaximumNArgs(n int) PositionalArgs {
```
use this helpers when possible.

2) Decide what to do with the "usage" messages:
I see the kubectl (i.e.) just does not print the usage message and only prints the error.
```
[nbrignon@localhost ~]$ kubectl delete
error: You must provide one or more resources by argument or filename.
Example resource specifications include:
   '-f rsrc.yaml'
   '--filename=rsrc.json'
   '<resource> <name>'
   '<resource>'

```

just as a note, the other issue with the usage is that: we support ":" and "/" notation but the "usage" message does not mention it.

3) Fix the "illegal" vs "extra" issue, to print always the same message.
4) Unify the error style? I see depending on how you call the command (with bad argument) sometimes the error contains the "usage" error, and sometime it doesn't.

ajssmith:
@nicob87 could you link your PRs? Can this issue be closed for now?

nicob87:
do we have  A backlog or TODO list?, the described behaviour in the issue is still the same. We can close also, but I think it is better to label it as backlog or something like that ...  what do you think?

[...]

PR #247, opened 2020-09-09
Using cobra helper for positional single arg.

nicob87:
Specific information about command "usage" is provided by the usage message itself. that is printed right below the positional argument error:

```

$ ./skupper connection-token
Error: accepts 1 arg(s), received 0
Usage:
  skupper connection-token <output-file> [flags]

```

TODO:
evaluate the other cases bind/unbind expose/unexpose serviceCreate... etc, the fact that we support ":" and "/" notations add a little bit of complexity to reusing cobra helpers... probably take a look to how kubectl handles that is a good idea.

related to #227


ajssmith (approved):
lgtm

PR #248, opened 2020-09-09
modifying cmd init msg

nicob87:
Cosmetic:
```

$ ./skupper init --help | head -n 2
Setup a router and other supporting objects to provide a functional skupper
installation that can then be connected to other skupper installations
```
Modified with the idea to match kubectl, i.e:
```

$ kubectl annotate --help | head -n 2
Update the annotations on one or more resources
```

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -172,8 +172,9 @@ func init() {
 	var cmdInit = &cobra.Command{
 		Use:   "init",
 		Short: "Initialise skupper installation",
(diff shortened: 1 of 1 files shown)


grs (approved):
I am surprised that cobra doesn't offer any functionality to do this automatically. (Perhaps we could define a function that splits strings at whitespace to ensure they remain under a defined length? Perhaps that is overkill at this point though).

ajssmith (approved):

PR #249, opened 2020-09-09
Use site ID to prevent self-connection

mgoulish:

  handling error cases first.

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -42,6 +42,17 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
 	return "conn" + strconv.Itoa(max)
 }
 
+func (cli *VanClient) isOwnToken(ctx context.Context, secret *corev1.Secret) (bool, error) {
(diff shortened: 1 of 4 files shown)


mgoulish:
This is a new PR because in the previous one I improperly added the new edge connectivity test, which is unrelated.  I will submit that test in a separate PR.

nicob87 (approved):

ajssmith (changes requested):
After attempt to connect to self, list-connectors return incorrect result.

$ ./skupper init
Skupper is now installed in namespace 'skupper'.  Use 'skupper status' to get more information.
$ ./skupper connection-token ~/sd/itsme.yaml
Connection token written to /home/ansmith/sd/itsme.yaml 
$ ./skupper list-connectors
There are no connectors defined.
$ ./skupper connect ~/sd/itsme.yaml
Failed to create connection:  Can't create connection to self with token |/home/ansmith/sd/itsme.yaml|
$ ./skupper list-connectors
Connectors:
    skupper-inter-router-skupper.127.0.0.1.nip.io:443 (name=conn1)

PR #253, opened 2020-09-11
Test edge connectivity counts

mgoulish:
* Use several different topologies, all with one edge router
  and one or more interior routers, to ensure that the edge
  router correctly counts how many interiors it is directly
  or indirectly connected to.


ajssmith (commented):
Great test but it is very chatty

Suggest squelching info if not verbose mode

PR #258, opened 2020-09-15
CI: using "--parallel 1" (ci vm has only 1 core anyway), and we see the live output during test execution.

nicob87:
By default "go test" when executed for multiple packages, paralellizes
each package test execution, that is why it only prints the test output
when package test execution has finished.
Otherwise it would print interleaved messages.
Since the CI vm has only 1 core, test execution is not paralellized. But
it prints no output, as it would be parallelized.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,8 @@ commands:
       - run:
           name: Run Integration Tests
           no_output_timeout: 30m
-          command: go test -tags=integration -timeout=30m -v ./test/integration/...
+          command: |
+            go test -tags=integration -v -p 1 ./test/integration/...
       - run:
           name: Run client tests in real cluster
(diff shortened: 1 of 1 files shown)


fgiorgetti:
@nicob87 There is an easy way to achieve the same, with no further maintenance
needed, which is by adding the `-p 1` parameter to `go test`. I believe the outcome
is going to be the same.

ajssmith (approved):
lgtm

nicob87:
> @nicob87 There is an easy way to achieve the same, with no further maintenance
> needed, which is by adding the `-p 1` parameter to `go test`. I believe the outcome
> is going to be the same. Example:
> `go test -tags=integration -v -p 1 ./test/integration/...`

ack!, thanks! trying -p right now

fgiorgetti (approved):

PR #259, opened 2020-09-15
Add client interface definition

ajssmith:
This patch introduces client interface to better support cli test goals.

Also:
* Use command constructors  
* Adopt RunE to return errors
* Example unexpose cmd test against k8s mock and real cluster

### DIFF
--- a/cmd/skupper/skupper_unexpose_cluster_test.go
+++ b/cmd/skupper/skupper_unexpose_cluster_test.go
@@ -37,85 +37,113 @@ func checkStringOmits(t *testing.T, got, expected string) {
 	}
 }
 
-func TestUnexposeCommandWithCluster(t *testing.T) {
+func newMockClient(namespace string) *client.VanClient {
+	return &client.VanClient{
+		Namespace:  namespace,
+		KubeClient: fake.NewSimpleClientset(),
+	}
+}
+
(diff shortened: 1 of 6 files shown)


grs (approved):

cmd/skupper/skupper_unexpose_cluster_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+
+	// along with cli these are package vars
+	namespace = "cmd-unexpose-cluster-test"
+	kubeContext = ""
+	kubeConfigPath = ""
+
+	// test cases assume running skupper, set up pre-condition
+	if *clusterRun {
```
nicob87, line 116: this is not needed right? it is done in the testClient function

ajssmith: This is needed here to set up the namespace as a pre-condition (e.g. separately from command execution)

nicob87 (approved):
This looks good!
I have only one comment:
If I am not wrong, we are still not able to test fmt.PrintX calls in the Run methods, that is something that would be a good improvement.
That is not addressed in this pr, and... AFAIK it wasn't also in any other proposal. So this is just that, a comment.

Nice Work!

fgiorgetti (approved):

PR #262, opened 2020-09-16
Test edge connectivity counts

mgoulish:
* Use several different topologies, all with one edge router
  and one or more interior routers, to ensure that the edge
  router correctly counts how many interiors it is directly
  or indirectly connected to.

### DIFF
--- a/client/edge_connectivity_test.go
+++ b/client/edge_connectivity_test.go
@@ -1,0 +1,421 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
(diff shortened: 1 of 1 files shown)


mgoulish:
Reduce chattiness a lot.

I didn't reduce it to zero though -- maybe 20x -- because this test takes so long (5 minutes!)  I am afraid it will freak people out if they don't see anything at all happening.

Now it just prints out the 5 test names as it starts each one.

mgoulish:
Closing this one in favor of #333 -- this test moved to integration framework.

PR #263, opened 2020-09-16
Use siteID to prevent self-connection

mgoulish:

  handling error cases first.

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -42,6 +42,17 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
 	return "conn" + strconv.Itoa(max)
 }
 
+func (cli *VanClient) isOwnToken(ctx context.Context, secret *corev1.Secret) (bool, error) {
+	siteConfig, err := cli.SiteConfigInspect(ctx, nil)
+	if err != nil {
+		return false, err
+	}
+	if author, ok := secret.ObjectMeta.Annotations[types.TokenGeneratedBy]; ok {
+		return author == siteConfig.Reference.UID, nil
+	}
+	return false, nil
+}
+
 func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile string, options types.ConnectorCreateOptions) (*corev1.Secret, error) {
 	secret, err := cli.ConnectorCreateSecretFromFile(ctx, secretFile, options)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


mgoulish:
This one is only different in that, if it discovers that the connection is a self-connection it does not leave it there as before -- but removes it.

client/connector_create.go
```diff
@@ -42,6 +42,17 @@ func generateConnectorName(namespace string, cli kubernetes.Interface) string {
...
+		return false, err
+	}
+	if author, ok := secret.ObjectMeta.Annotations[types.TokenGeneratedBy]; ok {
+		return author == siteConfig.Reference.UID, nil
+	}
+	return false, nil
+}
+
 func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile string, options types.ConnectorCreateOptions) (*corev1.Secret, error) {
 	secret, err := cli.ConnectorCreateSecretFromFile(ctx, secretFile, options)
```
ajssmith, line 57: Rather than create and then immediately delete a resource, would lean towards inspecting the token first and ownToken check before create.

In pkg/certs/certs.go  there is a 'GetSecretContent' that will return the data and annotations to check against as a map[string] e.g. content["skupper.io/generated-by"]

client/connector_create.go
```diff
@@ -50,6 +61,17 @@ func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile st
...
 		options.Name = secret.ObjectMeta.Name
 	}
+	ownToken, err := cli.isOwnToken(ctx, secret)
+	if err != nil {
+		return nil, fmt.Errorf("Can't check secret ownership: |%s|", err.Error())
+	}
+	if ownToken {
+		err := kube.DeleteSecret(options.Name, options.SkupperNamespace, cli.KubeClient)
+		if err != nil {
+			return nil, fmt.Errorf("Created connection to self with token |%s|! Error attempting to remove it: |%s|", secretFile, err.Error())
```
ajssmith, line 71: A vertical line common as a separator, would single quote be a better option?

ajssmith (changes requested):

client/connector_create.go
mgoulish: Ah, thanks. I should have seen that.

client/connector_create.go
mgoulish: Yes, good point.

mgoulish:
Closing this in favor of PR 271 from my fork.

PR #264, opened 2020-09-16
prevent selector update for headless service

ajssmith:
This patch inhibits the selector update for headless service. 

Verified via example https://github.com/grs/skupper-example-cockroachdb cluster formation

### DIFF
--- a/cmd/service-controller/controller.go
+++ b/cmd/service-controller/controller.go
@@ -316,7 +316,7 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
 			actual.Spec.Ports[0].TargetPort = intstr.FromInt(desired.ingressPort)
 		}
 	}
-	if !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
+	if desired.headless == nil && !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
 		update = true
 		actual.Spec.Selector = kube.GetLabelsForRouter()
 	}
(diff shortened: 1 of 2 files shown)


cmd/service-controller/controller.go
```diff
@@ -316,7 +316,7 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
 			actual.Spec.Ports[0].TargetPort = intstr.FromInt(desired.ingressPort)
 		}
 	}
-	if !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
+	if desired.headless == nil && !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
```
grs, line 319: Strictly speaking I think the test would be:

``` (desired.headless == nil || desired.origin != "") && !equivalentSelectors(...)```

i.e. we could update the service even if it is headless as long as it originated elsewhere. However the fix as is is sufficient for now and fixes a regression so :shipit:!

grs (approved):

cmd/service-controller/controller.go
grs: And thanks for fixing this @ajssmith !

ajssmith: Iiuc a headless service originating elsewhere would still have the same issue, e.g. we would not want to change the labels from proxy to router.

grs: Yes, you are correct. The labels required there would indeed be different.

PR #265, opened 2020-09-17
Simplify clusterTestRunner.

nicob87:
* testRunner.Initialize: removed
* testRunner object does not need to be global anymore
* Build renamed to BuildOrSkip since that is what it does

### DIFF
--- a/test/integration/basic/basic_test.go
+++ b/test/integration/basic/basic_test.go
@@ -4,16 +4,15 @@ package basic
 
 import (
 	"context"
-	"github.com/skupperproject/skupper/test/utils/base"
+	"os"
 	"testing"
-)
 
-var (
-	testRunner BasicTestRunner
+	"github.com/skupperproject/skupper/test/utils/base"
 )
 
 func TestMain(m *testing.M) {
-	testRunner.Initialize(m)
+	base.ParseFlags()
+	os.Exit(m.Run())
 }
 
 func TestBasic(t *testing.T) {
@@ -22,7 +21,8 @@ func TestBasic(t *testing.T) {
 		PublicClusters:  1,
 		PrivateClusters: 1,
 	}
-	testRunner.Build(t, needs, nil)
+	testRunner := &BasicTestRunner{}
+	testRunner.BuildOrSkip(t, needs, nil)
 	ctx, cancel := context.WithCancel(context.Background())
 	base.HandleInterruptSignal(testRunner.T, func(t *testing.T) {
 		testRunner.TearDown(ctx)
--- a/test/integration/http/http_test.go
+++ b/test/integration/http/http_test.go
@@ -4,22 +4,21 @@ package http
 
 import (
 	"context"
+	"os"
+	"testing"
+	"time"
+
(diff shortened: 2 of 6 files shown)


test/integration/http/http_test.go
```diff
@@ -28,7 +27,8 @@ func TestHttp(t *testing.T) {
 		PublicClusters:  1,
 		PrivateClusters: 1,
 	}
-	testRunner.Build(t, needs, nil)
+	testRunner := &HttpClusterTestRunner{}
```
fgiorgetti, line 30: Guess it should use the ClusterTestRunnerBase here, or am I missing something?

test/integration/basic/basic_test.go
```diff
@@ -22,7 +21,8 @@ func TestBasic(t *testing.T) {
 		PublicClusters:  1,
 		PrivateClusters: 1,
 	}
-	testRunner.Build(t, needs, nil)
+	testRunner := &BasicTestRunner{}
```
fgiorgetti, line 24: Maybe we don't even need a `BasicTestRunner` type, and instead
we could simply use the `ClusterTestRunnerBase`.

Same applies to the http and tcp_echo.

test/utils/base/cluster_test_runner.go
```diff
@@ -47,21 +45,17 @@ type ClusterTestRunnerBase struct {
...
 
 	// Initializing internal properties
 	c.T = t
 	c.vanClientProvider = vanClientProvider
 	c.ClusterContexts = []*ClusterContext{}
 
+	if len(TestFlags.KubeConfigs) == 0 && len(TestFlags.EdgeKubeConfigs) > 0 {
+		panic("at least one --kubeconfig must be provided when using --edgekubeconfig")
+	}
+
```
fgiorgetti, line 58: What is the benefit of moving this logic here instead of at the `ParseFlags` function?

test/integration/basic/basic_test.go
nicob87: I totally agree having a base class and "extending" was not a good idea. Propose to treat that in the next pr

test/utils/base/cluster_test_runner.go
nicob87: I seemed to me ... that when you do "parseFlags" you are not expecting to receive a panig because of "flags" interpretation.... 
you just want to parse...
anyway, it is true that probably is a change not related, and maybe not necessary, I can roll it back if you prefer

test/integration/basic/basic_test.go
fgiorgetti: Ok.

test/utils/base/cluster_test_runner.go
fgiorgetti: It makes more sense to me to keep it there, because the `Build` method can be called multiple times,
at the package level, while it just needs to be done once.

nicob87: Sure! restoring to its place

fgiorgetti (approved):

PR #267, opened 2020-09-18
integration_tests_runner: remove "T *testing.T" from base struct

nicob87:
* GetContext returns error instead of printing warning
* Just accomodating the rest of the code to previous changes
* Added unit-test for GetContext

TODO:

* remove "baseRunner" extentions on each integration test
(TcpEchoTestRunner, BasicTestRunner, etc), and just use functions.
* unify teardowns? for now teardown functions "looks like" duplicated, BUT,
that is because in the existing integration tests we are using one public,
and one private namespace/context, this is not necessarily a rule for all
tests.

### DIFF
--- a/test/utils/base/cluster_test_runner.go
+++ b/test/utils/base/cluster_test_runner.go
@@ -29,11 +29,11 @@ type ClusterTestRunner interface {
 	// Initialize ClusterContexts
 	BuildOrSkip(t *testing.T, needs ClusterNeeds, vanClientProvider VanClientProvider) []*ClusterContext
 	// Return a specific public context
-	GetPublicContext(id int) *ClusterContext
+	GetPublicContext(id int) (*ClusterContext, error)
 	// Return a specific private context
-	GetPrivateContext(id int) *ClusterContext
+	GetPrivateContext(id int) (*ClusterContext, error)
 	// Return a specific context
-	GetContext(private bool, id int) *ClusterContext
+	GetContext(private bool, id int) (*ClusterContext, error)
 }
 
 // ClusterTestRunnerBase is a base implementation of ClusterTestRunner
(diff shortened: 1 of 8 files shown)


fgiorgetti (changes requested):
Only thing that is missing here is to change the return type in the interface `type ClusterTestRunner interface` itsef, see:

```
// ClusterTestRunner defines a common interface to initialize and prepare
// tests for running against an external cluster
type ClusterTestRunner interface {
	// Initialize ClusterContexts
	BuildOrSkip(t *testing.T, needs ClusterNeeds, vanClientProvider VanClientProvider) []*ClusterContext
	// Return a specific public context
	GetPublicContext(id int) (*ClusterContext, error)
	// Return a specific private context
	GetPrivateContext(id int) (*ClusterContext, error)
	// Return a specific context
	GetContext(private bool, id int) (*ClusterContext, error)
}
```

test/utils/base/cluster_test_runner.go
```diff
@@ -44,6 +44,8 @@ type ClusterTestRunnerBase struct {
 	unitTestMock      bool
 }
 
+var _ ClusterTestRunner = &ClusterTestRunnerBase{}
```
fgiorgetti, line 47: Any particular reason for doing it?

nicob87: To force a compilation error if the ClusterTestRunnerBase type does not implements the ClusterTestRunner interface....
So, instead of you reviewing and discovering I missed to update the interface I would get a compilation error in the pipeline.

https://splice.com/blog/golang-verify-type-implements-interface-compile-time/

fgiorgetti: Ok, yes it makes sense, but maybe another approach would be to change the following declaration in the unit test to:
```
func TestGetContext(t *testing.T) {
	var c ClusterTestRunner = &ClusterTestRunnerBase{}
```

nicob87: Okey!

nicob87: nop, actually it does not work, I need access to the object internals for the test

fgiorgetti: It does, but will require type assertions when setting the internal fields.

nicob87: mmm yup, that could work... as I understand it, the test is testing the base type. I prefer to keep the test agnostic to the interface...

fgiorgetti: Ok. Not a big deal. Approving.

fgiorgetti (approved):

PR #271, opened 2020-09-23
Use siteID to prevent self-connection

mgoulish:

  handling error cases first.


mgoulish:
Andy -- I wanted to try right away with my fork...

mgoulish:
This one uses cert pkg to check the self-ownership of token before creation, so does not need to create and then delete.

mgoulish:
This latest commit fixes the immediate problem I was seeing in HTTP test -- but now I'm seeing a crash.
Investigating...

mgoulish:
Closing this to make a new. clean one.

PR #278, opened 2020-09-28
unit-tets for bind command

nicob87:
Also added verification for "bind" command "Args" function, so in case of user calling invalid type, error is returned instantaneously.


ajssmith (approved):
nice mod

PR #279, opened 2020-09-30
cli cluster tests

ajssmith:
* Use cobra command constructors

* Set up pre-conditions as needed


cmd/skupper/skupper_cluster_test.go
```diff
@@ -0,0 +1,1579 @@
...
+			expectedError:   "",
+			realCluster:     false,
+		},
+	}
+
+	namespace = "init-interior-cluster-test-" + strings.ToLower(utils.RandomId(4))
+	kubeContext = ""
+	kubeConfigPath = ""
+
+	if *clusterRun {
```
nicob87, line 203: you can just use the "testClient" function here I think ...

cmd/skupper/skupper_cluster_test.go
```diff
@@ -0,0 +1,1579 @@
...
+	} else {
+		cli = newMockClient(namespace)
+	}
+
+	if c, ok := cli.(*client.VanClient); ok {
+		_, err := kube.NewNamespace(namespace, c.KubeClient)
+		assert.Check(t, err)
+		defer kube.DeleteNamespace(namespace, c.KubeClient)
+	}
+	skupperInit(t, []string{"--edge"}...)
```
nicob87, line 268: if I understand correctly, the "delete" table case only works once, because skupper init is executed only once

cmd/skupper/skupper_cluster_test.go
```diff
@@ -0,0 +1,1579 @@
...
+
+func TestConnectWithEdgeCluster(t *testing.T) {
+	testcases := []struct {
+		doc             string
+		args            []string
+		expectedCapture string
+		expectedOutput  string
+		expectedError   string
+		realCluster     bool
+	}{
```
nicob87, line 417: just a comment: you could define a testCase type just private for this pks_test and save this boilerplate on every test

cmd/skupper/skupper_cluster_test.go
nicob87: and in all the other tests...

cmd/skupper/skupper_cluster_test.go
nicob87: which for now, does not produce any error

nicob87 (approved):
Nice ans useful work!

Just another cosmetic comment: this 10 lines are repeated in almost all or all tests

```

if *clusterRun {
		cli = NewClient(namespace, kubeContext, kubeConfigPath)
	} else {
		cli = newMockClient(namespace)
	}

	if c, ok := cli.(*client.VanClient); ok {
		_, err := kube.NewNamespace(namespace, c.KubeClient)
		assert.Check(t, err)
		defer kube.DeleteNamespace(namespace, c.KubeClient)

	}
```

cmd/skupper/skupper_cluster_test.go
ajssmith: That is correct. If needed for coverage, init could be moved to the test loop in a future patch.

cmd/skupper/skupper_cluster_test.go
ajssmith: Good suggestion, will change

PR #280, opened 2020-10-02
Restore connector tests

mgoulish:
* rename: self_connection test is really a connection
  creation test, so put it in connector_create_test.

* rewrite: For all other connector tests that were
  relying on self-connection, rewrite them to use two
  separate namespaces: one that creates a token, and
  one that uses the token to make a connection.

* rewrite: Factor out common code into a couple helper
  functions, and put them in connector_create_test file.

* cleanup: in connector_inspect.go, avoid nesting by
  handling errors first.


mgoulish:
Please don't do anything with this yet -- I will be debugging it here for a while -- because I can still only get it to fail in CI.

ajssmith (approved):
lgtm

PR #281, opened 2020-10-06
default simple setup for integration tests

nicob87:
Currently tcp_echo and http integration tests needs exactly the same skupper setup (1 public and 1 private connected sites). Also the incomming bookinfo example will use the same skupper setup.


fgiorgetti:
@nicob87 looks good. Maybe in the future we can have these kind of implementations as a `Topology` interface,
in a separate package, so we can provide different reusable topologies.

fgiorgetti (approved):

nicob87:
> @nicob87 looks good. Maybe in the future we can have these kind of implementations as a `Topology` interface,
> in a separate package, so we can provide different reusable topologies.

Totally agree!

ajssmith (approved):

PR #287, opened 2020-10-07
Fix expose address

ajssmith:
* should be target name


grs (approved):

grs:
BTW, the expose with target type 'service' is different from annotating the service. Using expose will cause the target of a new service to the old service.

Issue #293, opened 2020-10-09
'skupper connect' panics when router not running


ajssmith:
./skupper connect token-file.yaml will panic if skupper has not been initialized. It used to detect...

PR #294, opened 2020-10-09
Check Skupper is running before checking token

mgoulish:
This is for issue #293 :  'skupper connect' panics when router not running


mgoulish:
Andy -- the test failure is happening at the end in the minikube tests and it does not look to me as though it's related to this code.
Please see if you agree.

ajssmith:
Patch needs to be rebased to fix ci failure

PR #296, opened 2020-10-09
Annotated Resources Integration Test

### LINKED ISSUE #130: Test to validate skupper reaction to annotated resources

fgiorgetti:
* TestAnnodatedResources new integration test added
* Curl wrapper (to run curl on existing pods and avoid extra deployments)
* Refactored to use new ClusterTestRunner
* Adjusted parsing of DATA endpoint to retry when receiving 500 internal error
* Updated circleci timeout when running integration tests
* Resolves #130

### DIFF
--- a/test/integration/annotation/annotated_resource.go
+++ b/test/integration/annotation/annotated_resource.go
@@ -1,0 +1,288 @@
+package annotation
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/client"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"github.com/skupperproject/skupper/test/utils/base"
+	"github.com/skupperproject/skupper/test/utils/constants"
+	"gotest.tools/assert"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	timeout  = 120 * time.Second
+	interval = 20 * time.Second
+)
+
+var (
+	ctx, cancelFn = context.WithTimeout(context.Background(), constants.ImagePullingAndResourceCreationTimeout)
+)
+
+// test allows defining the matrix to run the test table
+type test struct {
+	name                  string
+	doc                   string
+	modification          func(*testing.T, base.ClusterTestRunner)
+	expectedSites         int
+	expectedServicesProto map[string]string
+}
+
+// Setup simply creates the namespaces using a single
+// cluster by default, or using two clusters if at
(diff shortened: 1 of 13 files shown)


grs (approved):
:shipit: Looks good!

test/integration/annotation/annotated_resource.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Set annotations to the service with target address
+	svcTargetAnnotations[types.ProxyQualifier] = "http"
+	svcTargetAnnotations[types.AddressQualifier] = fmt.Sprintf("nginx-%d-svc-exp-target", clusterIdx)
+}
+
+// CreateVan deploys Skupper to both clusters then generates a
+// connection token on cluster1 and connects cluster2 to cluster1
+// waiting for the network to be formed (or failing otherwise)
+func CreateVan(t *testing.T, testRunner base.ClusterTestRunner) {
```
nicob87, line 153: Just a comment:
This seems to be a duplication of:
```
func SetupSimplePublicPrivateAndConnect(ctx context.Context, r *ClusterTestRunnerBase, prefix string) error {
```

maybe you can unify them! in favor of any, the new one or the old one... I see in this implementation you are waiting for the connection to be stablished which is better IMO....

The only difference is that the other function also creates the namespaces so ... you will have to change the order to make it work (first setup skupper and then deploy the test deployments...) not sure if that causes any issue...

nicob87 (approved):
Nice!

test/integration/annotation/annotated_resource.go
```diff
@@ -0,0 +1,285 @@
...
+	assert.Assert(t, err)
+
+	// creating namespaces
+	assert.Assert(t, pub.CreateNamespace())
+	assert.Assert(t, prv.CreateNamespace())
+}
+
+// Teardown ensures all resources are removed
+// when test completes or is interrupted
+func TearDown(t *testing.T, testRunner base.ClusterTestRunner) {
```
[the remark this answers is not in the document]
fgiorgetti: Ok, I will try it.

test/integration/annotation/annotated_resource.go
fgiorgetti: I see. In this particular test I cannot do it this way, because the first test iteration expects that
all resources (deployment and services) are defined before Skupper is initialized in the namespace.

But I created #309 so that we can unify this process in a separate PR, and try to adjust the
`SetupSimplePublicPrivateAndConnect` so it can be used as a generic/common function to all
tests that use a single private and public namespaces.

Let me know what you think.

nicob87: Sounds good!

PR #300, opened 2020-10-15
Fix panic when checking token #293

mgoulish:
* Make sure that Skupper is running before
    checking token.

  * Make connector create test use real cluster
    so it gets expected error instead of "no router"
    error.

  * Unique the results in TestConnectorCreateInterior
    to avoid errors caused by multiple reports of
    same connector.

### DIFF
--- a/client/connector_create_test.go
+++ b/client/connector_create_test.go
@@ -19,24 +19,37 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+var lightRed string = "\033[1;31m"
+var resetColor string = "\033[0m"
+
 func TestConnectorCreateError(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	cli, err := newMockClient("my-namespace", "", "")
-	assert.Assert(t, err)
+	var cli *VanClient
+	var err error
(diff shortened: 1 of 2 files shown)


client/connector_create_test.go
```diff
@@ -19,24 +19,37 @@ import (
...
-	cli, err := newMockClient("my-namespace", "", "")
-	assert.Assert(t, err)
-
-	_, err = cli.ConnectorCreateFromFile(ctx, "./somefile.yaml", types.ConnectorCreateOptions{
-		Name: "",
-		Cost: 1,
+	var cli *VanClient
+	var err error
+	ns := "namespace-for-testconnectorcreateerror"
+	cli, err = NewClient(ns, "", "")
```
ajssmith, line 32: Does this need to run against cluster? Either setup mock or cluster client as is done in setupTwoNamespaces...

mgoulish: I was going to say 'yes' -- but I just ran full test suite against real-cluster and mock and it looks like the correct answer is 'no'!
This test passed both ways after I put in the code for checking clusterRun -- so I guess we can leave it in there.

ajssmith (approved):

Issue #299, opened 2020-10-14
Service annotated only with protocol (without address qualifier) stops working if protocol changes


fgiorgetti:
A service that is annotated only with: `skupper.io/proxy: tcp` works normally when skupper is
initialized in the given namespace.

But if the protocol changes from `tcp` to `http`, it stops working.

If the same update happens on services that have the `skupper.io/address` annotation,
they continue to work normally after the modification.

PR #301, opened 2020-10-15
Allows modifying annotations for already exposed service

fgiorgetti:
Fixes issue #299 allowing already exposed services to switch their
protocols. Previously the skupper-internal config map was being incorrectly
modified to use incorrect egress port and host.

When the annotations of an already exposed service was modified, the egress
port was being derived based on already modified target port (from updated
service), as it was already pointing to the bridge listener port. And the
egress host was also being incorrectly evaluated as being the "router" ip
(as the updated selector was matching the router pods).


fgiorgetti:
@ajssmith @grs One question about this proposed PR. In my test scenario, I am creating 
a service that has the `skupper.io/proxy` annotation only.

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    skupper.io/proxy: tcp
  labels:
    app: nginx
  name: nginx-1-svc-exp-notarget
  namespace: public-annotated-1
spec:
  ports:
  - name: web
    port: 8080
    targetPort: 8080
  selector:
    app: nginx
  type: LoadBalancer
```

When I do so, Skupper modifies the selector so traffic is redirected to the bridge.
If I change the protocol from `tcp` to `http` it stops working, as the `skupper-internal` cm is
updated based on the `bridge` selector info and so the service stops working.

This PR ignores parsing the modified selector again, as the service is already updated
to use `skupper.io/component=router` as the selector, so it uses cached information instead
(which still has the correct target port).

The change works for modified services, but if I remove the `skupper.io/proxy: tcp` annotation
the selector is not restored, and the service becomes unresponsive (as it continues to target
the bridge, which no longer exists).

Well, looking back at Skupper CLI, I noticed that the expose command has the target address
as a required option and I started thinking if this kind of annotation (without a target address) is
still valid/supported.

Using skupper 0.3.0 (latest images from quay) it works, as in this particular scenario the selector
is not modified for the `exposed` service.

What do you guys recommend? Should this kind of annotated service (without the target address)
be still supported, or not? If yes, should we keep the original selector in a separate annotation, so
if it is unexposed, or the annotation is removed the original selector is restored?

grs:
The behaviour is very different with and without the target annotation but both use cases are in my view still valid. I do feel a degree of discomfort about the annotated objects themselves being mutated though, so perhaps worth some discussion for the longer term. 

The fact that the controller cannot tell that the port has been modified already is a design defect. It is similar to the issue with the selector, but there at least you can workaround the problem as the skupperized selector will always be the same. I think ideally when modifying the service itself, skupper should add explicit annotations indicating the original values as a more robust short term fix.

fgiorgetti:
@grs, do you think that adding something in
`service-controller/controller.go`, around:

```
if desired.headless == nil && !equivalentSelectors(actual.Spec.Selector,
kube.GetLabelsForRouter()) {

   update = true
   actual.Spec.Selector = kube.GetLabelsForRouter()
}

```

... to save the original selector in a new annotation and using it

everytime the service is updated (from skupper perspective) would work?

grs:
Yes, also for the original port I think.

fgiorgetti:
Exactly what I was going to ask next. :) Thanks!
I will try adding both and you guys can review/evaluate if worth merging.

fgiorgetti:
The new integration test that validates this fix is not ready yet, but I have
some scripts that can be used to manually validate the same scenarios.
If someone needs/wants them, I can try to share somehow.

cmd/service-controller/controller.go
```diff
@@ -309,15 +320,24 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
 	if len(actual.Spec.Ports) > 0 {
 		if actual.Spec.Ports[0].Port != int32(desired.publicPort) {
 			update = true
+			if !hasOriginalPort(*actual) {
+				actual.ObjectMeta.Annotations[types.OriginalPortQualifier] = strconv.FormatInt(int64(actual.Spec.Ports[0].Port), 10)
```
grs, line 324: Do we ever modify the port of the service? (I don't think so).

cmd/service-controller/controller.go
```diff
@@ -309,15 +320,24 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
...
+			}
 			actual.Spec.Ports[0].Port = int32(desired.publicPort)
 		}
 		if actual.Spec.Ports[0].TargetPort.IntValue() != desired.ingressPort {
 			update = true
+			if !hasOriginalTargetPort(*actual) {
```
grs, line 330: Should we always update the originalTargetPort when modifying the current value? E.g. service is created and annotated, skupper modifies it including an update to the target port, user now updates the service themselves including a change to the target port, skupper will again update it, but now if restored it would set it back to the very first port.

It does seem simpler to reason about perhaps to always just save the last version we overwrote?

cmd/service-controller/definition_monitor.go
```diff
@@ -224,8 +225,15 @@ func (m *DefinitionMonitor) getServiceDefinitionFromAnnotatedService(service *co
 				Name:     service.ObjectMeta.Name,
 				Selector: utils.StringifySelector(service.Spec.Selector),
 			}
-			if targetPort := deduceTargetPortFromService(service); targetPort != 0 {
-				target.TargetPort = targetPort
+			if curSvc, ok := m.annotated[svc.Address]; !ok {
```
grs, line 228: I *think* what we want to do here is check whether there is an originalTargetPort annotation, and if so use that as the targetPort on the target?

cmd/service-controller/controller.go
```diff
@@ -309,15 +320,24 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
...
+				actual.ObjectMeta.Annotations[types.OriginalTargetPortQualifier] = strconv.FormatInt(int64(actual.Spec.Ports[0].TargetPort.IntVal), 10)
+			}
 			actual.Spec.Ports[0].TargetPort = intstr.FromInt(desired.ingressPort)
 		}
 	}
 	if desired.headless == nil && !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
 		update = true
+		if !hasOriginalSelector(*actual) {
+			actual.ObjectMeta.Annotations[types.OriginalSelectorQualifier] = utils.StringifySelector(actual.Spec.Selector)
+		}
```
grs, line 340: As above for targetPort, perhaps we should always just record the original selector when we change it.

cmd/service-controller/bridges.go
```diff
@@ -216,9 +217,14 @@ func (c *Controller) updateServiceBindings(required types.ServiceInterface, port
 		} else if bindings.headless != nil {
 			bindings.headless = nil
 		}
+
+		hasSkupperSelector := false
 		for _, t := range required.Targets {
 			targetPort := getTargetPort(required, t)
-			if t.Selector != "" {
+			if strings.Contains(t.Selector, "skupper.io/component=router") {
```
grs, line 224: I don't think the target definition should ever have `skupper.io/component=router` in it. I.e. I suspect that we want to prevent that happening in the first place?

cmd/service-controller/controller.go
fgiorgetti: I do not think so. Should I remove this (as well as the "OriginalPortQualifier")?

cmd/service-controller/controller.go
fgiorgetti: That is a good point. If user modifies the target port, so that the OriginalTargetPort annotation
exists but the "new" target port is different than the current one, it makes sense to me to update
the value of the OriginalTargetPort annotation.

cmd/service-controller/definition_monitor.go
fgiorgetti: Yes, now that we have it in the service, I think I can change it to validate based on the new annotation.

cmd/service-controller/controller.go
fgiorgetti: That is also correct, I will change it.

cmd/service-controller/bridges.go
fgiorgetti: Well, that is not what is currently happening. If you use latest image from quay (0.3),
the selector isn't changing, as you say. But if you build a fresh image from master,
then the selector is changing when you add `skupper.io/proxy` to an existing service
without specifying `skupper.io/address` (it is the actual initial reason for #299).

cmd/service-controller/controller.go
grs: I think it would be clearer not to record an 'original port' if we never change it.

cmd/service-controller/bridges.go
grs: I think this check - or an equivalent - should be happening in the definition_monitor.go, to prevent the service definition having the wrong selector written in to it in the first place.

I.e. in getServiceDefinitionFromAnnotatedService, where the selector is not nil and the target annotation is not set, we should check whether the selector is already the router selector, if not that is the selector to use. If it is already the router selector then we should use the original selector annotation value instead. (I think!)

cmd/service-controller/controller.go
fgiorgetti: I believe that won't work. In example, if we change the protocol, the `desired.ingressPort`
will change and will be different from `actual.Spec.Ports[0].TargetPort`, so updating the
`originalTargetPort` will make it unrecoverable.
Maybe if that is not too much, we can check if protocol has not changed, then in this case
I believe it makes sense to update the `originalTargetPort`. What do you think?
not changed?

grs: Stepping back, I think the root of the problem - for the case where target is not specified and therefore where we are editing the service that we are also using as the input into the definition - is that we need to be able to clearly distinguish changes made by the user from those made by the controller.

So perhaps what we could do is have two annotations per field of conflict. E.g. in this case originalTargetPort and assignedTargetPort. When the controller edits the TargetPort on the service, it first checks whether the current value matches the current assignedTargetPort annotation. If it does not, then it updates the originalTargetPort. Either way it then updates the service TargetPort and the assignedTargetPort.

The definition monitor can then also ignore the TargetPort if it matches the assignedTargetPort.

We could use a similar scheme for the selector. Does that make sense? Does it sound easy to understand and reason about?

cmd/service-controller/bridges.go
fgiorgetti: Gotcha! It makes sense to me (if I understood correctly). Let me implement it and I
will let you know once code has been updated and tested locally.

grs: Thanks Fernando! Sorry for the chopping and changing. I had not properly appreciated the complexities in this case.

cmd/service-controller/controller.go
fgiorgetti: Lets say the service is annotated with `skupper.io/proxy=tcp` only, with `spec.selector.targetPort=8080`.

Skupper will update it and, lets say, update target port to `1024`. When doing so, it will set
`originalTargetPort` to 8080 and `assignedTargetPort to `1024`.

Then user switches the protocol to `http`, controller will update the assigned target port to (i.e.) `1025`
and then it will differ from the `assignedTargetPort`.

Maybe we should store the `originalProtocol` or `originalProxy` instead and update the target port when
it has been modified? If the protocol has not changed, but the target port is now different, then we should
update the `originalTargetPort` and update it (as this would mean a user change and not a controller change).

Does it make sense?

fgiorgetti: Oh, so you mean comparing the *actual* target port (defined in the service) with the `assignedTargetPort` annotation? Correct, it makes sense. Sorry, but I guess I did not get it right earlier.
It seems correct to me and I will change/test it.

cmd/service-controller/controller.go
fgiorgetti: Removed

cmd/service-controller/controller.go
fgiorgetti: Done. It seems to be working just fine now.

cmd/service-controller/bridges.go
fgiorgetti: Come on, I am the one that have to thank you for your, always, valuable advices :+1:

cmd/service-controller/definition_monitor.go
```diff
@@ -132,7 +133,10 @@ func updateAnnotatedServiceDefinition(actual *types.ServiceInterface, desired *t
 		return true
 	}
 	if len(desired.Targets) > 0 {
-		if actual.Targets[0].Name != actual.Targets[0].Name || actual.Targets[0].Selector != actual.Targets[0].Selector {
```
fgiorgetti, line 135: This if statement was also incomplete and not actually validating anything.

cmd/service-controller/controller.go
```diff
@@ -313,11 +329,21 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
...
+			if actualTargetPort != originalAssignedPort {
+				actual.ObjectMeta.Annotations[types.OriginalTargetPortQualifier] = strconv.Itoa(actualTargetPort)
+			}
+			actual.ObjectMeta.Annotations[types.OriginalAssignedQualifier] = strconv.Itoa(desired.ingressPort)
 			actual.Spec.Ports[0].TargetPort = intstr.FromInt(desired.ingressPort)
 		}
 	}
 	if desired.headless == nil && !equivalentSelectors(actual.Spec.Selector, kube.GetLabelsForRouter()) {
 		update = true
+		if !hasOriginalSelector(*actual) || !hasRouterSelector(*actual) {
```
grs, line 344: The preceding if statement will only be true if the selector is not the router selector. So is there any value in the additional test for !hasRouterSelector() here? Also why would we not update the originalSelector annotation even if it already exists, since presumably the selector has somehow been changed again by the user in that case?

fgiorgetti: That is right and odd (not sure why I left it there). I will remove and double-check.
Thanks for pointing it out.

grs (approved):

cmd/service-controller/controller.go
```diff
@@ -313,11 +329,25 @@ func (c *Controller) checkServiceFor(desired *ServiceBindings, actual *corev1.Se
 		}
 		if actual.Spec.Ports[0].TargetPort.IntValue() != desired.ingressPort {
 			update = true
+			originalAssignedPort, _ := strconv.Atoi(actual.Annotations[types.OriginalAssignedQualifier])
+			actualTargetPort := actual.Spec.Ports[0].TargetPort.IntValue()
+			if actual.ObjectMeta.Annotations == nil {
```
fgiorgetti, line 334: @grs added one extra commit that simply initializes the actual service's annotations map.
This was needed because, when I ran the refactored version of the annotated resources integration test
repeatedly, it was sporadically failing.

Looking at the service-controller's log, I noticed that when it happened the pod was restarted because of a panic caused here, due to annotation map being nil.

After applying this fix, I ran the refactored integration test 100 times in a row and all attempts have passed.

Issue #305, opened 2020-10-26
Translate edge connectivity test to integration test framework.


mgoulish:
The edge connectivity test was done some time ago -- in the unit test framework -- but is still in the process of being translated to the integration test framework.

ajssmith:
https://github.com/skupperproject/skupper/pull/333

PR #325, opened 2020-11-20
Move edge connectivity test to integration(#305)

mgoulish:
* Introduce edge_connectivity directory and
  two files.

* Look at how many interior nodes the single
  edge node can see, both directly and
  indirectly. Make sure the count is correct.

### DIFF
--- a/test/integration/annotation/annotated_resource_test.go
+++ b/test/integration/annotation/annotated_resource_test.go
@@ -172,7 +172,7 @@ func TestAnnotatedResources(t *testing.T) {
 					// reaching service through proxy-controller's pod (with some attempts to make sure bridge is connected)
 					err = utils.Retry(backoff.Duration, backoff.Steps, func() (bool, error) {
 						endpoint := fmt.Sprintf("http://%s:8080", svc)
-						resp, err = tools.Curl(cluster.KubeClient, cluster.RestConfig, cluster.Namespace, "", endpoint, tools.CurlOpts{Timeout: 15})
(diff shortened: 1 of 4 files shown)


mgoulish:
I am having a timeout problem that I need some help with. The final test -- number 5 in the list -- runs fine if I run it alone, but always times out at 10m0s if I run all of them. I have changed all the timeouts I can find, but it keeps happening -- always at 10m0s.

mgoulish:
Aha! I see how to solve the timeout problem.
I could not find it in the code, because it is not in the code.
It works if I run it this way:
go test -v -count=1 -tags integration ./test/integration/edge_connectivity **-timeout 15m**

Is there some way that we can get it to run this way in the automated testing?

fgiorgetti:
> Aha! I see how to solve the timeout problem.
> I could not find it in the code, because it is not in the code.
> It works if I run it this way:
> go test -v -count=1 -tags integration ./test/integration/edge_connectivity **-timeout 15m**
> 
> Is there some way that we can get it to run this way in the automated testing?

@mgoulish in Circle CI it's been executed as:
`go test -tags=integration -timeout=60m -v -p 1 ./test/integration/...`

Do you think it is OK as is?

mgoulish:
Yes, -timeout=60m seems fine to me, I was just running it by hand with the
default -timeout=10m, and it wasn't enough.

mgoulish:
Closing this one in favor of #333

PR #327, opened 2020-11-24
http2 integration test

nicob87:


.circleci/config.yml
```diff
@@ -84,7 +84,7 @@ commands:
           name: Run Integration Tests
           no_output_timeout: 60m
           command: |
-            go test -tags=integration -timeout=60m -v -p 1 ./test/integration/...
+            go test -tags=integration -timeout=60m -v -p 1 ./test/integration/http/...
```
fgiorgetti, line 87: @nicob87 guess this needs to be restored

test/integration/http/http_test.go
```diff
@@ -38,13 +39,23 @@ func TestHttp(t *testing.T) {
 }
 
 func TestHttpJob(t *testing.T) {
+	testHttpJob(t, "http://httpbin:8080/")
+	testHttpJob(t, "http://nghttp2:8443/")
```
fgiorgetti, line 43: Is this second URL supposed to be here or moved inside `TestHttp2Job`?

test/integration/http/http_test.go
```diff
@@ -60,4 +71,6 @@ func TestHttpJob(t *testing.T) {
 
 	// Success is the percentage of non-error responses.
 	assert.Assert(t, metrics.Success > 0.95, "too many errors! see the log for details.")
+
+	fmt.Printf("Success!!\n")
```
fgiorgetti, line 75: One thing we might need to evaluate is adding or creating some mechanism
to handle logging levels, and stop using `Printf`. If you agree, let's open a separate
issue to take care of that later.

fgiorgetti (commented):
I will play with it locally, but overall it looks good to me.
Please just go through a few comments I have added here.

test/integration/http/http_test.go
nicob87: yup! but will be removed, good catch.

test/integration/http/http_test.go
nicob87: yup! agree!
just one comment, the fact that this is a test already gives you some logging control, you only see the printf if you use "-v" for go test.

.circleci/config.yml
nicob87: correct!

test/integration/http/http.go
```diff
@@ -69,33 +69,84 @@ var httpbinDep *appsv1.Deployment = &appsv1.Deployment{
...
+		assert.Assert(t, err)
+		cc.KubectlExec("logs job/" + jobName)
+		k8s.AssertJob(t, job)
+	}
+
+	runJob(pubCluster1, "http1", "TestHttpJob")
+	waitJob(pubCluster1, "http1")
+
+	runJob(pubCluster1, "http2", "TestHttp2Job")
+	waitJob(pubCluster1, "http2")
```
fgiorgetti, line 149: @nicob87 test is passing locally for me, but I believe you could add a sub-test here,
when running http1 and http2 jobs. Something like:
```
	t.Run("http1", func(t *testing.T) {
		runJob(pubCluster1, "http1", "TestHttpJob")
		waitJob(pubCluster1, "http1")
	})

	t.Run("http2", func(t *testing.T) {
		runJob(pubCluster1, "http2", "TestHttp2Job")
		waitJob(pubCluster1, "http2")
	})
```

The benefit is that the generated output (and eventual junit.xml) will have results
defined separately for each http1 and http2 tests.

nicob87: makes totally sense

fgiorgetti (approved):

ajssmith (approved):
lgtm

PR #330, opened 2020-11-30
Issue 329: version 4 nodes cannot connect to version 3

mgoulish:
* When encountering secret that has no 'generated-by'
  annotation, issue warning but do not fail.

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -3,8 +3,8 @@ package client
 import (
 	"context"
 	"fmt"
+	"github.com/prometheus/common/log"
 	"io/ioutil"
-	"log"
 	"regexp"
 	"strconv"
 
@@ -50,7 +50,7 @@ func secretFileAuthor(ctx context.Context, secretFile string) (author string, er
 	}
 	generatedBy, ok := content["skupper.io/generated-by"]
 	if !ok {
-		return "", fmt.Errorf("Can't find secret origin.")
+		return "unknown", nil
 	}
 	return string(generatedBy), nil
 }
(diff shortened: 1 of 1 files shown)


client/connector_create.go
```diff
@@ -50,7 +50,7 @@ func secretFileAuthor(ctx context.Context, secretFile string) (author string, er
 	}
 	generatedBy, ok := content["skupper.io/generated-by"]
 	if !ok {
-		return "", fmt.Errorf("Can't find secret origin.")
```
ajssmith, line 53: Suggest leaving as is and caller checks for error strings.

client/connector_create.go
```diff
@@ -90,10 +90,15 @@ func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile st
 	// secrets that we have used to make connections.
 	newConnectionAuthor, err := secretFileAuthor(ctx, secretFile)
 	if err != nil {
-
 		return nil, err
 	}
 
+	alreadyWarned := false
+	if newConnectionAuthor == "unknown" {
```
ajssmith, line 97: Suggest no using a magic string and check for error response instead.

client/connector_create.go
```diff
@@ -102,7 +107,10 @@ func (cli *VanClient) ConnectorCreateFromFile(ctx context.Context, secretFile st
 	for _, oldSecret := range secrets.Items {
 		oldConnectionAuthor, ok := oldSecret.Annotations["skupper.io/generated-by"]
 		if !ok {
-			return nil, fmt.Errorf("A secret has no author.")
+			if !alreadyWarned {
```
ajssmith, line 110: Lean towards not generating a log:

1. The client pattern is to return error indication and leave it to the caller to generate an indication or prompt to the user
2. With the 0.3 release, a user can make multiple connections "silently"

ajssmith (commented):
iiuc, the goal of the change is to make it easier for 0.4 sites to connect to 0.3 sites (e.g. not require them to annotate the connection  tokens generated from a 0.3 site). As it should be a short term transition, it would be ok to just let the connections complete "silently" and turn on enforcement in the 0.5 release. 

Thoughts?

mgoulish:
withdrawn in favor of PR 331

PR #331, opened 2020-12-02
Issue 329: version 4 can't connect to version 3

mgoulish:
* When ecnountering a secret that has no
    'generated-by' annotation, detect the
    condition but do not warn or fail -- yet.
    In version 5 and beyond we will want to
    warn or fail -- TBD.


ajssmith:
With enhancements to version (#380) etc, closing this. 

New issues should be created for any anomalies post 0.5 release

PR #333, opened 2020-12-03
Move edge connectivity test to integration(#305)

mgoulish:
* Introduce edge_connectivity directory and
    two files.

  * Look at how many interior nodes the single
    edge node can see, both directly and
    indirectly. Make sure the count is correct.

  * test 5 topologies


nicob87 (approved):
LGTM, Nice!

fgiorgetti (approved):
Looks great. :shipit:

PR #337, opened 2020-12-04
issue 334 : detect Skupper version when connecting

mgoulish:
* store version info in secret file annotation

* detect in ConnectorCreateFromFile and return


mgoulish:
This is just for evaluation purposes. Let me know if I missed something.
Once we like it we can merge it in sometime after the 4.0 release.

mgoulish:
Please note: the test failed only because I printed out a string that the test was not expecting. ( I want to print that string just so you can see that the code is actually doing something. )
This is just for review anyway. (For now.)

ajssmith:
Per proposal for version changes

PR #349, opened 2020-12-17
ci: trigger release on simple numeric tag

grs:


grs:
This change is on 0.3 and 0.4 branches but I never made it yet on master

nicob87 (approved):
Good!

PR #352, opened 2020-12-18
fix for extraneous system created configmap

grs:


grs:
@ajssmith @nicob87 I think something changed in the version of minikube used, and there is now an extraneous configmap that causes test failure. Are you ok for me to commit this to fix the build on master?

ajssmith (approved):
lgtm

nicob87 (approved):

PR #354, opened 2020-12-21
skupper_cli: removed --version option

### LINKED ISSUE #363: skupper-cli update: remove --version flag
remove the --version flag. We are keeping only "skupper version" command (which is already implemented)

For this, just removing the rootCmd.Version flag assignment is required.

nicob87:
Closes #363 
Capitalize "Skupper" in output.  Remove "version"

A couple of comments:
With this change, "$skupper --version" output changes from:

```

$ ./skupper --version
skupper version 0.4.0-5
```
to

```

$ ./skupper --version
Skupper 0.4.0-5
```

The requested change was to "also" renamte "skupper --version" to "skupper version" BUT, we also have an implementation for skupper verion:

```

$ ./skupper version
client version                 0.4.0-5
transport version              quay.io/skupper/qdrouterd:0.4 (sha256:037ec89c755a)
controller version             quay.io/skupper/service-controller:0.4 (sha256:13b10aa1250d)
```

So... not sure if the idea is to keep both, remove one or rename both

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -863,7 +863,6 @@ func init() {
 	cmdCompletion := NewCmdCompletion()
 
 	rootCmd = &cobra.Command{Use: "skupper"}
-	rootCmd.Version = version
 	rootCmd.AddCommand(cmdInit, cmdDelete, cmdConnectionToken, cmdConnect, cmdDisconnect, cmdCheckConnection, cmdStatus, cmdListConnectors, cmdExpose, cmdUnexpose, cmdListExposed,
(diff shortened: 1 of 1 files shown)


ajssmith:
My understanding of the the proposed change is to remove the --version flag and only have 'skupper version' command similar to kubectl, oc, etc.

Removing the rootCmd.Version assignment removes the --version flag but not sure there is a way to indicate it is deprecated.

nicob87:
Closed in favor of #387

nicob87:
Closed in favor of #387

PR #355, opened 2020-12-21
skupper_cli: bind and unbind, are now subcommands

nicob87:
Old top level commands are keeped for backward compatibility.

```
$ ./skupper bind --help
Command "bind" is deprecated, please use 'skupper service [bind|unbind]' instead
Bind a target to a service

Usage:
  skupper bind <service-name> <target-type> <target-name> [flags]
...
```

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -853,7 +853,10 @@ func init() {
 	cmdDebugDump := NewCmdDebugDump(newClient)
 
 	//backwards compatibility commands hidden
+	deprecatedMessage := "please use 'skupper service [bind|unbind]' instead"
 	cmdBind.Hidden = true
+	cmdBind.Deprecated = deprecatedMessage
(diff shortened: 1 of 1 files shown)


cmd/skupper/skupper.go
```diff
@@ -852,10 +852,16 @@ func init() {
 	cmdVersion := NewCmdVersion(newClient)
 	cmdDebugDump := NewCmdDebugDump(newClient)
 
+	//backwards compatibility commands hidden
+	cmdBind.Hidden = true
+	cmdUnbind.Hidden = true
+
```
ajssmith, line 858: Suggest setting cmdBind.Deprecated and cmdUnbind.Deprecated strings to indicate usage for Service sub-commands. Such as:

Command "bind" is deprecated, Use 'service bind' command instead.

nicob87:
Included in #360

PR #356, opened 2020-12-22
skupper_cli: new command "link [create|delete|status]"

nicob87:
Moreover the new functions have been moved to a dedicated file,
it is important to not that there is NO CHANGE at all in the functions,
code is  just the name.
Only one exception: the "flag" hack in the "LinkCreate" due to the deprecated
version has a different flag name (connection-name vs name).

A minor change in makefile in order to compile with this additional file.

new commands are:
skupper link create
skupper link delete
skupper link status

Old commands are just deprecated (hidden):
skupper connect
skupper disconnect
skupper list-connectors
skupper check-connection

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"time"
 
 	routev1 "github.com/openshift/api/route/v1"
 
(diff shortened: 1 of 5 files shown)


cmd/skupper/skupper_link.go
```diff
@@ -0,0 +1,181 @@
...
+	cmd := &cobra.Command{
+		Use:   "link create <input-token-file> [--name <name>] or link delete ...",
+		Short: "Manage skupper links definitions",
+	}
+	return cmd
+}
+
+var connectorCreateOpts types.ConnectorCreateOptions
+
+//deprecated
```
nicob87, line 26: wrong comment

cmd/skupper/skupper.go
```diff
@@ -852,6 +715,17 @@ func init() {
 	cmdVersion := NewCmdVersion(newClient)
 	cmdDebugDump := NewCmdDebugDump(newClient)
 
+	//backwards compatible commands tunning
+	cmdListConnectors := NewCmdListConnectors(newClient) //listconnectors just keeped
+	cmdListConnectors.Hidden = true
+
+	cmdConnect := NewCmdConnect(newClient)
+	cmdConnect.Hidden = true
```
ajssmith, line 723: Suggest adding cmd.Deprecated string for these command suggesting use of new 'skupper link [command]'

nicob87:
Included in #360

PR #357, opened 2020-12-23
skupper_cli: new command "token create"

nicob87:
This command deprecates the old "connection-token" which is still supported but hidden.

This pr is nested with the previous:
https://github.com/skupperproject/skupper/pull/356


nicob87:
Included in #360

PR #371, opened 2021-01-11
Adding bash completion, for now only "-n" completes with available namespaces

### LINKED ISSUE #367: skupper autocomplete namespaces
Similar to how kubectl autocompletes namespaces, the same may be done for skupper.
Ideally we will enable this feature for autocompleting available namespaces, and it could be possible to enable it for resouces like "--target <tab> <tab> (and autocomplete with available resources of the specified type.

nicob87:
Closes #367 

based/stolen from:
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/cmd.go#L293

It requires "kubectl" to be installed.

To give it a try just:

```
$ make build-cmd #build skupper cli
$ source <(./skupper completion) #source the autocompletion script
$ complete -F __start_skupper ./skupper #hack to make autocompletion work for "./skupper" also (normally it works for "skupper")
$ #try it!
$ ./skupper -n <tab> <tab> #autocompleting flag as first argument
default          kube-node-lease  kube-public      kube-system      private-http-1   private-mongo-1  public-http-1    public-mongo-1
$ ./skupper service -n <tab> <tab>  #autocompleting flag after calling a subcommand
default          kube-node-lease  kube-public      kube-system      private-http-1   private-mongo-1  public-http-1    public-mongo-1
$ ./skupper  <tab> <tab> #verify it is still working for normal subcommands
bind              connect           delete            init              service           unexpose          
check-connection  connection-token  disconnect        list-connectors   status            version           
completion        debug             expose            list-exposed      unbind 


```

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -30,6 +30,8 @@ __kubectl_parse_get()
     local kubectl_out
     if kubectl_out=$(__kubectl_debug_out "kubectl get $(__kubectl_override_flags) -o template --template=\"${template}\" \"$1\""); then
         COMPREPLY+=( $( compgen -W "${kubectl_out[*]}" -- "$cur" ) )
+    else kubectl_out=$(__kubectl_debug_out "oc get $(__kubectl_override_flags) -o template --template=\"${template}\" \"$1\"")
+        COMPREPLY+=( $( compgen -W "${kubectl_out[*]}" -- "$cur" ) )
     fi
 }
 
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -20,6 +20,28 @@ import (
 	"github.com/skupperproject/skupper/client"
 )
 
+const (
+	bash_completion_func = `# $1 is the name of resource (required)
+# $2 is template string for kubectl get (optional)
+__kubectl_parse_get()
+{
+    local template
+    template="${2:-"{{ range .items  }}{{ .metadata.name }} {{ end }}"}"
+    local kubectl_out
+    if kubectl_out=$(__kubectl_debug_out "kubectl get $(__kubectl_override_flags) -o template --template=\"${template}\" \"$1\""); then
+        COMPREPLY+=( $( compgen -W "${kubectl_out[*]}" -- "$cur" ) )
+    fi
+}
+
+
+__kubectl_get_resource_namespace()
+{
+    __kubectl_parse_get "namespace"
+}
+
+`
+)
+
 type ExposeOptions struct {
 	Protocol   string
 	Address    string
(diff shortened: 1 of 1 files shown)


cmd/skupper/skupper.go
```diff
@@ -22,6 +22,28 @@ import (
 
 var version = "undefined"
 
+const (
+	bash_completion_func = `# $1 is the name of resource (required)
+# $2 is template string for kubectl get (optional)
+__kubectl_parse_get()
```
ajssmith, line 28: What should the assumption be in an openshift environment (e.g. oc might be present but not kubectl)?

Would adding list to pkg/kube/namespace be an alternative?

cmd/skupper/skupper.go
```diff
@@ -862,14 +884,28 @@ func init() {
 
 	cmdCompletion := NewCmdCompletion()
 
-	rootCmd = &cobra.Command{Use: "skupper"}
+	rootCmd = &cobra.Command{
+		Use:                    "skupper",
+		BashCompletionFunction: bash_completion_func,
```
fgiorgetti, line 889: @nicob87 I would suggest writing the bash_completion in a separate .sh file,
which would make it way easier to maintain as we go.

Then we could tweak the build command for the `skupper` binary to include a
build-time variable that points to the `bash_completion.sh` file, like:

```
go build -ldflags "-X main.BashCompletionScript=`readlink -f bash_completion.sh`" ...
```

fgiorgetti (changes requested):

cmd/skupper/skupper.go
fgiorgetti: I tested it locally using a 65k shell script and it worked just fine.

nicob87: mmmm and that works for windows and macos? (windows does not have  a cat, to begin) will investigate options anyway...
awesome review!

nicob87: maybe just the "bash_completion_func" variable definition to a dedicated *.go file is a simple solution, if you want that file to be a ".sh" I think we should "read" that file from our go code, but, this makes a dependence and a possible runtime error, on a .sh file, which I do not like so much ...

cmd/skupper/skupper.go
nicob87: basically to auto-complete, bash executes "some executable", it may be anything, in this case we are executing "kubectl -n ..." we could execute "skupper -something" or build a particular new skupper executable like "skupper-completer" orrr whatever option we want.
Will take a look now if from bash I can "try" with "oc" and if fails, then try with "kubectl", which I think it is totally possible.

cmd/skupper/skupper.go
fgiorgetti: The dependency exists only at build-time, after that it no longer exists, but a separate go file just with the constant
might also work.

fgiorgetti:
@ajssmith @grs one comment here (just for tracking). Once we move to go 1.16 we
could use the new go:embed feature to load shell content directly into the go variable
at build time, so this way we can actually keep the bash completion script separately.

fgiorgetti:
Closing as issue #574 has been raised to track this proposal.

PR #374, opened 2021-01-18
Versioning

grs:


cmd/service-controller/console_server.go
```diff
@@ -537,6 +537,7 @@ func (services HttpServiceStatsMap) updateHttpRequestStats(siteId string, reques
 type Site struct {
```
ajssmith, line 537: low-lying but might Site and SiteInfo from site_query be coalesced...

ajssmith (approved):
lgtm

Built cli and controllers and overrode images sans change to pull policy
Checked version via cli and metadata in qdrouterd logs, etc.

cmd/service-controller/console_server.go
grs: Yes, they could actually. I am planning an upcoming fix to the console server to get the ip translation in the right phase, and will merge them at that time.

PR #375, opened 2021-01-19
Fix wait at start of service controller

grs:
This prevents the service controller waiting for a pod that was terminating at the
time the service controller was started (which can happen if skupper is deleted and
re-inited in the same namespace). Previously it looked for all pods matching a label
selector then waited for each of those to be ready. Now it just waits for there to
be at least one ready replica for the deployment.

### DIFF
--- a/cmd/service-controller/main.go
+++ b/cmd/service-controller/main.go
@@ -11,8 +11,6 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


fgiorgetti (approved):
Looks good to me.

ajssmith (approved):
lgtm

PR #380, opened 2021-01-21
update mechanism

grs:
* new cli subcommand, skupper update
* new check on site-contoller startup to update sites as needed
* new client methods to update the version
  - compares internal library version with current site version
  - compares images in use
* utility code to handle various version strings

### DIFF
--- a/pkg/utils/version.go
+++ b/pkg/utils/version.go
@@ -1,0 +1,90 @@
+package utils
+
+import (
+	"regexp"
+	"strconv"
+)
+
+type Version struct {
+	Major     int
+	Minor     int
+	Patch     int
+	Qualifier string
+}
+
+func ParseVersion(version string) Version {
+	var result Version
+	re := regexp.MustCompile(`^v?(\d+)(\.(\d+))?(\.(\d+))?\W?(.+)?`)
+	parts := re.FindStringSubmatch(version)
+	if len(parts) > 0 {
+		result.Major, _ = strconv.Atoi(parts[1])
+	}
+	if len(parts) > 6 {
+		result.Minor, _ = strconv.Atoi(parts[3])
+	}
+	if len(parts) > 5 {
+		result.Patch, _ = strconv.Atoi(parts[5])
+	}
+	if len(parts) > 6 {
+		result.Qualifier = parts[6]
+	}
+	return result
+}
+
+func (a *Version) MoreRecentThan(b Version) bool {
+	if a.Major > b.Major {
+		return true
+	} else if a.Major < b.Major {
+		return false
+	}
+	// a.Major == b.Major, so look at Minor
+	if a.Minor > b.Minor {
+		return true
+	} else if a.Minor < b.Minor {
+		return false
+	}
+	//a.Minor == b.Minor, so look at Patch
+	return a.Patch > b.Patch
+}
+
+func (a *Version) LessRecentThan(b Version) bool {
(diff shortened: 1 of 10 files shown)


pkg/utils/version.go
```diff
@@ -0,0 +1,90 @@
...
+	//a.Minor == b.Minor, so look at Patch
+	return a.Patch < b.Patch
+}
+
+func (a *Version) Equivalent(b Version) bool {
+	return a.Major == b.Major && a.Minor == b.Minor && a.Patch == b.Patch
+}
+
+func (v *Version) IsUndefined() bool {
+	return v.Major == 0 && v.Minor == 0 && v.Patch == 0 && v.Qualifier == ""
```
ajssmith, line 71: Maybe not strict semver but should qualifier be excluded from defined determination?

pkg/utils/version.go
```diff
@@ -0,0 +1,90 @@
...
+		return true
+	} else if a.Minor > b.Minor {
+		return false
+	}
+	//a.Minor == b.Minor, so look at Patch
+	return a.Patch < b.Patch
+}
+
+func (a *Version) Equivalent(b Version) bool {
+	return a.Major == b.Major && a.Minor == b.Minor && a.Patch == b.Patch
```
ajssmith, line 67: Should equivalent only be for "IsDefined" versions?

pkg/utils/version_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{"1.2.3", "1.3.2", false},
+		{"0.22.9993@bar-xyz", "0.22.9993", true},
+		{"0.22.9993@bar-xyz", "0.22.9993+xyz", true},
+		{"0.22.9993@bar-xyz", "0.22.999@bar-xyz", false},
+		{"10.22+whatever", "10.22", true},
+		{"10.22+whatever", "10.22_ignoreme", true},
+		{"10.22+whatever", "10.32_ignoreme", false},
+		{"10+whatever", "10", true},
+		{"10+whatever", "10%rrr", true},
+		{"cat", "dog", true},
```
ajssmith, line 72: Adding from comments above, should this be false either because version is not defined and/or qualifiers should be compare in numeric/ascii lexical order?

pkg/utils/version.go
grs: At present a version is considered 'defined' if it starts with a numeric value (or a 'v' prefixed numeric value). If it does not then the version won't parse. The IsUndefined test was really a way to avoid an error from the ParseVersion function.

pkg/utils/version.go
grs: I took the view that two undefined versions can be considered equivalent but no ordering between them can be determined.

pkg/utils/version_test.go
grs: I don't think it make sense to compare qualifiers in lexical order. As above I took the view that if you cannot determine ordering between two versions, they are essentially equivalent. Released version should never be undefined. So two released versions should always be comparable. A released version is always considered ahead of an undefined version.

ajssmith (approved):
Worked great for me. Checked tests and update of images. Did not check site-controller.

Only minor comment, --force-hup flag might be more easily inferred as as --force-restart

grs:
@ajssmith thanks for the review! Agree on the option name, I have pushed that change now.

PR #381, opened 2021-01-21
Expose more version information

grs:
Mainly used for manual checking at present, but I think will be useful for troubleshooting.

### DIFF
--- a/cmd/service-controller/console_server.go
+++ b/cmd/service-controller/console_server.go
@@ -67,6 +67,40 @@ func authenticated(h http.Handler) http.Handler {
 	}
 }
 
+type VersionInfo struct {
+	ServiceControllerVersion string `json:"service_controller_version"`
+	RouterVersion            string `json:"router_version"`
+	SiteVersion              string `json:"site_version"`
+}
+
+func (server *ConsoleServer) version() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		v := VersionInfo{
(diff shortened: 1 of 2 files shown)


cmd/service-controller/console_server.go
```diff
@@ -67,6 +67,40 @@ func authenticated(h http.Handler) http.Handler {
...
+type VersionInfo struct {
+	ServiceControllerVersion string `json:"service_controller_version"`
+	RouterVersion            string `json:"router_version"`
+	SiteVersion              string `json:"site_version"`
+}
+
+func (server *ConsoleServer) version() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		v := VersionInfo{
+			ServiceControllerVersion: client.Version,
```
ajssmith, line 79: Should  this be the controller image version instead?

iiuc ServiceControllerVersion and SiteVersion will both be 'client.Version' 

e.g. should there be a difference between 'skupper version' and consoler server '/Version'

grs: client.Version is set at compile time. The version of the service controller is the version of the library at the point the image was built (well, the executable within the image to be precise).

The site version is what is recorded in the router config. That will be the library version of the CLI or site controller that initialised or last updated the site. 

At present skupper version will just give the library version of the CLI and the images. You can't tell the version of the library that the service controller image contains from that directly (though the tag name may give a clue).

ajssmith (approved):
lgtm

built locally an accessed console server /Version path

PR #382, opened 2021-01-21
simplify get version

ajssmith:
A full RouterInspect is not needed for getting version and errors could be confusing to users expecting a simple version check.

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/kubernetes"
 )
 
 type ConnectorCreateOptions struct {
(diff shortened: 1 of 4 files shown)


api/types/client.go
```diff
@@ -93,4 +94,5 @@ type VanClientInterface interface {
 	SiteConfigRemove(ctx context.Context) error
 	SkupperDump(ctx context.Context, tarName string, version string, kubeConfigPath string, kubeConfigContext string) error
 	GetNamespace() string
+	GetKubeClient() kubernetes.Interface
```
grs, line 97: My initial instinct (not strong feeling) is that this is creating a leak in a useful abstraction. I wonder if it would be better, should we wish a simpler interface, to just add a GetVersion() method with some struct returned that has only the required information. As I say, not a strong objection, just a thought.

ajssmith: Agreed as updating for mock highlights loss of abstraction. In process of updating. thnx.

grs (approved):
Looks good for now. (Longer term, I don't like the component- and container- names being part of the public interface a sit seems brittle, but this works well for now).

Issue #383, opened 2021-01-22
Skupper unable to update config map after multiple concurrent updates to annotated resources due to conflicts


fgiorgetti:
When trying to update multiple annotated resources at once, sometimes we can see the following error:

E0121 21:40:56.439673       1 definition_monitor.go:537] Failed to delete service definition on service public-annotated-1/nginx-1-svc-target which is no longer annotated: Failed to update skupper-services config map: Operation cannot be fulfilled on configmaps "skupper-services": the object has been modified; please apply your changes to the latest version and try again

This issue has been identified after debugging constant failures in Circle CI for the annotated resource integration test.

fgiorgetti:
Verified that #384 fixes this issue.

PR #384, opened 2021-01-22
retry controller events on error

grs:
Should fix issue #383

### DIFF
--- a/cmd/service-controller/controller.go
+++ b/cmd/service-controller/controller.go
@@ -765,6 +765,12 @@ func (c *Controller) processNextEvent() bool {
 	}(obj)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


grs:
@fgiorgetti I ran the CI three times and it passed each time (the first failure was due to an aborted run as I force pushed an update, you can see it failed on checkout)

fgiorgetti:
> @fgiorgetti I ran the CI three times and it passed each time (the first failure was due to an aborted run as I force pushed an update, you can see it failed on checkout)

@grs I ran it 20 times in a row, using my local cluster and it seems to be passing consistently now.
Thank you!

fgiorgetti (approved):

PR #395, opened 2021-02-01
--ingress flag

nicob87:


client/router_create.go
```diff
@@ -176,7 +176,7 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 			}
 			annotations = map[string]string{"service.alpha.openshift.io/serving-cert-secret-name": "skupper-controller-certs"}
 		}
-	} else if !options.ClusterLocal {
+	} else if options.Ingress == types.IngressLoadBalancerString {
```
fgiorgetti, line 179: Isn't it supposed to be `options.Ingress != types.IngressNoneString`?

client/router_create.go
```diff
@@ -736,6 +736,10 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 // RouterCreate instantiates a VAN (router and controller) deployment
 func (cli *VanClient) RouterCreate(ctx context.Context, options types.SiteConfig) error {
 	// todo return error
+	if options.Spec.Ingress == types.IngressRouteString && cli.RouteClient == nil {
```
fgiorgetti, line 739: @nicob87 One good addition would be to attach methods to the `SiteConfig` or `SiteConfigSpec`, like `IsIngressRoute`, `IsIngressLoadBalancer` and `IsIngressNone` and avoid comparing values directly.

client/site_config_create.go
```diff
@@ -58,8 +58,8 @@ func (cli *VanClient) SiteConfigCreate(ctx context.Context, spec types.SiteConfi
 	if spec.Password != "" {
 		siteConfig.Data["console-password"] = spec.Password
 	}
-	if spec.ClusterLocal {
-		siteConfig.Data["cluster-local"] = "true"
+	if spec.Ingress != "" { // XXX ?! isn't this an error?
```
fgiorgetti, line 61: Same thing here. A good example for having an `IsIngressNone` method.

fgiorgetti (changes requested):
Just a few findings and a small suggestion.

client/router_create.go
nicob87: well "yes and no", see the first "if" is asking if it is a route, then I thought the code is more clear if this "else" asks specifically what it wants to know, since " if no route, and no clusterlocal" -> means it is loadbalancer. 
The resultant behaviour is the same if I ask for your suggestion, or this one implemented currently.

client/router_create.go
nicob87: absolutelly! will do!

nicob87: Done

client/client.go
```diff
@@ -81,3 +82,10 @@ func NewClient(namespace string, context string, kubeConfigPath string) (*VanCli
 
 	return c, nil
 }
+
+func GetIngressRouteIfPossibleLoadBalancerIfNot(cli *VanClient) string {
```
fgiorgetti, line 86: @nicob87 Looks good to me!

fgiorgetti (approved):

client/router_create.go
```diff
@@ -163,11 +163,11 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 	annotations := map[string]string{}
 
 	svcs := []*corev1.Service{}
-	if cli.RouteClient != nil {
```
ajssmith, line 166: Removing this requires setting default appropriately on OpenShift versus k8s. Otherwise, routes will not be created on OpenShift which is current/expected behavior. Comments below.

client/client.go
ajssmith, line 86: This appears to only be used by tests. If ingress is correctly defaulted on openshift versus k8s (see below) is this still needed?

client/site_config_inspect.go
```diff
@@ -73,10 +73,10 @@ func (cli *VanClient) SiteConfigInspect(ctx context.Context, input *corev1.Confi
 	} else {
 		result.Spec.Password = ""
 	}
-	if clusterLocal, ok := siteConfig.Data["cluster-local"]; ok {
-		result.Spec.ClusterLocal, _ = strconv.ParseBool(clusterLocal)
+	if ingress, ok := siteConfig.Data["ingress"]; ok {
+		result.Spec.Ingress = ingress
 	} else {
-		result.Spec.ClusterLocal = false
+		result.Spec.Ingress = types.IngressLoadBalancerString
```
ajssmith, line 79: See ingress defaulting comment below, will need to do similarly here i think...

cmd/skupper/skupper.go
```diff
@@ -243,7 +258,8 @@ installation that can then be connected to other skupper installations`,
 	cmd.Flags().StringVarP(&routerCreateOpts.AuthMode, "console-auth", "", "", "Authentication mode for console(s). One of: 'openshift', 'internal', 'unsecured'")
 	cmd.Flags().StringVarP(&routerCreateOpts.User, "console-user", "", "", "Skupper console user. Valid only when --console-auth=internal")
 	cmd.Flags().StringVarP(&routerCreateOpts.Password, "console-password", "", "", "Skupper console user. Valid only when --console-auth=internal")
-	cmd.Flags().BoolVarP(&routerCreateOpts.ClusterLocal, "cluster-local", "", false, "Set up skupper to only accept connections from within the local cluster.")
+	cmd.Flags().BoolVarP(&ClusterLocal, "cluster-local", "", false, "Set up skupper to only accept connections from within the local cluster.")
+	cmd.Flags().StringVarP(&routerCreateOpts.Ingress, "ingress", "", "loadbalancer", "[loadbalancer|route|none].")
```
ajssmith, line 262: iiuc from the notes, "the default is loadbalancer unless OpenShift in which case it is route"

Need to set default here so that checks against Ingress will do the right thing (such as creating routes on OpenShift)

ajssmith (commented):
Looking really good, comment on ingress default change needed...

client/client.go
grs: One thought would be to make this a method on the client interface, e.g. GetDefaultIngress(). That decision requires access to the client to determine if routes are supported. That method could then be used to set the default in the CLI.

cmd/skupper/skupper.go
nicob87: Aha!
* Well No, the intention (and what I think it is what we decided) is that "LoadBalancer" is the default. There is no more logic evaluating if routes are available or not... it just use what you say in the --ingress flag (and the default is just the default). And for backward compatibility, if you say --cluster-local -> this translates to "--ingress none".

* In line: https://github.com/skupperproject/skupper/pull/395/files#diff-57d99a8b313f5331b29bc0bb6fc79911009716b2aede40f5b67fed9ce7d8516bR739
I am doing a verification, of the "RouteClient" available for the "--ingress route" option.

*Only for tests is that I am using some discovery.

What do you think?

client/client.go
nicob87: (see below answer) :-)

client/router_create.go
nicob87: I think your comment is true, but (see below), IMO it is being set appropriatelly, only if the user says "--ingress route" is when routes are created.

client/client.go
nicob87: The intention was, "not to determine a default anymore", IIUC, we have now three options:
-- ingress none
-- ingress  loadbalancer
-- ingress route

The default for the cli is just one of those three. No magic.

This helper function is used (for now) only for tests. Because that is the case exactly for tests... The test just wants to use "route or loadbalancer", the available one....
(more comments below)

* making this function a method... (I like that idea!) ... It would be something like "GiveMeTheRecommendedIngressInThisCase" -> aka: GetDefaultIngress...

nicob87: Well, now I think, if the intention is to have a double default on cli --ingress(route or loadbalancer). I think I will use this function/method for that...

cmd/skupper/skupper.go
nicob87: iiuc from the notes, "the default is loadbalancer unless OpenShift in which case it is route" -> correct, you are right, I am wrong

client/client.go
```diff
@@ -81,3 +82,10 @@ func NewClient(namespace string, context string, kubeConfigPath string) (*VanCli
 
 	return c, nil
 }
+
+func (cli *VanClient) GetIngressRouteIfPossibleLoadBalancerIfNot() string {
```
ajssmith, line 86: Name suggestion: GetIngressDefault

ajssmith (changes requested):
Having the route and loadbalancer defaults is looking good.  One thing is:

In client/router_create.go there are checks on cli.RouteClient (lines 811,825,889) that should be checks on cli.IsIngressRoute() to prevent exception when '--ingress loadbalancer' is used on OpenShift. Otherwise, there is an exceptiong.

nicob87:
Awesome review @ajssmith  Thanks, working on this

client/router_create.go
```diff
@@ -736,6 +736,10 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 // RouterCreate instantiates a VAN (router and controller) deployment
 func (cli *VanClient) RouterCreate(ctx context.Context, options types.SiteConfig) error {
 	// todo return error
+	if options.Spec.IsIngressRoute() && cli.RouteClient == nil {
+		return fmt.Errorf("Routes client had not been initialized, is this an OCP cluster?")
```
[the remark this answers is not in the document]
nicob87: orr added "Also" to "siteConfigCreate"? or to the cli "Run" function?
there is a logic there in the cli "Run" function that I do not fully understand... 

```siteConfig, err := cli.SiteConfigInspect(context.Background(), nil)
                        if err != nil {
                                return err
                        }

                        if siteConfig == nil {
                                siteConfig, err = cli.SiteConfigCreate(context.Background(), routerCreateOpts)
                                if err != nil {
                                        return err
                                }
                        }
                        // IF siteConfig != nil ? routerCreateOpts is completely ignored?

```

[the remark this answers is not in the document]
ajssmith: Sorry, missed this comment earlier. Unless the skupper-site config map already exists, siteConfig will be nil and the config map will then be created using routerCreateOpts. If it does exist, the settings in the config map should be used. This is for the case that RouterCreate is called by the site-controller instead of the cli.

ajssmith (approved):

PR #396, opened 2021-02-02
Support skupper annotation of daemonsets and statefulsets

grs:


fgiorgetti (approved):
@grs you make our lives easier this way. Thanks for updating the tests as well.
I have submitted a PR to your branch with a minor cosmetic change, just in case
you find it helpful.

nicob87 (approved):
Good! looks like the annotations are here to stay

api/types/types.go
```diff
@@ -86,6 +86,11 @@ var ControllerEditPolicyRule = []rbacv1.PolicyRule{
 		APIGroups: []string{"apps"},
 		Resources: []string{"deployments", "statefulsets"},
 	},
+	{
```
ajssmith, line 89: should the deploy-watch-*.yaml's under the site controller be updated for daemonsets as well?

ajssmith (approved):
lgtm, nice enhancements

api/types/types.go
grs: No, the site controller doesn't watch them, only the service-controller.

PR #397, opened 2021-02-04
Support for configurable (and updatable) router logging configuration

grs:


grs:
This adds a 'router-logging' property to the skupper-site configmap and a --router-logging option to the skupper init command. The value for this is generally one of trace,debug,info,notice,warning or critical. (Specific router modules can also be set at different levels if desired, e.g. info,POLICY:notice,HTTP_ADAPTOR:debug).

For site-controller controlled sites, changes to the router-logging option in the configmap will be applied to the site. For cli controlled sites you can run skupper init again with the --router-logging option to update.

client/router_logging.go
```diff
@@ -0,0 +1,115 @@
+package client
```
ajssmith, line 1: Looking ahead, suggest moving this to pkg/qdr so that it could use by skupper-ce, etc. outside of the client.

grs: These routines are mapping between the type as exposed in the client API and the abstraction as exposed by the qdr module. Moving them into the qdr package would mean that package depended on a type that makes up part of the client API. That does not feel correct to me and I think would prevent you importing qdr without also importing the client API types.

nicob87 (approved):

client/router_update.go
```diff
@@ -95,6 +96,42 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
...
+	}
+	updated := configureRouterLogging(routerConfig, siteConfig.Spec.RouterLogging)
+	if updated {
+		routerConfig.WriteToConfigMap(configmap)
+		_, err = cli.KubeClient.CoreV1().ConfigMaps(settings.ObjectMeta.Namespace).Update(configmap)
+		if err != nil {
+			return false, err
+		}
+		if hup {
+			router, err := cli.KubeClient.AppsV1().Deployments(settings.ObjectMeta.Namespace).Get(types.TransportDeploymentName, metav1.GetOptions{})
```
fgiorgetti, line 120: @grs is this update going to recycle the router pod?
Is it worth doing it live through management API instead, to avoid recycling?

fgiorgetti (commented):
This is a great enhancement by the way.

client/router_update.go
grs: I see that as an optimisation/enhancement. I'd rather wait and see if we find that changing the log levels is common enough that avoidign restarts is worth it.

fgiorgetti (approved):

client/router_logging.go
ajssmith: Agree that the type dependency is problematic, just observing that the implementation is not too coupled to the client and that skupper-ce will need similar.

ajssmith (approved):
lgtm

Built pr locally and tried various log settings and verified running config

client/router_logging.go
grs: I agree with the observation. I see a couple of ways we could deal with it. One is to move the bulk of the logic for ParseRouterLogConfig (and the supporting methods), but leave the file as is. That simply reduces the amount of code to be copied. The other is to figure out how to reuse the api, or parts of it, between skupper-ce and skupper. For example we could move all this file *and* the type exposed to the api to a distinct package (e.g. log_config) and accept that package being exposed as part of the api. Thoughts?

ajssmith: I would lean towards leaving as is for now, it works great. We can come back around to the client/api/pkg sharing topics as there may be other things we want to consider before making changes. Thanks.

PR #401, opened 2021-02-10
update spec for router-mode

ajssmith:


grs (approved):
LGTM!

client/site_config_inspect.go
```diff
@@ -76,7 +85,16 @@ func (cli *VanClient) SiteConfigInspect(ctx context.Context, input *corev1.Confi
...
-		result.Spec.Ingress = types.IngressLoadBalancerString
+		// check for deprecated key
+		if clusterLocal, ok := siteConfig.Data["cluster-local"]; ok {
+			if clusterLocal == "true" {
+				result.Spec.Ingress = types.IngressNoneString
+			} else {
+				result.Spec.Ingress = types.IngressLoadBalancerString
+			}
+		} else {
+			result.Spec.Ingress = types.IngressLoadBalancerString
```
fgiorgetti, line 96: @ajssmith should we assume loadbalancer here or use `GetIngressDefault()`?

cmd/skupper/skupper.go
```diff
@@ -210,11 +209,17 @@ installation that can then be connected to other skupper installations`,
 			}
 
 			if routerModeFlag.Changed {
-				options := []string{routerModeInteriorValue, routerModeEdgeValue}
+				options := []string{string(types.TransportModeInterior), string(types.TransportModeEdge)}
```
fgiorgetti, line 212: TransportModeEdge seems to be missing the TransportMode type at `types.go`

client/router_create.go
```diff
@@ -399,7 +400,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		Name:          "http",
 		ContainerPort: types.TransportLivenessPort,
 	})
-	if !options.IsEdge {
+	if isEdge {
```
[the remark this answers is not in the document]
ajssmith: Great catch!

client/site_config_inspect.go
ajssmith: Very good suggestion. I was keeping the previous default but now that we have GetIngressDefault() we can make a better selection.

cmd/skupper/skupper.go
ajssmith: hmm, I see that it is present

fgiorgetti (approved):
It is working fine now Andy.
I have observed a few issues with http and mongodb tests, but they are
not related with this change (as I am already seeing it happening with
older skupper images).

PR #404, opened 2021-02-16
Service controller logging

grs:


cmd/service-controller/console_server.go
```diff
@@ -72,50 +81,184 @@ type VersionInfo struct {
...
+				for _, s := range d.Services {
+					var service *data.Service
+					if hs, ok := s.(data.HttpService); ok {
+						service = &hs.Service
+					}
+					if ts, ok := s.(data.TcpService); ok {
+						service = &ts.Service
+					}
+					if service != nil {
+						fmt.Fprintln(tw, fmt.Sprintf("%s\t%s\t%s", service.Address, service.Protocol, ""))
```
ajssmith, line 229: Not sure if intended but target is printed on a separate line. 

Also suggest instead of 'target at site name' have a separate column for site to be consistent with other get usage (e.g. would ease parsing)

grs: It is intended, as there may be multiple targets for a single service and I thought a line per target would be clearer.

I did wonder about creating a separate column for site. Will do that shortly.

grs: @ajssmith I have updated the patch to add the explicit column for site in the services view.

ajssmith:
Built a local version for testing. On occasion, exec to skupper-service controller for '-- get' command does not return response output. Have observed this with 'get --help' and 'get events' such as the following:

$ oc get pods
NAME                                          READY   STATUS    RESTARTS   AGE
skupper-router-7789c949f8-pzb7q               1/1     Running   0          11s
skupper-service-controller-58f8597989-rcqzn   1/1     Running   0          6s
$ oc exec -it skupper-service-controller-58f8597989-rcqzn -- get --help
A simple tool to retrieve information via a HTTP GET request

Usage:
  get [events|sites|services|version] [flags]

Flags:
  -h, --help            help for get
  -o, --output string   The output format to use (one of json or text)
$ oc exec -it skupper-service-controller-58f8597989-rcqzn -- get --help
$ oc exec -it skupper-service-controller-58f8597989-rcqzn -- get --help
$ oc exec -it skupper-service-controller-58f8597989-rcqzn -- get --help
A simple tool to retrieve information via a HTTP GET request

Usage:
  get [events|sites|services|version] [flags]

Flags:
  -h, --help            help for get
  -o, --output string   The output format to use (one of json or text)

ajssmith (approved):
lgtm. Really like the new 'get' facility.

Built locally and checked out different link and service setups, validated via get results.

grs:
I see very occasional exits without any output for the exec also, but so far only when run against openshift on aws, not against minikube. I see the same issue replacing an invocation of get with e.g. pwd, so I believe it is the ssh tunnel into the pod that is the issue here rather than anything intrinsic to the command. Less than ideal, but not sure there is anything we can do to avoid that.

Issue #403, opened 2021-02-15
Service controller crashes with --enable-service-sync=false


ted-ross:
Using version 0.4.2...
If --enable-service-sync=false is provided in the skupper init command, the skupper-service-controller container crashes with the following error:

$ kubectl logs skupper-service-controller-677f888cb4-24wvb
2021/02/15 19:14:04 Error getting tls configopen /etc/messaging/ca.crt: no such file or directory

PR #405, opened 2021-02-17
Fix handling of enable-service-sync option

grs:
Fixes issue #403


ajssmith (approved):
Tried locally with/without service sync enabled, verified in logs and no connection for service-sync

grs:
thanks @ajssmith

Issue #58, opened 2020-05-29
allow updating of existing skupper config with revised options


grs:
certainly through changing skupper-site and letting site-controller make changes
ideally also through re-running skupper init on existing install with different options

grs:
Will not address this in v1; v2 has different approach and does allow changes to config.

PR #412, opened 2021-02-24
Fix handling of enable-console option

grs:


grs:
> One comment or suggestion would be to allow modifying `--enable-console` value through `SiteConfigUpdate` so users can enable the console on an already initialized namespace, if needed. If you think it is a good idea, I can raise an issue.

I agree that would be a useful enhancement. (The intention is eventually that all aspects can be updated)

fgiorgetti:
> > One comment or suggestion would be to allow modifying `--enable-console` value through `SiteConfigUpdate` so users can enable the console on an already initialized namespace, if needed. If you think it is a good idea, I can raise an issue.
> 
> I agree that would be a useful enhancement. (The intention is eventually that all aspects can be updated)

And I believe this is going to be addressed by an existing issue #58. Thank you!

ajssmith:
iiuc, intent is the console should be enabled by default, should default be changed from false to true in cmd/skupper/skupper.go?

Otherwise, user will need to '--enable-console on init'

grs:
> iiuc, intent is the console should be enabled by default, should default be changed from false to true in cmd/skupper/skupper.go?

Yes! Had not realised that it was off by default there. Previously the flag was not properly honoured so would always be on regardless. With this fix though we do need to change the default which I have now done.

Well spotted!

ajssmith (approved):
lgtm

built and tried true/false on OpenShift and minikube.

PR #416, opened 2021-02-26
deprecate cluster-local

ajssmith:


grs (approved):

fgiorgetti (approved):
It is working as expected.
Flag is not showing up in the help description but still works with a deprecation message.

PR #421, opened 2021-03-02
service configuration checking

grs:
Adds a service configuration checking mechanism intended to automate intial part of troubleshooting.

See /servicecheck/<address> on controller 8080 port, or use kubectl exec -it <service-controller-pod> -- get servicecheck <address>

### DIFF
--- a/cmd/get/get.go
+++ b/cmd/get/get.go
@@ -9,22 +9,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func path(arg string) (string, error) {
-	if arg == "" {
-		return "version", nil
-	}
-	if arg == "events" || arg == "version" || arg == "sites" || arg == "services" {
-		return arg, nil
-	}
-	return "", fmt.Errorf("Invalid argument: %s", arg)
-}
-
-func get(arg string, output string) error {
-	path, err := path(arg)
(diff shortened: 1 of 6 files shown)


ajssmith:
Observation:

Difference between '-o text' and '-o json' on 'get servicecheck'

for example:
'-o text' returns "No issues found"

'-o json' returns full site_details content

grs:
The text mode just shows the observations, or if there are noe, 'no issues found'. The json gives the full data, the idea being that in that form something else can decide what to show and how to show it.

ajssmith:
Observed for non-existing service e.g. 'get servicecheck foo'

No such service "foo" on bd356eb3-0efe-4f49-aaa1-ce3b968d7169
There are no egress bindings

Suggest filtering additional output for non-existing service

grs:
The 'no such service' messages are scoped to a particular site. Whereas the 'no egress bindings' is network scoped. (I could suppress the latter if there is a 'no such service' observation for all the sites in the network I suppose.)

ajssmith (approved):
lgtm. This is such a valuable capability.

Built and setup two AWS clusters. Tried various sequences of creating/binding/unbinding
services, adding and removing links, etc.

PR #422, opened 2021-03-03
Renaming of certain kubernetes resources created by skupper

grs:


grs:
Oauth now working.

grs:
Update logic added.

pkg/kube/secrets.go
```diff
@@ -92,3 +92,29 @@ func DeleteSecret(name string, namespace string, cli kubernetes.Interface) error
 		return fmt.Errorf("Failed to delete secret: %w", err)
 	}
 }
+
+func CopySecret(src string, dest string, namespace string, kubeclient kubernetes.Interface) error {
```
ajssmith, line 96: Observation that secret type is not preserved in the copy (e.g. kubernetes.io/tls => Opaque). 

Should it be preserved ?

grs: Yes, will fix.

ajssmith:
Suggest that output of 'skupper update' command summarize changes that would affect the user. For example, console url may differ after the update:

https://skupper-controler- ==> https://skupper-

Not sure what else may be meaningful, maybe just doing a 'status' would be sufficient.

grs:
Note: tested with site-controller. However it is necessary to update the role the site controller is running with to allow monitoring of daemonsets. Without this the update will fail as the new service-controller role requires this permission.

grs:
I've made a change to log out the new url if console is changed and to notify if previous site tokens have been invalidated. The latter happens if the router was exposed as a loadbalancer implementation that gives an ip rather than a host name. Though the old ip is still valid (we keep the old service in that case), the certificate does not have the old ip in the subject which prevents correct validation by interconnect.

ajssmith (approved):
lgtm. 

Built with changes. Setup Two clusters, linked them, created and bound services. Performed update on each site. Verified resources copied, site connectivity and service definitions.

PR #423, opened 2021-03-04
Ensure event store is initialised before it is used

grs:


fgiorgetti (approved):
Looks good! Re-tested using a fresh service controller image and
did not face any other problem so far.

PR #425, opened 2021-03-04
Refactored test jobs that run inside the cluster to use a different build tag

### LINKED ISSUE #424: Improve build tags used by test jobs running inside the cluster
The integration tests contain a build tag named `integration`. Some of the integration tests
run a job inside the cluster, and the job functions are currently verifying if they have to be
skipped based on an environment variable.

I would like the tests that run inside the cluster to use a separate build tag, like `job` so that
they no longer need to call `SkipTestJobIfMustBeSkipped`, as they won't be executed
as part of the integration tests anymore.

The benefit is that it becomes a bit simpler to write the test jobs and the produced output
won't report those test functions as being skipped when running integration tests.

fgiorgetti:
The test jobs have been placed into another package and they now use a
`job` tag. This way there is no longer a need to call `SkipTestJobIfMustBeSkipped`
as those tests won't be executed as an integration test.

Resolves #424

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,10 @@ all: build-cmd build-get build-controllers build-tests
 
 build-tests:
 	mkdir -p ${TEST_BINARIES_FOLDER}
-	go test -c -tags=integration -v ./test/integration/tcp_echo -o ${TEST_BINARIES_FOLDER}/tcp_echo_test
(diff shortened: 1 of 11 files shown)


grs (approved):
LGTM!

Issue #420, opened 2021-03-02
Update to use go 1.16


fgiorgetti:
Use go 1.16 and probably need to do some clean up on `go.mod` with `go mod tidy`.
The benefit is that `go build` and `go test` should not affect `go.mod` and `go.sum`.

fgiorgetti:
Another improvement that can benefit from this update, is to use `go:embed` in order to define
the `client.Version` value in a text file that gets linked automatically at build time.

grs:
Is a text file better than taking it from git? One of the benefits of git is that it will be different for every commit.

fgiorgetti:
That's true. Version coming from git tag or commit id is easier and requires less intervention.
A better thing to use this would be the autocomplete shell script, in fact.

PR #435, opened 2021-03-10
Update ci to specify version correctly

grs:


ajssmith (approved):

fgiorgetti (approved):
Works fine.
I will add an extra note into #420 so that we also include the version in a text
file and so it gets linked to "client.Version" at build time automatically (using `go:embed`).

Issue #390, opened 2021-01-28
Bug Report: HTTP services with annotations over time will get into a bad state with the router


Kampe:
Hello again!

Sorry for the strange title, I'm unsure of how else to describe what is happening and I haven't actually caught what's happening in the act as this is something we notice over time while letting skupper run (in this case for almost exactly 7 days), eventually queues will get jammed up, seemingly in just the http proxies. However this is what I have noticed about the services when the odd behavior begins happening:

We'll see the proxies when listed out show the metadata from the router itself:
```
Services exposed through Skupper:
    cloud-api (http port 5443) with targets
      => app=cloud-api,application=skupper-router,skupper.io/component=router name=cloud-api
```

The solution is to delete the service and recreate it, which will allow skupper to take over the pod's ports again.  
  
Once recreated, it'll be in working order again and will look as such:
```
Services exposed through Skupper:
  cloud-api (http port 5443) with targets
    => app=cloud-api name=cloud-api
```

Given some of the behavior I assume this may be related to the service controller, however I don't see anything out of wack in the logs aside from 
```
Failed to authenticate : read /etc/console-users: is a directory
```

Let me know if there's any clues as to where I can poke around at to give you some further details, for context we deploy skupper in a very typical cloud "hub spoke" edge pattern using the site controller.

grs:
The annotation on the service has turned out to be a somewhat problematic mechanism. The original selector defines the target pods, but skupper then overwrites this. It tries to keep the original as an annotation, but there are ways that might be lost. 

A simpler and safer option is to annotate the deployment instead. Would that be an option for you? If so it might be worth deprecating the service annotation (except where the service is itself the target. not the pods it points at)

If not we will try and make things more robust for service annotations. (I see one obvious fix that I think would alleviate your particular case, though the mechanism remains in my view somewhat fragile).

Kampe:
I figured that was potentially what was happening there with how skupper manages the pod ports. As far as annotating the deployments themselves that would work just fine actually, and would make a little more sense the more I think about it.  Is that currently possible with v0.4.2?  I can make that change and give it a good test.

grs:
Yes, annotations on deployments has been possible since 0.2.

Kampe:
Will swap and report back with any findings, is this work tracked on another issue with a better description? Want me to leave the issue open for now?

grs:
Yes, lets leave this open for now.

Kampe:
[...]

PR #441, opened 2021-03-11
ensure skupper selector does not get corrupted when original service …

grs:
…is re-applied

Fixes root cause in issue #390


fgiorgetti (approved):
Fix looks good.
One comment though about the unit test `TestGetServiceDefinitionFromAnnotatedService` in `definition_monitor_test`.
If you run just this single test, like shown below, it fails. See:

```
go test -v -count=1 -run TestGetServiceDefinitionFromAnnotatedService ./cmd/service-controller/...
```

If you call `event.StartDefaultEventStore(nil)` inside it, then it works.

grs:
Fixed TestGetServiceDefinitionFromAnnotatedService

PR #444, opened 2021-03-12
Allow annotations for router and service-controller pods to be specified

grs:
This can be done either as annotations on the skupper-site configmap, for use
with site-controller, or via command line.


ajssmith (approved):
lgtm

Built locally, tried single and multiple annotations via cli. Verified annotations on deployment.

PR #447, opened 2021-03-15
bump service controller to 0.5

ajssmith:


grs (approved):
(Ideally we would use 'master' here and have that automatically updated on every merged commit, but this is good for now).

PR #454, opened 2021-03-17
provide version when no config

ajssmith:


cmd/skupper/skupper.go
```diff
@@ -820,6 +823,11 @@ the .bash_profile. i.e.: $ source <(skupper completion)
 type cobraFunc func(cmd *cobra.Command, args []string)
 
 func newClient(cmd *cobra.Command, args []string) {
+	if cmd.Name() == "version" {
```
grs, line 826: Not an out and out objection, but this feels a little ugly to me.

What about having an alternative to newClient, that does not exit, but records any error (and leaves the cli var unset? This could then be used as the parameter to NewCmdVersion, whose RunE function could then check whether the cli had been set before invoking on it and print the error if not?

ajssmith: I initially started down that path but did not work it all the way thru, will make another pass at it as it would be cleaner for sure...

cmd/skupper/skupper.go
```diff
@@ -750,20 +759,26 @@ func NewCmdUnbind(newClient cobraFunc) *cobra.Command {
...
 		Args:   cobra.NoArgs,
 		PreRun: newClient,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			silenceCobra(cmd)
-
 			fmt.Printf("%-30s %s\n", "client version", client.Version)
-			fmt.Printf("%-30s %s\n", "transport version", cli.GetVersion(types.TransportComponentName, types.TransportContainerName))
-			fmt.Printf("%-30s %s\n", "controller version", cli.GetVersion(types.ControllerComponentName, types.ControllerContainerName))
-
+			if !IsZero(reflect.ValueOf(cli)) {
```
grs, line 775: can you not just test if cli != nil?

grs (approved):

cmd/skupper/skupper.go
ajssmith: I found that I could not as a cli != nil check would crash, iiuc related to the pointer versus underlying value, there may be a more direct way but found this approach as recommended

fgiorgetti (approved):
@ajssmith I see that you have modified the signature of the NewClient, but instead
you could do like this:

```
func NewClient(namespace string, context string, kubeConfigPath string) *client.VanClient {
	return NewClientHandleError(namespace, context, kubeConfigPath, true)
}

func NewClientHandleError(namespace string, context string, kubeConfigPath string, exitOnError bool) *client.VanClient {
...
```

And that would save you from refactoring all the unit tests.

But other than that, it seems to be working fine.

ajssmith:
@fgiorgetti I like your suggestion, will revert NewClient signature...a cleaner approach

cmd/skupper/skupper.go
```diff
@@ -156,10 +157,22 @@ func silenceCobra(cmd *cobra.Command) {
 }
 
 func NewClient(namespace string, context string, kubeConfigPath string) *client.VanClient {
+	return NewClientHandleError(namespace, context, kubeConfigPath, true)
+}
+
+func NewClientHandleError(namespace string, context string, kubeConfigPath string, exitOnError bool) *client.VanClient {
```
grs, line 163: (You could further simplify, by removing the exitOnError flag, returning the error and handling that error by exiting in NewClient(). The newClientSansExit() function could then just ignore the error, or set some variable if there is ever a future need for it).

grs (approved):

PR #458, opened 2021-03-19
Set inter-router session window (configurable via hidden cli options)

grs:


ajssmith (approved):
lgtm, tried local build with and without settings.

Observation, is there any kind of validity/bounds check for the settings to be made?

grs:
> Observation, is there any kind of validity/bounds check for the settings to be made?

Probably, but I'm not sure what at present. I dont *think* the router does any bounds checking itself as yet.

fgiorgetti (approved):
Working just fine.

PR #474, opened 2021-04-16
Use qdrouterd nightly image

fgiorgetti:


fgiorgetti:
@grs @ajssmith What do you guys think about updating qdrouterd image to nightly?

grs (approved):
Just for my information, do we run the nightly against skupper tests before we push it?

fgiorgetti:
Currently not. We only run the qdr-image tests at the moment, but we could
try adding the Skupper tests to run as part of it as well.

Em sex., 16 de abr. de 2021 16:14, grs ***@***.***> escreveu:

grs:
Thats fine for now. It means that any given PR may pull in a new router and might hit issues from the router not the PR itself. That is something we can improve on later though. This way at least we are always using the latest.

fgiorgetti:
@grs this is correct. In case the router image has problems, it might impact the test execution and eventually Skupper itself (if someone builds from sources).

ajssmith (approved):

PR #476, opened 2021-04-26
Ensure connectors have unique names

grs:
This fixes issue where a single deployment is exposed through multiple skupper services.


granzoto (approved):
The fix worked, I could connect to the server using both HTTP and TCP connectors.  Thanks !!

PR #484, opened 2021-05-12
Replace site config key literals with constants

grs:


api/types/types.go
```diff
@@ -27,6 +27,7 @@ const (
 	DefaultVanName      string = "skupper"
 	DefaultSiteName     string = "skupper-site"
 	ClusterLocalPostfix string = ".svc.cluster.local"
+	SiteConfigMapName   string = "skupper-site"
```
fgiorgetti, line 30: I believe the following source files could refer to this new constant:
* client_debug.go:81
* site-controller/controller.go:410
* cli/delete.go:49

client/site_config_create.go
```diff
@@ -12,117 +12,152 @@ import (
...
+	SiteConfigConsoleIngressKey        string = "console-ingress"
+
+	//router options
+	SiteConfigRouterConsoleKey          string = "router-console"
+	SiteConfigRouterLoggingKey          string = "router-logging"
+	SiteConfigRouterDebugModeKey        string = "router-debug-mode"
+	SiteConfigRouterCpuKey              string = "router-cpu"
+	SiteConfigRouterMemoryKey           string = "router-memory"
+	SiteConfigRouterAffinityKey         string = "router-affinity"
+	SiteConfigRouterAntiAffinityKey     string = "router-anti-affinity"
```
fgiorgetti, line 35: One comment on these `affinity` and `anti-affinity` keys. I noticed that most of the
keys match the command line arguments, but the affinity related ones do not, example:
```
Key = router-affinity
Arg = router-pod-affinity
```

and

```
Key = router-anti-affinity
Arg = router-pod-antiaffinity
```

Not sure if they are intended to match each other. What do you think?

fgiorgetti (commented):
@grs A few observations for your evaluation.

api/types/types.go
grs: Agreed, fixed in latest update.

client/site_config_create.go
grs: Agreed, I changed them in the latest update so they match.

fgiorgetti (approved):

PR #494, opened 2021-06-14
Allow access to a site to be revoked

grs:
Provides a CLI command with which to revoke access for any inbound links to a site, by regenerating the CA used to issue certificates for TLS connections.

PR #495, opened 2021-06-14
Allow access to a site to be revoked

grs:
This is a version of https://github.com/skupperproject/skupper/pull/494 that also deals with claims as introduced in https://github.com/skupperproject/skupper/pull/493. Ultimately this is what I believe we should end up with, PR 494 is there in case we want to commit that first as PR 493 is already quite large.

### DIFF
--- a/client/revoke_all.go
+++ b/client/revoke_all.go
@@ -1,0 +1,96 @@
+package client
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+)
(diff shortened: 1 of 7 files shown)


client/revoke_all.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+	siteServerSecret := types.Credential{
+		Name:    types.SiteServerSecret,
+		Subject: types.TransportServiceName,
+	}
+	usingRoutes := false
+	if cli.RouteClient != nil {
+		route, err := kube.GetRoute(types.InterRouterRouteName, cli.Namespace, cli.RouteClient)
```
ajssmith, line 35: Should the types.ClaimRedemptionRouteName route be included in regeneration?

grs: Yes. Should be fixed now.

ajssmith (approved):
lgtm, built service controller, linked two clusters, revoked access, verified services restored and re-linked site with re-generated token.

PR #497, opened 2021-06-17
Initial support for nodeport as ingress

grs:
Also adds a --router-ingress-host by which the user can supply the appropriate name through which the node would be accessed.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -192,6 +192,8 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 			}
 		} else if options.IsConsoleIngressLoadBalancer() {
 			svctype = corev1.ServiceTypeLoadBalancer
+		} else if options.IsConsoleIngressNodePort() {
+			svctype = corev1.ServiceTypeNodePort
 		}
 		if options.IsConsoleIngressRoute() {
 			routes = append(routes, &routev1.Route{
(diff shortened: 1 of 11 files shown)


cmd/skupper/skupper.go
```diff
@@ -351,6 +351,7 @@ installation that can then be connected to other skupper installations`,
 	cmd.Flags().StringVar(&routerCreateOpts.Router.NodeSelector, "router-node-selector", "", "Node selector to control placement of router pods")
 	cmd.Flags().StringVar(&routerCreateOpts.Router.Affinity, "router-pod-affinity", "", "Pod affinity label matches to control placement of router pods")
 	cmd.Flags().StringVar(&routerCreateOpts.Router.AntiAffinity, "router-pod-antiaffinity", "", "Pod antiaffinity label matches to control placement of router pods")
+	cmd.Flags().StringVar(&routerCreateOpts.Router.IngressHost, "router-ingress-host", "", "Host through which node is accesible when using nodeport as ingress.")
```
pwright, line 354: not sure what 'node' means here? site?

grs: Kubernetes node

ajssmith:
Suggest updating --ingress flag string to include nodeport, e.g.

"Setup Skupper ingress to one of: [loadbalancer|route|nodeport|none]." ...

grs:
> Suggest updating --ingress flag string to include nodeport, e.g.
> 
> "Setup Skupper ingress to one of: [loadbalancer|route|nodeport|none]." ...

Good point, done!

ajssmith:
Observed that 'skupper status' does not provide a console url. Should RouterInspect be updated to add case for nodeport?

cmd/skupper/skupper.go
fgiorgetti: Typo here: `accesible`.

fgiorgetti: @grs if you select ingress as `nodeport` but leave the `ingressHost` empty, the generated
token does not seem to work (edge-host and inter-router-host fields are empty). Should it
be a required parameter when using `--ingress nodeport`?

grs: Agreed. Both issues fixed.

grs:
> Observed that 'skupper status' does not provide a console url. Should RouterInspect be updated to add case for nodeport?

I've pushed a change to do this.

client/router_create.go
```diff
@@ -624,22 +626,60 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
...
+		} else if options.IsIngressNodePort() {
+			credentials = append(credentials, types.Credential{
+				CA:          types.SiteCaSecret,
+				Name:        types.SiteServerSecret,
+				Subject:     options.Router.IngressHost,
+				Hosts:       []string{types.TransportServiceName + "." + van.Namespace},
+				ConnectJson: false,
+				Post:        false,
+			})
+			if options.Controller.IngressHost != "" {
```
ajssmith, line 655: Would this work if rather than options.Controller.IngressHost,  options.Router.IngressHost is used for the claims secret? 

I tried `skupper init --ingress nodeport --router-ingress-host 192.168.39.231` and it failed as claims secret did not get created. When i tried  `skupper init --ingress nodeport --router-ingress-host 192.168.39.231 --controller-ingress-host 192.168.39.231` it worked fine.

It seems reasonable that claims service would  correlate to router-ingress rather than console-ingress if that makes sense.

grs: The claims are processed by the controller at present, therefore the host used has to be of the node on which the controller is running. The later PR that includes nginx ingress adds a --ingress-host option which can also be used for nodeport in the event that both controller and router are running on the same node.

ajssmith (approved):
lgtm. Setup two minikube clusters on host. Configured skupper with nodeport ingress, created token and then linked the clusters. Verified inter-router connection, etc. as well as console access via status url.

fgiorgetti (approved):
LGTM.

PR #498, opened 2021-06-23
Initial proxy support

ajssmith:
skupper proxy commands to create proxy definition, instance and download package


client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+	} else {
+		return false
+	}
+}
+
+func generateProxyName(namespace string, cli kubernetes.Interface) (string, error) {
+	proxies, err := cli.CoreV1().ConfigMaps(namespace).List(metav1.ListOptions{LabelSelector: "skupper.io/type=proxy-definition"})
+	max := 1
+	if err == nil {
+		proxy_name_pattern := regexp.MustCompile("proxy([0-9])+")
```
grs, line 206: This has the same issue as Fernando pointed out for the link/connector name generation (fixed in my claims PR). The regex should be proxy([0-9]+)+ in order to match multi digit values (unusual for proxy, but worth fixing). Also the label selector will only look at proxy configmaps, so a configmap that is not that type but clashes with the name would block new proxies.

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+		}
+		proxyConfig.Listeners["amqp"] = qdr.Listener{
+			Name: "amqp",
+			Host: "localhost",
+			Port: int32(amqpPort),
+		}
+	}
+	// store the url for instance queries
+	url := fmt.Sprintf("amqp://127.0.0.1:%s", strconv.Itoa(amqpPort))
+	err = ioutil.WriteFile(proxyDir+"/config/url.txt", []byte(url), 0644)
```
grs, line 382: Check error?

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+			return fmt.Errorf("Failed to marshall router config: %w", err)
+		}
+
+		err = ioutil.WriteFile(proxyDir+"/config/qdrouterd.json", []byte(mc), 0644)
+		if err != nil {
+			return fmt.Errorf("Failed to write config file: %w", err)
+		}
+
+		url, err := ioutil.ReadFile(proxyDir + "/config/url.txt")
+		agent, err := qdr.Connect(string(url), nil)
```
grs, line 717: Minor point, but if for any reason you can't read the url, the error will be overwritten.

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+		if err != nil {
+			return fmt.Errorf("Agent error adding tcp connector: %w", err)
+		}
+		record := map[string]interface{}{}
```
grs, line 721: Maybe have `defer agent.Close()` up here, so that connection is not left open if there is an error processing the endpoint?

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+			return fmt.Errorf("Failed to write config file: %w", err)
+		}
+
+		url, err := ioutil.ReadFile(proxyDir + "/config/url.txt")
+		agent, err := qdr.Connect(string(url), nil)
+		if err != nil {
+			return fmt.Errorf("qdr agent error: %w", err)
+		}
+		if err = agent.Delete("org.apache.qpid.dispatch.tcpConnector", proxyName+"-egress-"+address); err != nil {
+			return fmt.Errorf("Error removing tcp connector : %w", err)
```
grs, line 773: Either close the connection on error, or have a `defer agent.Close()` after successful connect?

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+
+	// todo: only remove service if not used, is this necessary
+	if si != nil && len(si.Targets) == 0 && si.Origin == "" {
+		err := cli.ServiceInterfaceRemove(ctx, address)
+		if err != nil {
+			return fmt.Errorf("Failed to removes service: %w", err)
+		}
+	}
+
+	// Note: unexpose will stop and remove proxy independent of bridge configuration
```
grs, line 841: Another option here would be to only stop the proxy if there are no more service bindings.

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+
+func (cli *VanClient) ProxyInspect(ctx context.Context, proxyName string) (*types.ProxyInspectResponse, error) {
+	proxyDir := getDataHome() + "/skupper/" + proxyName
+
+	_, err := getRootObject(cli)
+	if err != nil {
+		return nil, fmt.Errorf("Skupper not initialized in %s", cli.Namespace)
+	}
+
+	proxyVersion, err := exec.Command("qdrouterd", "-v").Output()
```
grs, line 863: Minor point: Could also get the version via the agent (infact agent.local.Version would gve it to you but not at present accessible outside the package. Could add a `func (a *Agent) getLocalRouterVersion() string` though. Not essential, not necessarily even better, just mentioning in passing.

client/proxy.go
```diff
@@ -0,0 +1,1055 @@
...
+		if err != nil {
+			return fmt.Errorf("Failed to write config file: %w", err)
+		}
+
+		url, err := ioutil.ReadFile(proxyDir + "/config/url.txt")
+		agent, err := qdr.Connect(string(url), nil)
+		if err != nil {
+			return fmt.Errorf("qdr agent error: %w", err)
+		}
+
```
grs, line 984: `defer agent.Close()` again perhaps

client/proxy.go
ajssmith: I had considered that but felt that the spirit of expose/unexpose is to simplify
what commands the user needs to know for the primary use scenario (e.g. expose
process to the skupper network). If a user makes additional bind/forward additions
to the proxy they created via expose then documenting what they will need to
do when they unexpose will get confusing. With the current approach, it will
be clear that if they unexpose, everything goes away. And if they perform
additional bind/forwards then we need to document that they should not
execute the unexpose if they want the proxy config to persist.

Thoughts?

grs: I'm fine with that also.

client/proxy.go
```diff
@@ -0,0 +1,1066 @@
...
+	}
+
+	err = cli.KubeClient.CoreV1().Secrets(cli.GetNamespace()).Delete(proxyPrefix+proxyName, &metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("Unbable to remove proxy secret: %w", err)
+	}
+
+	err = cli.KubeClient.CoreV1().ConfigMaps(cli.GetNamespace()).Delete(proxyPrefix+proxyName, &metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("Unbable to remove proxy config map: %w", err)
```
fgiorgetti, line 652: Typo on `Unbable`.

client/proxy.go
```diff
@@ -0,0 +1,1066 @@
...
+	}
+
+	err := cli.proxyStop(ctx, proxyName)
+	if err != nil {
+		return fmt.Errorf("Not able to stop proxy %w", err)
+	}
+
+	err = cli.KubeClient.CoreV1().Secrets(cli.GetNamespace()).Delete(proxyPrefix+proxyName, &metav1.DeleteOptions{})
+	if err != nil {
+		return fmt.Errorf("Unbable to remove proxy secret: %w", err)
```
fgiorgetti, line 647: Typo on `Unbable`.

client/proxy.go
```diff
@@ -0,0 +1,1066 @@
...
+
+	expand := expandVars()
+	err = writeTar("expandvars.py", []byte(expand), time.Now(), tw)
+	return nil
+}
+
+func (cli *VanClient) proxyStart(ctx context.Context, proxyName string) error {
+	proxyDir := getDataHome() + "/skupper/" + proxyName
+	svcDir := getConfigHome() + "/systemd/user"
+
```
fgiorgetti, line 576: If `$HOME/.local` or `$HOME/.config` do not exist, the call to `startProxyUserService`
fails with:
```
Error: Failed to create start service: Unable to write user unit file: open /home/sample-user/.config/systemd/user/proxy1.service: no such file or directory
```

Seems to resolve by adding:
```
	if err := os.MkdirAll(proxyDir, 0755); err != nil {
		return err
	}
	if err := os.MkdirAll(svcDir, 0755); err != nil {
		return err
	}
```

ajssmith: Done. The local share directories get create in setupLocalDir. Was missing the svcDir creation.

client/proxy.go
ajssmith: fixed

client/proxy.go
ajssmith: fixed

client/proxy_test.go
```diff
@@ -0,0 +1,445 @@
...
+	kubeContext := ""
+	kubeConfigPath := ""
+
+	//isCluster := *clusterRun
+	if *clusterRun {
+		cli, err = NewClient(namespace, kubeContext, kubeConfigPath)
+	} else {
+		cli, err = newMockClient(namespace, kubeContext, kubeConfigPath)
+	}
+	assert.Check(t, err, "test-proxy-download-")
```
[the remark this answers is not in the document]
ajssmith: done

client/proxy.go
```diff
@@ -0,0 +1,1066 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed to write config file: %w", err)
+	}
+
+	return nil
+}
+
+func (cli *VanClient) ProxyInit(ctx context.Context, options types.ProxyInitOptions) (string, error) {
+	var err error
+	proxyName := options.Name
```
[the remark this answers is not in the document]
ajssmith: yes, added validation for letters, numbers and "-"

fgiorgetti (approved):
Changes look good. I have tested it on Fedora, Ubuntu, RHEL 7 and RHEL 8.
Only comment on RHEL 7 is that systemd user instance is not enabled by default,
so it would be nice to add some note explaining it in the documentation (cc @pwright).

grs (approved):
Tried it out and generally looks good. We can always adjust things later.

* I couldn't see any way to configure the port used in skupper proxy init, which means that the download function relies on auto-discovery of free port on a different host (rather than explicit configuration). Did I miss how you would choose the port?

* unexpose deletes the proxy even if the service specified does not exist (e.g. skupper proxy expose foo localhost 9090 then skupper proxy unexpose proxy1 bar) which is not immedeiately obvious (unexpose I think is the most awkward of the subcommands)

* I think 'gateway' might actually be a more helpful name than 'proxy'?? i.e. it provides a gateway into a network

pwright:
very small comment:
```
skupper proxy init
Error: Skupper not initialized in <ns>
```

is good, but slightly inconsistent compared to 

```
skupper init
Skupper is now installed in namespace 'git'. 
```

When the ns is named badly (eg named by me), then the quotes help.

pwright:
question: do proxies appear as blank sites in gilligan?

image

pwright:
typo in error eg when target not specified:

```
$ skupper unexpose
Error: expose target and name must be specified (e.g. 'skupper expose deployment <name>'
```
should be `e.g. 'skupper unexpose'`

PR #507, opened 2021-07-12
Updating token CLI integration test to validate revoke-access

fgiorgetti:


test/utils/skupper/cli/revokeaccess.go
```diff
@@ -0,0 +1,155 @@
...
+
+	//
+	// Creating informers to monitor secrets and pods (before revoke-access is issued):
+	// - Removal of those labeled as 'skupper.io/type=token-claim-record'
+	// - Updates to: skupper-site-ca secret, skupper-site-server, skupper-claims-server
+	// - router and service-controller pods restarted (after deployment updated)
+	//
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	doneCh := d.initializeInformers(cluster, stopCh)
```
grs, line 52: Out of curiousity, why use informers rather than a synchronous check after the command has run?

fgiorgetti: The main reason is to avoid retrieving the secrets and pods before and after issuing the revoke-access
command, as I would need to compare the resourceVersion (or some other attribute) of the secrets and pods.
Using informers I just watch what I am expecting to happen and assert everything is done.

@grs Do you have any other recommendation in this case?

grs: No, that's fine. The only issue with the informers is that they are asynchronous, so you need to wait for some amount of time to be sure you have received necessary events (or end the test only when you get expected events). No issue from me doing it this way though, I was just curious.

grs (approved):

test/utils/skupper/cli/revokeaccess.go
grs: With secrets you don't need to poll. Once the revoke-access returns, the secrets should be updated or deleted. As you say though, both approaches will work.

fgiorgetti: Correct.

ajssmith (approved):
lgtm, ran integration test

PR #510, opened 2021-07-16
Nodeport CA subject fix

fgiorgetti:
The `skupper-site-server` secret's subject was not set to the `router-ingress-host` when using
`nodeport` and it was causing a host verification problem at the client side (bypassed if forcibly
using verifyHostname as false).

Another fix included here is to avoid calling `AppendSecretVolume` when the secret is already
part of the volumes list. Without it, I was facing the following error:
```
error updating deployment: Deployment.apps "skupper-router" is invalid: [spec.template.spec.volumes[4].name: Duplicate value: "conn1", spec.template.spec.containers[0].volumeMounts[4].mountPath: Invalid value: "/etc/qpid-dispatch-certs/conn1-profile/": must be unique]
```

@grs please check this out and let me know if you think it makes sense to you.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -653,10 +653,14 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 				controllerHosts = append(controllerHosts, controllerIngressHost)
 			}
 		}
+		siteServerSubject := types.TransportServiceName
+		if options.IsIngressNodePort() {
+			siteServerSubject = options.GetRouterIngressHost()
+		}
 		credentials = append(credentials, types.Credential{
 			CA:          types.SiteCaSecret,
 			Name:        types.SiteServerSecret,
-			Subject:     types.TransportServiceName,
+			Subject:     siteServerSubject,
 			Hosts:       routerHosts,
 			ConnectJson: false,
 			Post:        post,
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -228,6 +228,7 @@ func (cli *VanClient) ConnectorCreate(ctx context.Context, secret *corev1.Secret
 			return err
 		}
 		updated := false
+		added := false
 		//read annotations to get the host and port to connect to
 		profileName := options.Name + "-profile"
 		if _, ok := current.SslProfiles[profileName]; !ok {
@@ -259,6 +260,7 @@ func (cli *VanClient) ConnectorCreate(ctx context.Context, secret *corev1.Secret
 			}
 		} else {
 			current.AddConnector(connector)
+			added = true
 			updated = true
 		}
 		if updated {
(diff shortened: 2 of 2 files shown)


grs:
@fgiorgetti where were you seeing a verification error? The node host should always be in the SAN list and so the cert should be valid for that host. Was it an IP only maybe?

client/router_create.go
```diff
@@ -653,10 +653,14 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 				controllerHosts = append(controllerHosts, controllerIngressHost)
 			}
 		}
+		siteServerSubject := types.TransportServiceName
+		if options.IsIngressNodePort() {
+			siteServerSubject = options.GetRouterIngressHost()
```
grs, line 658: I think we need to check the length of the host name here, and only use it as the subject if it is less than 64 bytes.

fgiorgetti:
> @fgiorgetti where were you seeing a verification error? The node host should always be in the SAN list and so the cert should be valid for that host. Was it an IP only maybe?

I see this in an attempt to create a scenario with two sites. Site 1 using a nodeport ingress and Site 2 is just a basic site (no custom flags). In my Site2 when I create a link to Site 1, I see the following error at the router logs:

```
2021-07-20 13:03:01.015954 +0000 SERVER (info) [C17] Connection to 192.168.39.129:31383 failed: amqp:connection:framing-error SSL Failure: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed
```

Comparing the certificates generated by the Skupper for my Site 1 when using the default ingress vs nodeport, I can see that in both cases, the SAN list is defined correctly. But host verification fails. Comparing both certificates, the one generated when using just skupper init, has the loadbalancer IP as the Subject CN, while the certificate for the site using node port has the Subject CN as `skupper-router`. See it below.

Regular initialization:
```
skupper -n sk1 init
openssl x509 -text -in <(k -n sk1 get secrets skupper-site-server -o yaml | yq -r '.data."tls.crt"' | base64 -d) | egrep '(Subject:|DNS|IP)'
        Subject: CN = 10.97.178.156
                DNS:skupper-router.sk1, IP Address:10.97.178.156
```

Nodeport initialization:
```
skupper -n sk1 init --console-auth unsecured --console-user admin --console-password admin --ingress nodeport --router-ingress-host `minikube ip` --controller-ingress-host `minikube ip`
openssl x509 -text -in <(k -n sk1 get secrets skupper-site-server -o yaml | yq -r '.data."tls.crt"' | base64 -d) | egrep '(Subject:|DNS|IP)'
        Subject: CN = skupper-router
                DNS:skupper-router.sk1, IP Address:192.168.39.129
```

This change basically defines the Subject CN as the router-ingress-host and then host verification works.

I have confirmed that by manually deleting the skupper-service-controller deployment from my Site 2 namespace (the one that connects to Site 1) and patching the inter-router connector disable hostname verification. After doing that and restarting the router pod, the connection was established.

client/router_create.go
fgiorgetti: Ok, I will include that extra length validation.

grs:
I think actually this may be the right fix long term for the ip issue: https://github.com/skupperproject/skupper/pull/514

I reproduced the issue with an ip host and verified this fixes it.

fgiorgetti:
> I think actually this may be the right fix long term for the ip issue: #514
> 
> I reproduced the issue with an ip host and verified this fixes it.

Perfect. I believe this is the way to go.
@grs Are you ok if I close this PR and open another one just for the connector_create issue?

grs:
>  Are you ok if I close this PR and open another one just for the connector_create issue?

Sure.

PR #511, opened 2021-07-16
Updated basic test to include nodeport validation

fgiorgetti:


client/router_create.go
```diff
@@ -652,10 +652,14 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 				controllerHosts = append(controllerHosts, controllerIngressHost)
 			}
 		}
+		siteServerSubject := types.TransportServiceName
+		if options.IsIngressNodePort() {
+			siteServerSubject = options.GetRouterIngressHost()
```
grs, line 657: Is this needed still (the ingress host should be added as a SAN)? The subject is limited to  less than 64 bytes as I recall.

fgiorgetti: Nope, it was a leftover from another PR I have closed. I have just updated the PR.

grs (approved):

PR #514, opened 2021-07-20
Include IPs as names as well as IPs 

grs:
(router does not recognise the separate IP list)


fgiorgetti (approved):
Worked just fine. Thank you @grs.

PR #526, opened 2021-07-26
Added retries in the Connector Remove test 

nluaces:
In the Connector Remove test, when the connector has been removed,  ConnectorInspect is executed to check that there is no connector and secrets;  but this assertion fails in the master build. 

I could reproduce it only a few times, when the test was executed faster than usual.

For that reason, in this PR the test includes a retry function to call ConnectorInspect (for a specific duration of times and number of attempts) until it returns an error related with a non-existent connector, making pass the test.

### DIFF
--- a/client/connector_remove_test.go
+++ b/client/connector_remove_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/skupperproject/skupper/pkg/utils"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


fgiorgetti:
@nluaces after debugging it locally, I believe that the retry won't help much in this case because the
`ConnectorRemove` function was trying to update the `skupper-router` deployment, which was being
modified concurrently by the `service-controller` after the secret was being removed. It was causing a
conflict that was not being caught by the test because `ConnectorRemove` was using `RetryOnConflict`
and the second attempt was returning nil, instead of the expected error.

With the changes proposed by @grs through #528 the root cause for this issue should be fixed and I
believe this PR can be closed.

PR #528, opened 2021-07-26
revise deletion of links

grs:
Separate the removal of the link from config, from the deletion of the secret. The
latter is the trigger for the former. The former is handled by the service-controller.

This avoids conflicts from both cli and service-controller executing the same code.

### DIFF
--- a/client/connector_remove.go
+++ b/client/connector_remove.go
@@ -4,64 +4,25 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


fgiorgetti:
@grs that is the same conclusion I found after debugging the issue that @nluaces was looking at.
So with your PR, there will be no more conflict when updating the router deployment correct?

grs:
@fgiorgetti right, as the service controller is the only thing altering the configmap, (and it never tries to delete the secret) that should avoid the conflicts. I should have made this change when I moved the token handling into the service-controller.

fgiorgetti (approved):
@grs Working just fine. Ran TestConnectorRemove multiple times against a local cluster and I could not see a single conflict anymore.

nluaces (approved):

Issue #541, opened 2021-08-02
Skupper is using the operator namespace name if no site-name provided


hguerrero:
When installing the operator in a cluster wide mode, if I create a site in my namespace but don't provide a site-name the default used is the namespace where the operator was installed, for example `openshift-operators`

ajssmith:
@fgiorgetti did #542 address this issue?

fgiorgetti:
Fixed by #542

PR #542, opened 2021-08-03
Use correct namespace for default site name

grs:
This fixes an issue using site-controller, where the name defaults to the namespace under which the site controller is running.


grs:
Fixes #541

fgiorgetti (approved):
@grs worked just fine.

PR #549, opened 2021-08-10
Adjusted data endpoint test to work against multiple clusters

fgiorgetti:


fgiorgetti:
@granzoto adjusted this test to use the appropriate ingress, based on the available environment
and deferring the call to TearDown prior to calling Setup.

granzoto (approved):
LGTM

Issue #437, opened 2021-03-10
skupper link status 


pwright:
After creating link for 'hello world' 

```
west $ skupper link status
There are no connectors configured or active

east $ skupper link status
Connection for conn1 is active
```

On west `skupper status` gives me

```
It is connected to 1 other site.
```
which is more info than link status  gives me, should `link status` provide the same info on either side of the connection?

ted-ross:
On east, `skupper link status` shows the status of the connections for the connectors.  On west, it simply says that there are no connectors but doesn't mention incoming connections.
I think an argument can be made that link status should include incoming connections as well as outgoing.

grs:
A counter argument is that `skupper link status` gives information on objects created by `skupper link create` which is outgoing links. Those outgoing links can be deleted with `skupper link delete`.

While I agree that providing more insight into incoming links as well is useful, I'm not convinced `skupper link status` in its current form is the right place for that. Mixing directions muddies the pattern in my view.

nluaces:
This will be addressed  by #878 showing a difference between local links (initiated from the site) and remote links.

PR #554, opened 2021-08-17
Add a breakdown list for incoming and outgoing connections when checking the link status

nluaces:
This PR aims to solve #437 issue by providing a breakdown list of incoming/outgoing connections on each skupper installation.

Using the [Getting Started](https://skupper.io/start/index.html) example, the output of `skupper network status` command will be like:
`west`
```
$ skupper network status
There is 1 incoming link.
There are no outgoing links.
```

`east`
```
$ skupper network status
There are no incoming links.
There is 1 outgoing link.
```
Using the example of [Online Boutique](https://github.com/skupperproject/skupper-example-grpc), the results would be like the following:
`private1`
```
$ skupper network status
There is 1 incoming link.
There are 2 outgoing links.
```
`public1`
```
$ skupper network status
There are 2 incoming links.
There are no outgoing links.
```

 `public2`
```
$ skupper network status
There is 1 incoming link.
There is 1 outgoing link.
```

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -281,4 +281,5 @@ type VanClientInterface interface {
 	GetVersion(component string, name string) string
 	GetIngressDefault() string
 	RevokeAccess(ctx context.Context) error
+	BreakDownConnectionsList() (map[string]int, error)
 }
--- a/client/connector_list.go
+++ b/client/connector_list.go
@@ -66,3 +66,35 @@ func (cli *VanClient) ConnectorList(ctx context.Context) ([]types.LinkStatus, er
 	}
 	return links, nil
 }
+
+func (cli *VanClient) BreakDownConnectionsList() (map[string]int, error) {
+	numIncConnections := 0
+	numOutConnections := 0
+
+	current, err := cli.getRouterConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	connections, err := qdr.GetConnections(cli.Namespace, cli.KubeClient, cli.RestConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	routerType := "inter-router"
+	if current.IsEdge() {
+		routerType = "edge"
+	}
+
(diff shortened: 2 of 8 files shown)


grs:
I think mixing the terms 'connections' and 'links' like this is confusing. The pattern for `skupper xxx status` is that if no name is provided it lists all the relevant objects. If a name is provided it gives information about that object (if it exists).

If we change `skupper link status` to cover incoming as well as outgoing links, then we would need to decide what the identity (i.e. name) of the incoming links is. Incoming links would always be active, since we could not detect them otherwise.

Another approach would be to list the links in both directions as part of a more verbose `skupper status` or perhaps as a distinct `skupper network status`. In this approach `skupper link status` remains focused only on outgoing links, i.e. those configured through `skupper link create`.

nluaces:
I can see now why changing the behavior of `skupper link status` can be confusing, and I agree with the approach of creating a new command `skupper network status` as you mentioned to show incoming/outgoing links. Thanks!

fgiorgetti (approved):
@nluaces It looks good to me.

grs (commented):
This is a good start! I think it would be important in this view - eventually at least - to display more information. E.g. what sites it is linked to (url for outgoing links, some identity for incoming links), possibly also indirectly connected sites (e.g. link foo connects us indirectly to sites x, y, z).

nluaces:
Thanks for your review @grs ! 
@ajssmith mentioned that there is an epic to implement this new command, and maybe it would be useful to have a reference about all the information that could be displayed, the information  that you have mentioned for example. In order to implement this changes, there is the possibility that it's going to be needed a different implementation, so I will keep this PR on hold until we figure it out if it's still useful or should be closed.

nluaces:
This approach is not going to be needed for the new skupper network command, that's why this pull request can be closed.

Issue #551, opened 2021-08-11
Unable to create a token claim when using nodeport ingress on OpenShift


fgiorgetti:
```
$ skupper -n sk1 token create /tmp/sk1.yaml
Error: Failed to create token: routes.route.openshift.io "claims" not found
```

PR #557, opened 2021-08-18
Creating token with ingress host info when using nodeport on OpenShift

fgiorgetti:


client/connector_token_create.go
```diff
@@ -99,67 +99,68 @@ func configureHostPorts(result *RouterHostPorts, cli *VanClient, namespace strin
 	if namespace == "" {
 		namespace = cli.Namespace
 	}
-	ok, err := configureHostPortsFromRoutes(result, cli, namespace)
```
grs, line 102: I *believe* that the logic for cert tokens already checks whether routes are actually in use. The problem is with claim tokens.

client/token_claim_create.go
```diff
@@ -73,26 +73,26 @@ func (cli *VanClient) TokenClaimCreate(ctx context.Context, name string, passwor
 	}
 	host := fmt.Sprintf("%s.%s", types.ControllerServiceName, cli.Namespace)
 	localOnly := true
-	if cli.RouteClient != nil {
-		route, err := cli.RouteClient.Routes(cli.Namespace).Get(types.ClaimRedemptionRouteName, metav1.GetOptions{})
```
grs, line 77: Another approach - the one used for creating cert tokens - is to check the error and to treat IsNotFound as implying routes are not in use.

client/token_claim_create.go
```diff
@@ -73,26 +73,26 @@ func (cli *VanClient) TokenClaimCreate(ctx context.Context, name string, passwor
...
+		localOnly = false
+	} else if cli.RouteClient != nil {
+		route, err := cli.RouteClient.Routes(cli.Namespace).Get(types.ClaimRedemptionRouteName, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
+		host = route.Spec.Host
+		port = 443
+		localOnly = false
+	} else if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
```
grs, line 94: Will this not result in the same error if ingress=loadbalancer on openshift? (As previous else if clause will be true).

client/connector_token_create.go
fgiorgetti: Correct, it is not returning an error when not found occurs. I will revert it.

client/token_claim_create.go
fgiorgetti: I like that approach and will make it similar here as well.

client/token_claim_create.go
fgiorgetti: Yes, the approach you recommended is supposed to fix both cases.

grs (approved):

fgiorgetti:
This resolves #551

PR #558, opened 2021-08-20
Reporting version on service and site controllers

fgiorgetti:


fgiorgetti:
@grs I have used flag.Parse for simplicity. If you think cobra should be used, I will change it.

grs (approved):
Fantastic! Thanks Fernando!

PR #560, opened 2021-08-23
Update ci golang version to 1.15.

nluaces:
Replace mac-i386 platform by mac-arm64, because the former is no longer supported by golang 1.15.


nluaces:
darwin/arm64 does not support Go internal linking: https://github.com/golang/go/issues/39033

PR #563, opened 2021-08-30
Allow network policy to be created to isolate exposed service access …

grs:
…to current namespace

### DIFF
--- a/pkg/kube/networkpolicies.go
+++ b/pkg/kube/networkpolicies.go
@@ -1,0 +1,77 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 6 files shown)


nluaces (approved):

fgiorgetti (approved):

pkg/kube/networkpolicies.go
```diff
@@ -0,0 +1,77 @@
...
+					//all pods from the same network can access any port
+					From: []networkingv1.NetworkPolicyPeer{
+						{
+							PodSelector: &metav1.LabelSelector{},
+						},
+					},
+				},
+				{
+					//everything can access specific TLS protected ports
+					From: []networkingv1.NetworkPolicyPeer{},
```
fgiorgetti, line 53: @grs does it help specifying the router pod's labels here?

grs: Not sure I understand the question. The router pods labels are used in the PodSelector above, which identifies the target. The Ingress rules define the source of connections (the 'From') and the ports.

PR #568, opened 2021-09-03
Support Service TLS

nluaces:
- [x] Generate a CA related to the site
- [x] Generate a certificate when a service is created
- [x] Add ssl profile to HTTP2 connector/listener using service's TLS secret and client secret.
- [x] Remove service's certificate when the service is removed from skupper.
- [x]  Modify `service create` command to add --enableTls  flag
- [x]  Generate secret for the client with site CA


pkg/kube/services.go
```diff
@@ -199,3 +208,20 @@ func WaitServiceExists(name string, namespace string, cli kubernetes.Interface,
...
 }
+
+func createCertificateForService(serviceName string, nameSpace string, hosts string, kubeclient kubernetes.Interface) (*corev1.Secret, error) {
+	caCert, err := kubeclient.CoreV1().Secrets(nameSpace).Get(types.SiteCaServicesSecret, metav1.GetOptions{})
+
+	if err != nil {
+		return nil, err
+	}
+
+	serviceCert := certs.GenerateSecret(serviceName, serviceName, hosts, caCert)
```
grs, line 219: I think we should prefic the secret name with skupper-, just sp when you look at all the secrets you know which ones were created by skupper.

Also, ultimately not every service will use tls, so we need to think about how that option is selected.

api/types/types.go
```diff
@@ -122,6 +122,7 @@ const (
 	SiteCaSecret             string = "skupper-site-ca"
 	ConsoleServerSecret      string = "skupper-console-certs"
 	OauthRouterConsoleSecret string = "skupper-router-console-certs"
+	SiteCaServicesSecret     string = "skupper-site-ca-services"
```
grs, line 125: Minor issue but I think the name could be 'skupper-service-ca' (that aligns well with the other two existing cas, skupper-local-ca and skupper-site-ca.

client/cert_authority_site_create.go
```diff
@@ -0,0 +1,20 @@
+package client
+
+import (
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/kube"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (cli *VanClient) CASiteCreate(ownerRef *metav1.OwnerReference) error {
```
grs, line 10: Again, minor issue on naming, but I think the purpose of the method would be clearer as ServiceCACreate or similar.

pkg/kube/services.go
nluaces: Agree on adding the prefix for the secret.

How about adding a flag to the init command to disable tls on http2 connector?

client/cert_authority_site_create.go
nluaces: Sure! makes more sense

api/types/types.go
nluaces: Done!

client/cert_authority_site_create.go
nluaces: Done!

pkg/kube/services.go
grs: Yes, it needs to be per-service. Thinking aloud, the choice of using tls for ingress and egress are in theory orthogonal. For the ingress we probably do want it to be in ServiceBinding so that it can be communicated to other sites so they can also generate a secret for it. Even for egress, if we move to having the service-controller do the generation at some point, then we would want it in ServiceBinding. What if we have tlsCredentials as a string, holding the name of the secret to be used, in both the main ServiceBinding and also in ServiceTarget? (That way we could in future perhaps allow user specified credentials as well as skupper generated ones).

Would then also need to think if the flags for the cli that would enable these options.

Wdyt?

nluaces: Sounds good!

nluaces: Solved except the part of making the tls config optional.

nluaces: By default, services are not going to have TLS config, only when it is specified in with the flag. If the argument for the --tlsCredentials flag is "default", the generated certificates for the service will be used. What do you thing about this?

client/serviceinterface_remove.go
```diff
@@ -19,11 +21,14 @@ func (cli *VanClient) ServiceInterfaceRemove(ctx context.Context, address string
...
 			delete(current.Data, address)
 			_, err = cli.KubeClient.CoreV1().ConfigMaps(cli.Namespace).Update(current)
+
 			if err != nil {
 				return fmt.Errorf("Failed to update skupper-services config map: %v", err.Error())
-			} else {
-				return nil
 			}
+
+			removeServiceCertificate(cli.Namespace, address, cli.KubeClient)
```
grs, line 29: Probably want to make this conditional on the tlsCredentials being defined for the service?

nluaces: You are correct! done

api/types/client.go
```diff
@@ -252,4 +253,9 @@ type VanClientInterface interface {
 	GetVersion(component string, name string) string
 	GetIngressDefault() string
 	RevokeAccess(ctx context.Context) error
+	CreateServiceCA(ownerRef *metav1.OwnerReference) error
```
ajssmith, line 256: As these are resource operations, suggest not adding to the client interface/model and look at leveraging/enhancing the associated pkg routines.

client/router_create.go
```diff
@@ -1180,6 +1180,14 @@ sasldb_path: /tmp/qdrouterd.sasldb
 			return err
 		}
 	}
+
+	// Create a CA associated with the site to create certificates for each service
+	// and a generic client secret with that CA.
+	err = cli.CreateServiceCA(siteOwnerRef)
```
ajssmith, line 1186: An alternative to consider would be to add the ServiceCaSecret to the van.CertAuthoritys so that the CA gets created in a fashion similar to the others in the van definition.

client/tls_support.go
```diff
@@ -0,0 +1,162 @@
...
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cli *VanClient) AppendSecretToRouter(secretName string, waitForRestart bool) error {
```
ajssmith, line 72: Suggest considering whether this is a general operation on the qdr config resource and/or an general operation on a deployment to append/remove volumes for a deployment. Possibly a combination of both.

ajssmith (commented):
@nluaces this is very well done. Suggestions mostly for considering whether addition to client interface or resource operations (e.g. part of pkg's).

client/router_create.go
nluaces: Thanks, this is very helpful. The CA for the skupper site is now created with the rest of the CAs that Skupper manages.

client/tls_support.go
nluaces: Thanks, this was removed from the client and moved into the qdr package, let me know if that makes sense to you.

client/router_create.go
```diff
@@ -1197,6 +1201,13 @@ sasldb_path: /tmp/qdrouterd.sasldb
 			return err
 		}
 	}
+
+	// Create a generic client secret with the CA associated with the site.
+	err = cli.createGenericClientSecret(siteOwnerRef)
```
ajssmith, line 1206: Is it possible to create the secret by adding the van.TransportCredentials? 

Related suggestion would be to add a 'simple bool' indicator to kube.NewSecret  to consolodate/supplant NewSimpleSecret

ajssmith (commented):
Thanks for the updates.

fgiorgetti:
One comment. If you do a `skupper delete`, when you have existing tls services, the respective secrets are not being deleted.

nluaces:
> One comment. If you do a `skupper delete`, when you have existing tls services, the respective secrets are not being deleted.

It is true, thanks! I will work on that.

client/router_create.go
nluaces: Done! with this refactor, the secret is created before the router; and this avoids restarting the router after it is created.

api/types/types.go
```diff
@@ -234,6 +235,10 @@ const (
 	ServiceSyncAddress = "mc/$skupper-service-sync"
 )
 
+const (
+	SkupperServiceCertPrefix string = "skupper-"
```
fgiorgetti, line 239: One minor question here.
Is it safer/better to be more specific on the prefix, like: `skupper-tls-`?

The code is already validating if there is a conflict, but using a more specific
prefix, it is easier to list "all" secrets related to tls services.

pkg/kube/secrets.go
```diff
@@ -139,3 +144,27 @@ func RegenerateCredentials(credential types.Credential, namespace string, ca *co
 	current.Data = regenerated.Data
 	return cli.CoreV1().Secrets(namespace).Update(current)
 }
+
+func t(name string, caSecret *corev1.Secret, owner *metav1.OwnerReference, namespace string, cli kubernetes.Interface) (*corev1.Secret, error) {
```
fgiorgetti, line 148: Is it being used?

fgiorgetti (commented):
Only a few comments, but overall great job @nluaces ! :)

api/types/types.go
nluaces: Sounds good, could be useful when executing skupper delete command

pkg/kube/secrets.go
nluaces: It is not, thanks for noticing this!

api/types/types.go
nluaces: Done.

ajssmith (commented):
An alternative to removing the flag would be to use cobra's feature to set the flag as hidden e.g.

f := cmd.Flags().BoolVar(&exposeOpts.EnableTls,...
f.Hidden = true

This will allow the patch to be used for testing, etc. as the tls feature rolls out

nluaces:
> An alternative to removing the flag would be to use cobra's feature to set the flag as hidden e.g.
> 
> f := cmd.Flags().BoolVar(&exposeOpts.EnableTls,... f.Hidden = true
> 
> This will allow the patch to be used for testing, etc. as the tls feature rolls out

Indeed! thank you

ajssmith (approved):
lgtm, thanks!

client/serviceinterface_create.go
```diff
@@ -16,6 +18,59 @@ func (cli *VanClient) ServiceInterfaceCreate(ctx context.Context, service *types
...
+			serviceSecret, err := kube.NewSecret(serviceCredential, &ownerReference, cli.Namespace, cli.KubeClient)
+			if err != nil {
+				return err
+			}
+
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+
+				err := qdr.AddSslProfile(serviceSecret.Name, cli.Namespace, cli.KubeClient)
+				if err != nil {
+					return err
```
fgiorgetti, line 54: @nluaces Should this be part of the retry logic? I am wondering if just the deployment
update fails, then the `sslProfile` will be already defined, in case of a possible retry attempt.

client/serviceinterface_remove.go
```diff
@@ -46,3 +51,43 @@ func (cli *VanClient) ServiceInterfaceRemove(ctx context.Context, address string
...
+	secret, err := cli.KubeClient.CoreV1().Secrets(cli.Namespace).Get(certName, metav1.GetOptions{})
+
+	if err == nil && secret != nil {
+
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+
+			err := qdr.RemoveSslProfile(secret.Name, cli.Namespace, cli.KubeClient)
+			if err != nil {
+				return err
+			}
```
fgiorgetti, line 67: Same thing here.

cmd/service-controller/controller.go
```diff
@@ -857,3 +861,106 @@ func (c *Controller) processNextEvent() bool {
...
+			if err != nil {
+				return err
+			}
+
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+
+				err := qdr.AddSslProfile(serviceSecret.Name, c.vanClient.Namespace, c.vanClient.KubeClient)
+				if err != nil {
+					return err
+				}
```
fgiorgetti, line 905: ... and here

cmd/service-controller/controller.go
```diff
@@ -857,3 +861,106 @@ func (c *Controller) processNextEvent() bool {
...
+}
+
+func (c *Controller) handleRemovingTlsSupport(tlsCredentials string) error {
+
+	if len(tlsCredentials) > 0 {
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+
+			err := qdr.RemoveSslProfile(tlsCredentials, c.vanClient.Namespace, c.vanClient.KubeClient)
+			if err != nil {
+				return err
```
fgiorgetti, line 938: ... and here

client/serviceinterface_create.go
nluaces: Agree, it is changed now. Thanks!

client/serviceinterface_remove.go
nluaces: Done.

cmd/service-controller/controller.go
nluaces: Done.

cmd/service-controller/controller.go
nluaces: Done.

fgiorgetti:
@nluaces I believe that this PR has a dependency on a [pending change](https://github.com/interconnectedcloud/qdr-image/pull/50) to qdrouterd image, which builds proton using `-DBUILD_TLS=ON`.

nluaces:
> @nluaces I believe that this PR has a dependency on a [pending change](https://github.com/interconnectedcloud/qdr-image/pull/50) to qdrouterd image, which builds proton using `-DBUILD_TLS=ON`.

Thanks for the update, do you think that the pending change is a blocker to merge the pull request?
The `enable-tls` flag will be hidden while the TLS feature is not available in the version of the router that skupper is using (I was thinking on creating a new issue to make this flag available again when the dependency with the router is solved). What do you think?

nluaces:
#665 issue created.

PR #572, opened 2021-09-10
add http and export config

ajssmith:


nluaces (approved):

fgiorgetti (commented):
@ajssmith I had issues with `gateway service bind` or `forward` on a 
local gateway (active), as template variables in the config file were not
being replaced as expected.

fgiorgetti (approved):
LGTM!

PR #573, opened 2021-09-15
serve console over https

### LINKED ISSUE #566: Console must be served by https before reading and writing to clipboard is supported
The new site-linking UI in the console uses the clipboard to store link tokens. However, with Chrome, the clipboard is not available unless the console is served via https (or from localhost as was the case in my initial testing).

Requested changes:
- The controller should use https to serve the console.
- The console link displayed via a skupper status command should contain the full url
- The links displayed in the kubernetes dashboard should also contain the full url

grs:
fixes #566


fgiorgetti (approved):
Worked for me with K8S and OCP.

PR #584, opened 2021-09-27
Add extra labels to skupper deployments

ajssmith:
Define build time via labels.go or run-time via env vars

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"os"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


nluaces (approved):

grs (approved):
Looks fine to me. Couple of minor questions:

* do we need to be able to override via env var (over compile time value, and instead of modifying the skupper-site configmap)?
* are we sure we don't want to allow different labels on router v controller?
* was there a reason to choose ':' over '='?

None of these are objections to merging though.

client/router_create.go
```diff
@@ -50,6 +52,35 @@ func OauthProxyContainer(serviceAccount string, servicePort string) *corev1.Cont
...
+		if labels == "" {
+			labels = envLabels
+		} else {
+			labels = labels + "," + envLabels
+		}
+	}
+
+	if labels != "" {
+		// "key1:value1,key2:value2" where key and value are lowercase RFC 1123 compliant
+		labelRegex := regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9]):[a-z0-9]([-a-z0-9]*[a-z0-9])+(,[a-z0-9]([-a-z0-9]*[a-z0-9]):[a-z0-9]([-a-z0-9]*[a-z0-9])+)*`)
```
fgiorgetti, line 71: This regexp could be defined as a public variable as I see potential for it being
reused in other places.

fgiorgetti (approved):

ajssmith:
@grs @fgiorgetti updating per suggestions, will keep same label for router, service-controller for now

grs (approved):

PR #590, opened 2021-10-08
Fix console data to handle gateways better

grs:


grs:
@fgiorgetti Should be fixed now

PR #637, opened 2021-11-11
Properly retrying during service creation or removal

fgiorgetti:


nluaces (approved):

fgiorgetti:
Just to give some context. This helps fixing issues with repeated calls to create services using the API, like done by the hipster shop integration test.

I created a reproducer that runs two threads against two linked sites creating and removing services in a loop. With this fix, the reproducer ran for 30 minutes without issues.

ajssmith (approved):

grs (approved):

PR #641, opened 2021-11-18
Skupper Network status command

nluaces:
In order to execute this command:
`$ skupper network status`

Ouput example:
Screenshot from 2021-12-22 20-09-19

### DIFF
--- a/client/connector_list.go
+++ b/client/connector_list.go
@@ -83,7 +83,7 @@ func (cli *VanClient) getLinkStatusByNamespace(namespace string) (map[string]*ty
 	connections, _ := qdr.GetConnections(namespace, cli.KubeClient, cli.RestConfig)
 	for _, s := range secrets.Items {
 		var connectedTo string
-		connectedTo = s.ObjectMeta.Annotations[types.TokenGeneratedBy]
+		connectedTo = s.ObjectMeta.Annotations[types.TokenGeneratedBy][:7]
 		linkStatus := getLinkStatus(&s, edge, connections)
 		mapLinks[connectedTo] = &linkStatus
 	}
--- a/client/connector_list.go
+++ b/client/connector_list.go
@@ -69,7 +69,7 @@ func (cli *VanClient) ConnectorList(ctx context.Context) ([]types.LinkStatus, er
 	return links, nil
 }
 
-func (cli *VanClient) getLinkStatusByNamespace(namespace string) (map[string]*types.LinkStatus, error) {
+func (cli *VanClient) getLinkStatusByNamespace(namespace string, siteNameMap map[string]string) (map[string]*types.LinkStatus, error) {
 	mapLinks := make(map[string]*types.LinkStatus)
 	secrets, err := cli.KubeClient.CoreV1().Secrets(namespace).List(metav1.ListOptions{LabelSelector: "skupper.io/type in (connection-token, token-claim)"})
 	if err != nil {
@@ -89,7 +89,8 @@ func (cli *VanClient) getLinkStatusByNamespace(namespace string) (map[string]*ty
 
 	for _, s := range secrets.Items {
 		var connectedTo string
-		connectedTo = s.ObjectMeta.Annotations[types.TokenGeneratedBy][:7]
+		siteId := s.ObjectMeta.Annotations[types.TokenGeneratedBy]
+		connectedTo = siteId[:7] + "-" + siteNameMap[siteId]
 		linkStatus := getLinkStatus(&s, edge, connections)
 		mapLinks[connectedTo] = &linkStatus
 	}
--- a/client/connector_list.go
+++ b/client/connector_list.go
@@ -68,3 +68,24 @@ func (cli *VanClient) ConnectorList(ctx context.Context) ([]types.LinkStatus, er
 	}
 	return links, nil
 }
+
+func (cli *VanClient) getLinkStatusByNamespace(namespace string) (map[string]*types.LinkStatus, error) {
+	mapLinks := make(map[string]*types.LinkStatus)
+	secrets, err := cli.KubeClient.CoreV1().Secrets(namespace).List(metav1.ListOptions{LabelSelector: "skupper.io/type in (connection-token, token-claim)"})
+	if err != nil {
+		return nil, err
+	}
+	current, err := cli.getRouterConfig()
+	if err != nil {
+		return nil, err
+	}
+	edge := current.IsEdge()
+	connections, _ := qdr.GetConnections(namespace, cli.KubeClient, cli.RestConfig)
+	for _, s := range secrets.Items {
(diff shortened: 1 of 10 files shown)


grs (approved):

client/connector_list.go
```diff
@@ -68,3 +68,24 @@ func (cli *VanClient) ConnectorList(ctx context.Context) ([]types.LinkStatus, er
...
+	secrets, err := cli.KubeClient.CoreV1().Secrets(namespace).List(metav1.ListOptions{LabelSelector: "skupper.io/type in (connection-token, token-claim)"})
+	if err != nil {
+		return nil, err
+	}
+	current, err := cli.getRouterConfig()
+	if err != nil {
+		return nil, err
+	}
+	edge := current.IsEdge()
+	connections, _ := qdr.GetConnections(namespace, cli.KubeClient, cli.RestConfig)
```
ajssmith, line 83: Suggest checking error returned on GetConnections (e.g. could be router exec or unmarshal issue)

client/network_status.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var listSites []*types.SiteInfo
+
+	for _, site := range versionCheckedSites {
+
+		if len(site.Namespace) == 0 {
+			return nil, errors.New("unable to provide site information")
```
ajssmith, line 47: Suggest use of fmt.Errorf in place of errors.New, has generally been the convention used.

cmd/skupper/skupper_network.go
```diff
@@ -0,0 +1,105 @@
...
+		Short:  "Shows information about the current site, and connected sites.",
+		Args:   cobra.MaximumNArgs(1),
+		PreRun: newClient,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			silenceCobra(cmd)
+
+			sites, err := cli.NetworkStatus()
+
+			if err != nil {
+				fmt.Println("Unable to retrieve network information.")
```
ajssmith, line 32: It would be helpful to output the error so the user can understand what the issue might be

client/network_status.go
```diff
@@ -0,0 +1,152 @@
...
+	"errors"
+	"fmt"
+	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/pkg/server"
+	"github.com/skupperproject/skupper/pkg/utils"
+	"time"
+)
+
+func (cli *VanClient) NetworkStatus() ([]*types.SiteInfo, error) {
+
```
ajssmith, line 13: Suggest checking that skupper router/deployment is running as network status relies on such

ajssmith (commented):
Observed that following 'skupper gateway init', network status command does not succeed. May need to check for/exclude gateways rather than returning 'Unable to retrieve network information'

cmd/skupper/skupper_network.go
```diff
@@ -0,0 +1,105 @@
...
+						location = "local"
+					}
+
+					newItem := fmt.Sprintf("[%s] %s - %s ", location, site.SiteId[:7], site.Name)
+
+					newItem = newItem + fmt.Sprintln()
+					detailsMap := map[string]string{"name": site.Name, "namespace": site.Namespace, "URL": site.Url, "version": siteVersion}
+
+					if len(site.Links) > 0 {
+						detailsMap["sites linked to"] = fmt.Sprint(strings.Join(site.Links, ", "))
```
fgiorgetti, line 71: It would be easier to read if you could include the name as well. Example:
```
sites linked to: c3e2204 (site1)
```

client/network_status.go
```diff
@@ -0,0 +1,152 @@
...
+		links = append(links, connection)
+	}
+
+	for _, link := range links {
+		var formattedLink string
+
+		if mapLinkStatus[link].Connected {
+			formattedLink = link
+		} else {
+			formattedLink = fmt.Sprintf("%s%s(link not active)%s", lightRed, link, resetColor)
```
fgiorgetti, line 93: What do you think about adding a blank space between the site id and the parenthesis.

client/connector_list.go
nluaces: Good point, I will check it

client/network_status.go
nluaces: Sure thing, I agree

client/network_status.go
nluaces: Good point, I will work on it

cmd/skupper/skupper_network.go
nluaces: Nice feedback, thanks!

client/network_status.go
nluaces: Agree on that, I will work on it

pkg/utils/retry.go
```diff
@@ -48,18 +48,14 @@ func Retry(interval time.Duration, maxRetries int, f ConditionFunc) error {
 
 	for i := 0; ; i++ {
 		ok, err := f()
-		if err != nil {
+		if err != nil && i == maxRetries {
```
nluaces, line 51: Returns the error when the max retries number has reached.

fgiorgetti: The `RetryError` type is no longer used after this changed.
Should it be kept or removed as well?

nluaces: Good point, I reckon that I could remove it if it is no longer used.

fgiorgetti (approved):

client/network_status.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var listSites []*types.SiteInfo
+
+	for _, site := range versionCheckedSites {
+
+		if len(site.Namespace) == 0 {
+			return nil, fmt.Errorf("temporarily unable to provide site information")
```
ajssmith, line 39: The gateway will show up in the site list without a namespace. Suggest that gateways be excluded from list (for now) and that site loop 'continue' here rather than return error.

If namespace check was also being used to detect incomplete site info, we should figure out a more deterministic way to identify.

client/network_status.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		listLinks, err := cli.getSiteLinksStatus(site.Namespace, siteNameMap)
```
ajssmith, line 42: Suggest that the link status should be for local site only

client/network_status.go
nluaces: Thanks for finding this, gateways are filtered for now.

client/network_status.go
nluaces: Thanks for finding this, the status of the link is showed only for local site.

ajssmith (approved):
lgtm thanks for this

PR #656, opened 2021-12-10
Router scaling

grs:


nluaces (approved):

ajssmith (commented):
An observation is that prior to this change, qdr connections were interpreted as "sites".  

With the change, for example, two sites with --routers 3 will report the following status

"It is connected to 5 other sites (3 indirectly)"

grs:
@ajssmith Thanks! I've fixed that now.

ajssmith (approved):
lgtm, thanks!

fgiorgetti (commented):
LGTM

PR #659, opened 2021-12-13
disable named gateways

ajssmith:
This patchset:
- moves to single gateway per user/host (e.g. no named gateway)
- no longer uses default amqp port
- fixes directory permissions
- other bug fixes

### DIFF
--- a/client/gateway.go
+++ b/client/gateway.go
@@ -5,6 +5,9 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


nluaces (approved):

client/gateway.go
```diff
@@ -770,12 +875,181 @@ func (cli *VanClient) GatewayInit(ctx context.Context, gatewayName string, gatew
...
+					return gatewayName, fmt.Errorf("Error adding http entity: %w", err)
+				}
+			}
+
+			routerConfigCurrent.Bridges = *bridgeConfigFromFile
+			routerConfigCurrent.UpdateConfigMap(configmap)
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				_, err = cli.KubeClient.CoreV1().ConfigMaps(cli.GetNamespace()).Update(configmap)
+				return err
+			})
```
fgiorgetti, line 1028: @ajssmith don't you need to include these two lines 1023:1024 inside the `RetryOnConflict` block,
along with a "reload" of the original `configmap` variable (to make sure you're on the latest version)?

fgiorgetti (commented):
Looks solid to me. One small observation though.

client/gateway.go
ajssmith: @fgiorgetti good observation, all retry logic needs re-factoring and will update

nluaces (approved):

fgiorgetti (approved):
LGTM! Thanks Andy.

PR #666, opened 2022-01-17
Allow a period in gateway names

markmc:
The gateway name is derived from a hostname, so periods should be allowed.
    
Fixes:

```    
$> skupper gateway init
Skupper gateway: 'markmc-machine.lab.redhat.com-markmc'. Use 'skupper gateway status' to get more informaiton.
```


ajssmith (approved):
lgtm

thanks for the fix!

PR #668, opened 2022-01-21
Policies for network, services and gateway

fgiorgetti:


client/policy_validator.go
```diff
@@ -0,0 +1,348 @@
...
+		return false
+	}
+	return true
+}
+
+func (p *ClusterPolicyValidator) NoPermission(err error) bool {
+	if err == nil {
+		return true
+	}
+	return !strings.Contains(err.Error(), "is forbidden")
```
nluaces, line 140: I was thinking that if the error text is changed (for instance by replacing "is forbidden" with "is not allowed"), would imply also a change in the behaviour if it is not detected by the tests. What do you think about checking if the error contains some sort of alphanumeric code? I am aware that the problem would persist in the same way, but I think that error messages directed to users have more chance of being modified.

fgiorgetti: Agreed. The code can be retrieved by the concrete error implementation.
I will update it and see if I can use the appropriate error code.

pkg/qdr/amqp_mgmt.go
```diff
@@ -5,11 +5,10 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	amqp "github.com/interconnectedcloud/go-amqp"
```
grs, line 8: Just out of curiousity, what is the thinking on changing the import order here?

Makefile
```diff
@@ -53,6 +53,9 @@ docker-push: docker-push-test-image
 format:
 	go fmt ./...
 
+generate-client:
```
grs, line 56: Is this target just triggered manually for now?

cmd/service-controller/service_sync.go
```diff
@@ -128,7 +128,14 @@ func (c *Controller) ensureServiceInterfaceDefinitions(origin string, serviceInt
 
 	c.heardFrom[origin] = time.Now()
 
+	policy := client.NewClusterPolicyValidator(c.vanClient)
+
 	for _, def := range serviceInterfaceDefs {
+		res := policy.ValidateImportService(def.Address)
```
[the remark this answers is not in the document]
fgiorgetti: If we move the logic to the service controller itself, lets say under `updateServiceBindings`, then there is
no need to handle it here, as part of the service sync.

The reason why I added it here, is to ignore a not authorized service from existing in a remote site at all.
I know that this will generate lots of events every time services are synced and ignored, but the intention is
actually to see that the service creation was rejected.

What do you see as a benefit of moving this somewhere else, like the controller itself?

pkg/qdr/amqp_mgmt.go
fgiorgetti: I can revert that, this is my IDE "organizing" imports when I opened it for some reason.

Makefile
fgiorgetti: That's correct. If we run this as a dep to other targets, then eventual changes won't be committed.

api/types/crds/skupper_cluster_policy_crd.yaml
```diff
@@ -0,0 +1,65 @@
...
+                    type: string
+                allowedExposedResources:
+                  type: array
+                  items:
+                    type: string
+                allowedServices:
+                  type: array
+                  items:
+                    type: string
+                allowGateway:
```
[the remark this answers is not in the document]
fgiorgetti: That is a good point. Actually if incoming links are disabled, then you cannot create gateways at all.
Should we get rid of the AllowGateway flag and let user decide if they want to enable it or not based
on incoming links only?

cmd/service-controller/service_sync.go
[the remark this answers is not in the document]
grs: An aim of the policy is to give the cluster owner assurance about what will and won't be exposed to or by skupper. I think we need to enforce the services that are allowed in the service controller to ensure that someone cannot workaround the restrictions by directly editing skupper-services.

If we do that, then there is actually no way that a service advertised by some other site through service-sync is exposed to workloads in the cluster except through allowed service names. That being the case, I wouldn't see the need to prevent the advertised services being locally recorded.

pkg/qdr/amqp_mgmt.go
grs: No need to revert, I was just asking whether it was an explicit choice.

api/types/crds/skupper_cluster_policy_crd.yaml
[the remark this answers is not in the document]
fgiorgetti: Agreed. I will remove it.

cmd/service-controller/service_sync.go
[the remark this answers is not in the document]
fgiorgetti: It works. But after removing this, `skupper status`  and `skupper service status` will list
the unauthorized service as well.

If we also keep this one, then the unauthorized service won't be listed, otherwise, I can
add an extra validation to: `skupper service status` and `skupper network status`
so that unauthorized services are listed as "not authorized". In this case then the console/services endpoint
might need some work either.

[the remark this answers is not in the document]
grs: This is perhaps something to discuss a little more. Personally, I like the approach of updating the skupper-services definition, but flagging that the service is not creating due to policy. However others may prefer the simplicity of not allowing them in in the first place.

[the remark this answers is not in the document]
fgiorgetti: @grs Removed this validation from the service sync. Services will show up now and, if not authorized,
the respective kubernetes service won't be created and `skupper service status` as well as 
`skupper network status` will display a `not authorized` message.
@nluaces pls review again.

api/types/crds/skupper_cluster_policy_crd.yaml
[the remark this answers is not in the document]
fgiorgetti: Removed `allowGateway` from the CRD definition to make it more future proof.

api/types/types.go
```diff
@@ -75,21 +76,23 @@ var TransportPrometheusAnnotations = map[string]string{
...
+	ControllerConfigPath                 string = "/etc/messaging/"
+	ControllerServiceName                string = "skupper"
 )
 
 var ControllerPolicyRule = []rbacv1.PolicyRule{
 	{
 		Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
 		APIGroups: []string{""},
-		Resources: []string{"services", "configmaps", "pods", "secrets"},
+		Resources: []string{"services", "configmaps", "pods", "pods/exec", "secrets"},
```
grs, line 95: Why does the controller need to exec into other pods?

cmd/service-controller/service_sync.go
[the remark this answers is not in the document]
fgiorgetti: > This is perhaps something to discuss a little more. Personally, I like the approach of updating the skupper-services definition, but flagging that the service is not creating due to policy. However others may prefer the simplicity of not allowing them in in the first place.

Initially my preference was no to let them in at all, but after your brought that point of manually editing the skupper-services, I had a second thought and actually I believe having the services showing up as not-authorized makes more sense, as it will be clear to users that skupper is being notified about those services and they will know the reason why they do not work.

If those services are not listed at all, then users won't have any clue why, unless they query for internal skupper events.

api/types/types.go
fgiorgetti: This can be removed now. I have exposed an API to validate policies, which CLI is using through the get command.
At some point I used that API from inside the controller, which does not make sense, so I probably left this behind.

I will remove and retest.

fgiorgetti: Right now `ConnectorCreateSecretFromData` and `TokenClaimTemplateCreate` are using the `PolicyAPIClient` (which
uses the get command in order to verify policies from within the controller).
These VanClient methods are used by the CLI, by TokenManager and by LinkManager.

I will evaluate the impact of using the `ClusterPolicyValidator` component itself (which is meant for internal controller
 usage) and get rid of this extra resource permission.

grs (approved):

cmd/get/get.go
```diff
@@ -45,6 +61,39 @@ func simplePathCommand(path string, description string) *cobra.Command {
...
+			RunE: func(cmd *cobra.Command, args []string) error {
+				path := fmt.Sprintf("policy/%s", pathCmd)
+				for i, _ := range args {
+					path = path + "/" + args[i]
+				}
+				return get(path, output)
+			},
+		})
+	}
+	simplePathPolicyCommand("gateway", nil, "Validates if gateway can be created")
```
ajssmith, line 89: iiuc, gateway is no longer separate policy

cmd/get/get.go
```diff
@@ -45,6 +61,39 @@ func simplePathCommand(path string, description string) *cobra.Command {
...
+	simplePathPolicyCommand("incominglink", nil, "Validates if incoming links can be creted")
+	simplePathPolicyCommand("outgoinglink", []string{"hostname"}, "Validates if an outgoing link to the given hostname is allowed")
+	simplePathPolicyCommand("expose", []string{"target-type", "target-name"}, "Validates if the given resource can be exposed")
+	simplePathPolicyCommand("service", []string{"name"}, "Validates if service can be created or imported")
+	return policyCmd
```
ajssmith, line 94: Suggestion for future enhancement would be to have an "inspect" or similar that enumerates the policies applied to the namespace (e.g. or of all the policies defined for the site applied to this namespace). This would also help for namespaces not in the allow list to review what is being applied.

fgiorgetti: Created issue #698 to track this.

Issue #670, opened 2022-02-04
Avoiding router restarts


nluaces:
### Context
Skupper is currently appending new volumes to the router each time that a secret is created. In order to make the secret available for the router within the new volume, the router has to be restarted. An inconvenience of this action is that the config synchronization could be interrupted. 

One of the last improvements that has been made is the config synchronization abstraction into a sidecar container for the router, which will help to manage the secrets needed for the router without restarting it.

### Situations in which the router has to be restarted:

- Revoking access by regenerating site and claim secret
- Creating links 
- Removing links
- Creating a new service enabling TLS with http2 protocol
- When the service controller is updating service bindings and one of the new services has TLS credentials
- Deleting a service that had TLS enabled
- When the service controller is updating service bindings and the service that has to be deleted has TLS credentials

### Proposed solution
**Router creation**
At the creation of the router, an empty volume would be configured in the deployment file, and router container and config-sync sidecar would have mounted that empty volume. That way, the sidecar would be able to provide secrets for the router in the shared volume. 

**New secret creation**
When the service controller synchronizes a new service with TLS enabled, a new sslProfile is added to the skupper-internal config map, which sends an event that will be processed by the config-sync sidecar in order to create a new connector/listener, it would need access to the VAN client to get the secret data and create the required files in a specific path mapped to the empty volume that shares with the router.

**Router restart**
The config-sync sidecar would need access to the VAN client in order to get all the TLS certificates that have been specified in the router configuration, and create again the files in the shared volume.

PR #675, opened 2022-02-11
Avoid router restarts

nluaces:
Solves #670

### DIFF
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -65,7 +65,7 @@ var TransportPolicyRule = []rbacv1.PolicyRule{
 	{
 		Verbs:     []string{"get", "list", "watch"},
 		APIGroups: []string{""},
-		Resources: []string{"pods", "configmaps"},
+		Resources: []string{"secrets", "pods", "configmaps"},
 	},
 }
 
--- a/client/router_update.go
+++ b/client/router_update.go
@@ -93,6 +93,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 	addMultiportServices := false
 	addClusterPolicy := false
 	updateRouterPolicyRule := false
+	addCertsSharedVolume := false
 	inprogress, originalVersion, err := cli.isUpdating(namespace)
 	if err != nil {
 		return false, err
@@ -103,6 +104,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 		addMultiportServices = utils.LessRecentThanVersion(originalVersion, "0.8.0")
 		addClusterPolicy = utils.LessRecentThanVersion(originalVersion, "0.9.0")
 		updateRouterPolicyRule = utils.LessRecentThanVersion(originalVersion, "0.9.0")
+		addCertsSharedVolume = utils.LessRecentThanVersion(originalVersion, "0.9.0")
 	} else {
 		originalVersion = site.Version
 	}
@@ -118,6 +120,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 			if utils.LessRecentThanVersion(originalVersion, "0.9.0") {
 				addClusterPolicy = true
 				updateRouterPolicyRule = true
+				addCertsSharedVolume = true
 			}
 
 			err = cli.updateStarted(site.Version, namespace, configmap.ObjectMeta.OwnerReferences)
@@ -204,6 +207,14 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 			Subject:     types.LocalTransportServiceName,
 			Hosts:       []string{},
 			ConnectJson: true,
+		})
+		credentials = append(credentials, types.Credential{
+			CA:          types.ServiceCaSecret,
+			Name:        types.ServiceClientSecret,
+			Hosts:       []string{},
+			ConnectJson: false,
+			Post:        false,
+			Simple:      true,
 		})
 
 		usingRoutes, err = cli.usingRoutes(namespace)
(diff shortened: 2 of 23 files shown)


cmd/config-sync/config_sync.go
```diff
@@ -171,3 +202,257 @@ func (c *ConfigSync) syncConfig(desired *qdr.BridgeConfig) error {
...
 	return nil
 }
+
+func (c *ConfigSync) syncRouterConfig(desired *qdr.RouterConfig) error {
+	agent, err := c.agentPool.Get()
+	if err != nil {
+		return fmt.Errorf("Could not get management agent : %s", err)
+	}
+	var synced bool
+	for i := 0; i < 3 && err == nil && !synced; i++ {
```
fgiorgetti, line 212: Why not delegating the retry mechanism to the rate limited queue `*ConfigSync.events` ?

cmd/config-sync/config_sync.go
```diff
@@ -171,3 +202,257 @@ func (c *ConfigSync) syncConfig(desired *qdr.BridgeConfig) error {
...
+		}
+		return false, nil
+	}
+}
+
+func syncSecrets(configSync *ConfigSync, changes *qdr.BridgeConfigDifference, sharedTlsFilesDir string) error {
+	for _, added := range changes.HttpListeners.Added {
+		if len(added.SslProfile) > 0 {
+			log.Printf("Copying cert files related to HTTP Connector sslProfile %s", added.SslProfile)
+			return configSync.ensureSslProfile(added.SslProfile, added.SslProfile, sharedTlsFilesDir)
```
fgiorgetti, line 256: If something gets added here, the remaining changes are being ignored.
Maybe should only return at this point in case of an error.

cmd/config-sync/config_sync.go
```diff
@@ -171,3 +202,257 @@ func (c *ConfigSync) syncConfig(desired *qdr.BridgeConfig) error {
...
+		}
+	}
+
+	for _, added := range changes.HttpConnectors.Added {
+		if len(added.SslProfile) > 0 {
+			log.Printf("Copying cert files related to HTTP Connector sslProfile %s", added.SslProfile)
+			return configSync.ensureSslProfile(added.SslProfile, added.SslProfile, sharedTlsFilesDir)
```
fgiorgetti, line 263: ... same here

cmd/config-sync/config_sync.go
```diff
@@ -171,3 +202,257 @@ func (c *ConfigSync) syncConfig(desired *qdr.BridgeConfig) error {
...
+			log.Printf("Creating ssl profile %s", sslProfile.Name)
+			err := agent.CreateSslProfile(sslProfile)
+			if err != nil {
+				return err
+			}
+
+			log.Printf("Copying cert files related to Connector sslProfile %s", added.SslProfile)
+			secretName := strings.TrimSuffix(added.SslProfile, "-profile")
+
+			return c.ensureSslProfile(added.SslProfile, secretName, sharedTlsFilesDir)
```
fgiorgetti, line 369: It should only return on error. Otherwise the deleted entries won't be processed.

cmd/config-sync/config_sync_test.go
```diff
@@ -0,0 +1,347 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+	"os"
+	"testing"
+)
+
+const TEST_TLS_DIRECTORY = "~/.test/skupper-router/tls"
```
fgiorgetti, line 18: I would not write to `~`. It might be better to use `./tmp` or some internal location inside the cloned repo.

fgiorgetti (commented):
Link creation/deletion is working fine without restarts.

api/types/types.go
```diff
@@ -65,7 +65,7 @@ var TransportPolicyRule = []rbacv1.PolicyRule{
 	{
 		Verbs:     []string{"get", "list", "watch"},
 		APIGroups: []string{""},
-		Resources: []string{"pods", "configmaps"},
+		Resources: []string{"secrets", "pods", "configmaps"},
```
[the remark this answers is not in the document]
fgiorgetti: Maybe worth doing that if version is greater or equal to 0.9.0 (similarly to how `addClusterPolicy` logic is performed).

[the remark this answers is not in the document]
nluaces: Done, let me know if the change works for you.

cmd/config-sync/config_sync.go
nluaces: Done, thanks for mentioning this.

cmd/config-sync/config_sync.go
nluaces: Fixed! thanks!

cmd/config-sync/config_sync.go
nluaces: Fixed! thanks!

grs (commented):
If you delete the pod of a router with a link to another site, it seems to take approximately a minute for it to reconnect.

It looks like the issue (in part at least) is that config-sync fails to connect to the router on the first event (possibly because router has not yet started the local listener). It retries but gives up after 5 retries.

Either we need to keep trying to connect until successful, or we should proceed to download all certs on first event anyway, regardless of whether we have connected to router. That should reduce the time to restore connectivity.

nluaces:
@grs thank you for your input!

> Either we need to keep trying to connect until successful, or we should proceed to download all certs on first event anyway, regardless of whether we have connected to router. That should reduce the time to restore connectivity.

I agree with the second option of downloading the certs on first event, will update this pull request with that.

grs (approved):
Looks good, nice work!

client/router_update.go
```diff
@@ -515,6 +529,19 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
...
+		err = kube.UpdateRole(namespace, types.TransportRoleName, types.TransportPolicyRule, cli.KubeClient)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	if addCertsSharedVolume {
+		kube.AppendSharedVolume(&router.Spec.Template.Spec.Volumes, &router.Spec.Template.Spec.Containers[0].VolumeMounts, &router.Spec.Template.Spec.Containers[1].VolumeMounts, "skupper-router-certs", "/etc/skupper-router-certs")
+		updateRouter = true
+	}
```
fgiorgetti, line 543: This must be done before line 435: `if updateRouter || updateSite || hup`
Otherwise this change is not applied to the skupper-router deployment.

nluaces: Thanks! I will fix it.

fgiorgetti (approved):
LGTM. Nice job!

Issue #658, opened 2021-12-13
Document the `ingress` key in the `skupper-site` configMap


waldner:
I found out the hard way that one can specify the `ingress` key in the `skupper-site` configMap as an alternative to the `--ingress` command-line switch to `skupper init`.
This should probably be documented (eg here https://skupper.io/docs/declarative/index.html#yaml-reference or here https://github.com/skupperproject/skupper/blob/master/cmd/site-controller/README.md).

I can try to put together a patch/PR if you confirm that my understanding is correct.

Thanks.

PR #676, opened 2022-02-11
Added service and ingress annotation options

### LINKED ISSUE #653: Feature request: control LB IP address
(Apologies if this feature is already available, but I have not found any reference in the docs).

When skupper is deployed inside a k8s cluster, my understanding is that you only deploy the skupper-site-controller deployment and pod, which then does all the hard work of bringing up the other deployments, pods and services.

In particular, if the loadbalancer ingress is deployed, skupper creates a `skupper-router` service of type `LoadBalancer` with a randomly assigned IP address (one of the available IPs in the LB pool). While this could be fine as default behavior, would it be possible to have an option to choose exactly which IP address is assigned to the service? This would be useful for example if you need to associate a DNS entry to the IP, in which case you need to know the IP in advance.
Thanks

ThomasK33:
Fixes #653

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -311,6 +311,10 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 		}
 	}
 
+	for key, value := range options.Controller.ServiceAnnotations {
+		annotations[key] = value
+	}
+
 	svcs := []*corev1.Service{}
 	if len(controllerPorts) > 0 {
 		svcs = append(svcs, &corev1.Service{
@@ -730,7 +734,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		routerIngressHost := options.GetRouterIngressHost()
 		controllerIngressHost := options.GetControllerIngressHost()
 		post := false // indicates whether credentials need to be revised after creating appropriate ingress resources
-		if options.IsIngressNginxIngress() {
+		if options.IsIngressNginxIngress() || options.IsIngressKubernetes() {
 			if routerIngressHost != "" {
 				routerHosts = append(routerHosts, strings.Join([]string{"inter-router", van.Namespace, routerIngressHost}, "."))
 				routerHosts = append(routerHosts, strings.Join([]string{"edge", van.Namespace, routerIngressHost}, "."))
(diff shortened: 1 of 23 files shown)


ThomasK33:
Also fixes #658

ThomasK33:
Also added a missing step for removing the config-sync image in the `remove_from_registry` job.

client/router_create.go
```diff
@@ -327,7 +331,13 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 				Ports:    controllerPorts,
 				Type:     svctype,
 			},
-		})
+		}
+
+		if options.Controller.LoadBalancerIp != "" {
```
grs, line 336: Should this also be conditional on the type being lodadbalancer?

client/router_create.go
```diff
@@ -900,7 +911,13 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 				},
 				Type: svcType,
 			},
-		})
+		}
+
+		if options.Router.LoadBalancerIp != "" {
```
grs, line 916: As above, should this be conditional on service type being loadbalancer?

client/router_create.go
ThomasK33: Technically it shouldn't make a difference, as the field only applies to `LoadBalancer` service types, and others ignore it.
Added the additional check nonetheless.

grs (approved):

ThomasK33:
Hey @grs @fgiorgetti, any chance this can get merged & released?

client/site_config_inspect.go
```diff
@@ -115,6 +115,9 @@ func (cli *VanClient) SiteConfigInspectInNamespace(ctx context.Context, input *c
 			result.Spec.Ingress = cli.GetIngressDefault()
 		}
 	}
+	if ingressAnnotations, ok := siteConfig.Data[SiteConfigIngressAnnotationsKey]; ok {
+		result.Spec.IngressAnnotations = asMap(splitWithEscaping(ingressAnnotations, ',', '\\'))
```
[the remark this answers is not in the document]
ThomasK33: Yes, as the formating of annotations is a flat string.

E.g.:
```go
ingressAnnotations := "kubernetes.io/ingress.class=nginx,tag1=abc,tag2=123"
```

When stripping the escape character, this would result in a map of:
- kubernetes.io/ingress.class=nginx
- tag1=abc
- tag2=123

If one does not strip the escape character, this will result in the following annotations:
- kubernetes.io/ingress.class=nginx,
- tag1=abc,
- tag2=123

Additionally, if one explicitly needs a comma in an annotation, it can be escaped using `\\,` to be part of the annotation and not split into separate ones.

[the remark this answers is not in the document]
fgiorgetti: Sounds fair. After going through it, I believe the original `asMap` under `cmd/skupper/skupper.go` might
need an enhancement task, so that it can escape characters properly, as today it simply splits by `=`.
But that is not a concern for your PR.

Thanks @ThomasK33 !

fgiorgetti (approved):

PR #684, opened 2022-03-07
Refactored service-sync

grs:


grs:
Ideas here are:

* separate out kubernetes related stuff from non-kubernetes related stuff
* fix concurrency issues
* move things out of the cmd sub directory into separate packages
* make things a little easier to unit test

nluaces (approved):
lgtm, thanks!

pkg/service_sync/service_sync.go
```diff
@@ -0,0 +1,266 @@
...
+		service := types.ServiceInterface{
+			Address:      original.Address,
+			Protocol:     original.Protocol,
+			Ports:        original.Ports,
+			Origin:       original.Origin,
+			Headless:     original.Headless,
+			Labels:       original.Labels,
+			Aggregate:    original.Aggregate,
+			EventChannel: original.EventChannel,
+			Targets:      []types.ServiceInterfaceTarget{},
```
fgiorgetti, line 112: Missing `EnableTls` and `TlsCredentials`.

pkg/service_sync/service_sync.go
```diff
@@ -0,0 +1,266 @@
...
+
+	return ServiceUpdate{
+		origin:      c.origin,
+		version:     c.version,
+		definitions: latest,
+	}
+}
+
+func (c *ServiceSync) updateRemoteDefinitions(origin string, serviceInterfaceDefs map[string]types.ServiceInterface) {
+	event.Recordf(ServiceSyncEvent, "Received update from site %s", origin)
```
fgiorgetti, line 139: This is going to generate lots of events. Is it desired or should it only generate an event
when something has changed?

fgiorgetti (commented):
A few minor comments, but overall it looks much better.

pkg/service_sync/service_sync.go
nluaces: I missed that, thank you

fgiorgetti (approved):

pkg/service_sync/service_sync.go
grs: It was intentional, but it may not be correct. My thinking was (and this came from debugging a bug), that it gives a good indication that communication is still active. Events from the same site will all be coalesced. It is true that for networks with several sites, these updates will quickly obliterate others however. So perhaps a better solution to ensuring communication is occurring is needed.

grs: I've removed that for now. Updates are still recorded (and this patch fixes them so that it is not just changes to services with local targets that are reported).

pkg/qdr/messaging.go
```diff
@@ -0,0 +1,105 @@
+package qdr
```
ajssmith, line 1: An observation,  might this be better placed in messaging pkg as it does not contain qdr capabilties?

grs: The aim with the messaging package was to be a fairly pure interface. In this package we have a particular implementation that can be used to talk through qdr. I agree that strictly speaking it is an amqp implementation and not tied to the router, but the intent of it is use with the router.

Having the interface is mostly just to ease testing. It also helps reduce direct dependencies between components.

ajssmith (approved):
lgtm, tried locally, ran examples and tests, thnx

PR #692, opened 2022-03-11
Returned Retry() to previous logic, created RetryError

hash-d:
`Retry()`'s workings was not clear from its documentation, and it was being used differently by different parts of the code.

Originally, it was to be interpreted as _retry until f() bool response is true; fail immediatelly on f() error_.  A different understanding was _retry f() until it stops erroring out_.

The function was being used in the second sense in `skupper_network.go`, but it was not working as intended, so it was fixed.  That, however, changed its meaning for other code that uses the same function.  Only test code was affected by that.

This PR brings `Retry()` back to its original meaning and adds `RetryError()` for use by `skupper_network.go` and other such uses.

The documentation of `Retry()` was also updated to make its purpose clear.

### DIFF
--- a/pkg/utils/retry.go
+++ b/pkg/utils/retry.go
@@ -22,10 +22,19 @@ import (
 
 type ConditionFunc func() (bool, error)
 
+type CheckedFunc func() error
+
 // Retry retries f every interval until after maxRetries.
+//
 // The interval won't be affected by how long f takes.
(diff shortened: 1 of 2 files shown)


nluaces (approved):
Looks good to me, thank you

fgiorgetti (approved):

PR #694, opened 2022-03-14
enable route host name

ajssmith:
Customer request to provide custom host name for the route rather than have OpenShift auto generate. The route host should be the concatenation of service name, namespace and host name suffix. Use the existing --ingress-host parameter to specify when ingress is type route.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -218,7 +218,7 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 					TargetPort: intstr.FromInt(int(types.ConsoleOpenShiftOauthServiceTargetPort)),
 				}
 			}
-			host := options.GetRouterIngressHost()
+			host := options.GetControllerIngressHost()
 			if host != "" {
 				host = types.ConsoleRouteName + "-" + van.Namespace + "." + host
 			}
@@ -290,7 +290,7 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 			Port:     types.ClaimRedemptionPort,
 		})
 		if options.IsIngressRoute() {
-			host := options.GetRouterIngressHost()
+			host := options.GetControllerIngressHost()
 			if host != "" {
 				host = types.ClaimRedemptionRouteName + "-" + van.Namespace + "." + host
 			}
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -218,6 +218,10 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 					TargetPort: intstr.FromInt(int(types.ConsoleOpenShiftOauthServiceTargetPort)),
 				}
 			}
+			host := options.GetRouterIngressHost()
+			if host != "" {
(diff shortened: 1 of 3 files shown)


client/router_update.go
```diff
@@ -320,6 +320,11 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 			// routes: skupper-controller -> skupper
 			original, err := cli.RouteClient.Routes(namespace).Get("skupper-controller", metav1.GetOptions{})
 			if err == nil {
+				siteConfig, err := cli.SiteConfigInspect(ctx, nil)
+				host := siteConfig.Spec.GetRouterIngressHost()
```
grs, line 324: Do we want to change this on update? I think probably not. It will not affect the certificates unless those are recreated anyway. I think the simpler and perhaps safer thing is just to say that this enhancement requires re-initialisation of the site.

client/router_create.go
```diff
@@ -218,6 +218,10 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 					TargetPort: intstr.FromInt(int(types.ConsoleOpenShiftOauthServiceTargetPort)),
 				}
 			}
+			host := options.GetRouterIngressHost()
```
grs, line 221: Should this be IsConsoleIngressRoute()?

client/router_create.go
```diff
@@ -285,6 +290,10 @@ func (cli *VanClient) GetVanControllerSpec(options types.SiteConfigSpec, van *ty
 			Port:     types.ClaimRedemptionPort,
 		})
 		if options.IsIngressRoute() {
+			host := options.GetRouterIngressHost()
```
grs, line 293: Here again, should this be GetControllerIngressHost()? Likely in most cases these will be the same. Not even sure if we need or want to have separate options TBH.

client/router_update.go
```diff
@@ -1000,6 +1000,14 @@ func (cli *VanClient) createClaimsServerSecret(ctx context.Context, namespace st
...
 
 func (cli *VanClient) createClaimsRedemptionRoute(ctx context.Context, namespace string) error {
+	siteConfig, err := cli.SiteConfigInspect(ctx, nil)
+	if err != nil {
+		return err
+	}
+	host := siteConfig.Spec.GetRouterIngressHost()
+	if host != "" {
+		host = types.ClaimRedemptionRouteName + "-" + namespace + "." + host
+	}
```
grs, line 1010: Just as a philosophical point, I don't think we needed to do anything here as this is only used when upgrading from an old version where there was no expectation that the host could be set on routes. That said I can't see that it will cause any harm so fine to leave it.

fgiorgetti (approved):
Other than @grs comments, it seems to work fine.
I have tested it using custom CNAME records and it seems to work fine.

client/router_create.go
ajssmith: I went back and forth but agree it makes sense to allow separation for now. Will update.

client/router_update.go
ajssmith: Agree, will remove

grs (approved):

grs (approved):

PR #700, opened 2022-03-17
Renamed skupper-router resources

fgiorgetti:


grs (commented):
LGTM... on update, is there anything else we need to consider? E.g. the mount paths of certs? Presumably the config file will have the old paths in them and that may work, will newly added links also work?

Does the router network form ok with skupper-router and qdrouterd? If not we should at least advise of that somehow.

grs (approved):

fgiorgetti:
> I wonder why service-sync does not work... that does not directly use management, just messaging. It suggests that messages don't flow between the routers. If that is the case then there is no value in trying to do anything for backward compatibility in the controller on its own.

That is what it looks like. The old router has some error messages, like:
`Received MAR at protocol version 2 from sk3-skupper-router-6bd75894d5-9td87.  Ignoring.` or even
`Error performing QUERY: No such entity type 'io.skupper.router.connection'`.

fgiorgetti:
> I will research what else needs to be updated. There are environment variables and mount paths in the skupper-router deployment that certainly need to be updated.

I have added extra update changes needed and tested updating three connected sites (with services)
from 0.8.7 to the latest.

All participating sites must be updated.

nluaces (approved):

grs (approved):

PR #701, opened 2022-03-21
Site controller fixes permissions issues when policy is enabled

fgiorgetti:


cmd/site-controller/controller.go
```diff
@@ -209,13 +220,21 @@ func (c *SiteController) checkSite(key string) error {
...
 		log.Println("Error checking skupper-site config map: ", err)
 		return err
 	} else if exists {
 		configmap := obj.(*corev1.ConfigMap)
+		// if policy is enabled, ensure ClusterRoleBinding exists even for not controlled sites
+		if c.validator.Enabled() {
+			c.addClusterRoleBinding(configmap)
+		}
+		// If site is not controlled, ignore it
+		if !c.isSiteControlled(configmap.ObjectMeta.Namespace) {
```
grs, line 235: Why not just check configmap above rather than re-retrieving it from the api server?

cmd/site-controller/deploy-watch-all-ns.yaml
```diff
@@ -77,6 +83,7 @@ rules:
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
+  - clusterrolebindings
```
grs, line 86: What do we actually need here?

cmd/site-controller/deploy-watch-all-ns.yaml
```diff
@@ -86,6 +93,14 @@ rules:
   - create
   - delete
   - update
+- apiGroups:
```
grs, line 96: I wonder if we should have separate yamls for site controller with and without policy?

fgiorgetti: From a technical standpoint, this can be applied without issues, even if the CRD is not defined.
One thing that might be impacted though if we separate them, would be the operator.

grs: Ok, if it won't cause errors when those CRDs do not exist, then we don't need separate yaml.

cmd/site-controller/deploy-watch-all-ns.yaml
grs: Can we detect if the skupper policy CRD exists without having permission to read it?

grs: Ideally, if the CRD is installed, nothing will work if the service controller does not have permission to read policy. Otherwise you can avoid the policy entirely by initialising without the necessary permissions.

grs: But of course, if the CRD is not installed, then ideally everything would work as before.

fgiorgetti: > Can we detect if the skupper policy CRD exists without having permission to read it?

No. Today policy component check if it has permission to read it and if it is defined. In case
any of the validations is false, policy will be considered as disabled.

>Ideally, if the CRD is installed, nothing will work if the service controller does not have permission to read policy. Otherwise you can avoid the policy entirely by initialising without the necessary permissions.

That is how it is supposed to work. If permission is revoked policy will be considered as disabled.

nluaces (approved):

cmd/site-controller/deploy-watch-all-ns.yaml
grs: > Today policy component check if it has permission to read it and if it is defined. In case any of the validations is false, policy will be considered as disabled.

I think we need to change that. Could we make permission to read the CRD whether or not it exists (not a sepecific CR) a strict requirement? That way we would always be able to tell if it is installed or not. If not policy is disabled, if it is policy is enforced and inability to read a matching policy whether due to permission or anything else, would prevent the operation in question.

This would mean of course that you would have to have the extra permission to use skupper even without policy. However as is, it seems policy cannot be reliably enforced.

grs: I think if the CRD is defined, whether or not we have permission to read any associated CRs, we should apply policy. In other words the test for whether policy is enabled should simply be the discovery of the CRD. E.g. the test would be something like this:

```
func isPolicyEnabled(cli *VanClient) bool {
	dc, err := discovery.NewDiscoveryClientForConfig(cli.RestConfig)
	resources, err := dc.ServerResourcesForGroupVersion("skupper.io/v1alpha1")
	if errors.IsNotFound(err) {
		log.Printf("Policy is NOT enabled")
		return false
	} else if err != nil {
		log.Printf("Cannot determine if policy is enabled: %s", err)
		return true
	} else {
		for _, r := range resources.APIResources {
			if r.Kind == "SkupperClusterPolicy" {
				log.Printf("Policy IS enabled")
				return true
			}
		}
		log.Printf("Policy is NOT enabled")
		return false
	}
}
```

That way, once the CRD is installed, you require the cluster role to be able to use skupper at all. If you don't have it, it is there are no matching rules that can be read and so nothing is enabled.

fgiorgetti: @grs yes, it works just fine! Thanks for advising. I will update the main policy component to use
the discovery client instead and enforce policy in spite of permission being granted.

cmd/site-controller/controller.go
```diff
@@ -403,3 +414,43 @@ func (c *SiteController) updateChecks() {
...
+	for {
+		select {
+		case <-period.C:
+			if c.validator.Enabled() && !state {
+				state = true
+				log.Println("policy is enabled, validating permissions")
+				for _, site := range c.siteInformer.GetStore().List() {
+					// Check if ClusterRoleBinding exists for namespace
+					cm := site.(*corev1.ConfigMap)
+					c.addClusterRoleBinding(cm)
```
grs, line 430: router_update will add the cluster role bindings on an update, and do so regardless of whether the CRD is actually installed, is right? And in the yaml for site controller below we bind the cluster role to the site controller account unconditionally also.

I'm not clear on why we need this conditional logic here? What is the use case it is anticipating?

cmd/site-controller/deploy-watch-all-ns.yaml
```diff
@@ -86,6 +86,20 @@ rules:
...
+  verbs:
+  - create
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - clusterroles
+  verbs:
+  - bind
+  resourceNames:
+  - skupper-service-controller
```
grs, line 102: This is not really a comment on this PR, but I missed it earlier and now I've been looking at this more closely it seemed worth suggesting: is there a better name for the role? E.g. skupper-policy-reader?

fgiorgetti: If you find it clearer, I will do that in a separate PR.

grs: I think it is better if the name describes what the role allows. Makes people less nervous about binding that role. Its a minor thing, but if you get a chance to do a separate PR with the change I think it would be a small but significant improvement.

fgiorgetti: I agree. I will raise an issue and work on it. Thanks, I think it makes it clearer for sure.

fgiorgetti: Created #706 to track this.

cmd/site-controller/controller.go
fgiorgetti: @grs done.

grs (approved):

ajssmith (approved):
lgtm, thanks

PR #703, opened 2022-03-22
Fixes namespace label expression for policies

### LINKED ISSUE #702: Policy namespace selection by label is not accepting labels that contain a slash
To reproduce:

* Create a namespace with the label `test.skupper.io/test-namespace=policy`
* Add the policy below
* `skupper token create` will fail, whereas it should have worked
* If only `test-namespace=policy` is used, for both the namespace label and policy, token creation works as expected.

```
apiVersion: skupper.io/v1alpha1              
kind: SkupperClusterPolicy                   
metadata:                                    
  name: cluster-policy-sample-01             
spec:                                        
  namespaces:                                
    - "test.skupper.io/test-namespace=policy"
  allowIncomingLinks: true                   
  allowedExposedResources:                   
    - "*"                                    
  allowedOutgoingLinksHostnames: []          
  allowedServices:                           
    - "*"                                    
```

According to the label parsing documentation [1], label keys may have at most one slash.

Additionally, according to [2], no underscores should be accepted on the key part of the label.

[1] https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#Parse
[2] https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

fgiorgetti:
Fixes #702

### DIFF
--- a/client/policy_validator.go
+++ b/client/policy_validator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/skupperproject/skupper/pkg/event"
 	"github.com/skupperproject/skupper/pkg/generated/client/clientset/versioned/typed/skupper/v1alpha1"
 	"github.com/skupperproject/skupper/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -86,8 +87,9 @@ func (p *ClusterPolicyValidator) LoadNamespacePolicies() ([]v1alpha12.SkupperClu
 	if err != nil {
 		return policies, err
 	}
+	namespace, _ := p.cli.KubeClient.CoreV1().Namespaces().Get(p.cli.Namespace, v1.GetOptions{})
 	for _, pol := range policyList.Items {
-		if p.appliesToNS(&pol) {
+		if p.appliesToNS(&pol, namespace) {
 			policies = append(policies, pol)
 		}
 	}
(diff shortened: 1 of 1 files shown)


client/policy_validator.go
```diff
@@ -104,22 +104,21 @@ func (p *ClusterPolicyValidator) AppliesToNS(policyName string) bool {
...
-		if p.labelRegex.MatchString(ns) {
-			namespace, err := p.cli.KubeClient.CoreV1().Namespaces().Get(p.cli.Namespace, v1.GetOptions{})
-			if err == nil {
-				selector, _ := labels.Parse(ns)
-				if selector.Matches(labels.Set(namespace.Labels)) {
-					namespaces = append(namespaces, namespace.Name)
-				}
+		if ns == "*" {
+			return true
+		} else if selector, err := labels.Parse(ns); nsErr == nil && err == nil {
```
grs, line 111: Minor issue, but I find the flow here a bit odd and hard to follow. It wasn't obvious why nsErr was tested here in the loop for example.

I'd suggest the test for "*" first, as if that is the case there is no need to do anything else.

Then try to match the namespaces in policy as regexes against the actual name. Only if that all fails do you then need to try get the labels of the current namespace and try the policy to see if any of those are label selectors.

Alternatively you could determine which in the namespaces list on policy were label selectors and which not first, then get the current namespace labels if there is at least one valid selector, then test each entry in turn until a match.

I do wonder whether this is all a bit too flexible. However a counter argument is that it is nice to be able to do some kind of matching without granting the permission to read namespaces. Label selectors is a nice pattern but not without cost.

fgiorgetti: OK, I have just pushed a simplified version. See if it looks better now.
One change to reduce the cost of reading namespace is that the Namespace instance must be provided now,
which might help in the LoadNamespacePolicies case, as the Namespace is retrieved only once.

PR #708, opened 2022-04-03
add site verify step

ajssmith:
Check for self and duplicates

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -143,7 +143,7 @@ func (cli *VanClient) ConnectorCreateSecretFromData(ctx context.Context, secretD
 				return nil, err
 			}
 			// Verify if site link can be created
-			err = cli.verifyNotSelfOrDuplicate(secret, siteConfig.Reference.UID)
+			err = cli.verifyNotSelfOrDuplicate(secret, siteConfig.Reference.UID, options)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


client/connector_create.go
```diff
@@ -275,6 +213,33 @@ func (cli *VanClient) VerifySiteCompatibility(siteVersion string) error {
...
+		return fmt.Errorf("Can't create connection to self with token")
+	}
+	currentSecrets, err := cli.KubeClient.CoreV1().Secrets(options.SkupperNamespace).List(metav1.ListOptions{LabelSelector: "skupper.io/type=connection-token"})
+	if err != nil {
+		return fmt.Errorf("Could not retrieve secrets: %w", err)
+	}
+	for _, currentSecret := range currentSecrets.Items {
+		currentAuthor, ok := currentSecret.Annotations[types.TokenGeneratedBy]
+		if !ok {
+			return fmt.Errorf("A secret has no author.")
```
fgiorgetti, line 234: I see Skupper was already doing this earlier, but should it really return an error or
simply log an event and continue to the next iteration?

ajssmith: That's a good question. If a secret has the connection-token label but does not have the annotation, it is exceptional based on what we currently expect (e.g. it "shouldn't" happen).

grs:
Just a thought: should these checks be performed by the CLI before creating the secret, by the controller that processes the secret and configures the router to create the necessary link, or both?

client/connector_create.go
fgiorgetti: True, it is an internal error or maybe some incorrect manual modification.

My concern was that new links would not be created, unless this is fixed,
but I believe it is better to fail fast in this case otherwise the event would
not be noticed.

nluaces:
Based on #675  PR, if the config-sync is the only one verifying the secrets before creating a new connector in the router, at that point the skupper-internal configmap is already updated with the new desired connector for the link. 

I think that having these checks in the CLI is useful to avoid unnecessary changes in the config-map; and at the same time, it looks like the config-sync has the responsibility of sending the correct configuration to the router. 

Maybe both?

grs (approved):

fgiorgetti (approved):

PR #716, opened 2022-04-08
Adjusted incorrect Policy CRD verification

fgiorgetti:


nluaces (approved):

client/policy_validator.go
```diff
@@ -25,13 +24,12 @@ import (
 
 type PolicyValidationResult struct {
 	err             error
+	crdDefined      bool
```
grs, line 27: Maybe 'enabled' would be a better name for this field?

fgiorgetti: Certainly more friendly. I will change it.

client/policy_validator.go
```diff
@@ -61,13 +59,12 @@ type ClusterPolicyValidator struct {
 	cli           *VanClient
 	dc            *discovery.DiscoveryClient
 	skupperPolicy v1alpha1.SkupperClusterPolicyInterface
-	labelRegex    *regexp.Regexp
+	mockEnabled   bool
```
grs, line 62: Myabe 'disablePolicyDiscovery' would be a nicer name for this?

fgiorgetti: +1

client/policy_validator.go
```diff
@@ -171,7 +168,13 @@ func (p *ClusterPolicyValidator) HasPermission() bool {
 }
 
 func (p *ClusterPolicyValidator) CrdDefined() bool {
+	if p.mockEnabled {
+		return true
+	}
 	if p.dc == nil {
+		if p.cli.RestConfig == nil {
+			return false
```
grs, line 176: Is this the right result? Feels like it is safer to fail the attempted operation if we cannot determine whether or not policy is enabled?

grs (approved):

client/policy_validator.go
fgiorgetti: The `RestConfig` attribute is always set, except when using `newMockClient` to create the `*VanClient` instance.
When it happens (mock tests only), we have to consider CRD as absent, so existing unit tests continue to work.

That's why we now have the `disablePolicyDiscovery` flag. If a unit test needs the policy validator to consider
CRD as present, we set it to true.

PR #717, opened 2022-04-11
Fix: Connectors sync prevents automesh of the router replicas

nluaces:
Context: 
When synchronising connectors (associated with skupper links), it was assumed that connectors included in the router but not included in the updated config-map for the router should be deleted; but when skupper is initiated with more than one replica, the skupper-internal sslProfile and the connector that uses it are deleted in the replicas for this reason.

- Certificate files shared between the config-sync sidecar and the router will be stored in a new path, to not interfere with the current lifecycle of other certificates.
- When detecting connectors to delete, the connectors with profiles like skupper-internal or skupper-amqps will be filtered and therefore, not deleted.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -567,7 +567,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		AuthenticatePeer: true,
 	})
 
-	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-config-router-certs",
+	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-shared-certs",
 		qdr.SslProfile{
 			Name: types.ServiceClientSecret,
 		})
@@ -697,7 +697,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		}
 	}
 
-	kube.AppendSharedVolume(&volumes, &mounts[qdrouterd], &mounts[configSync], "skupper-router-certs", "/etc/skupper-config-router-certs")
+	kube.AppendSharedVolume(&volumes, &mounts[qdrouterd], &mounts[configSync], "skupper-router-certs", "/etc/skupper-shared-certs")
 
 	van.Transport.Volumes = volumes
 	van.Transport.VolumeMounts = mounts
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -567,7 +567,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		AuthenticatePeer: true,
 	})
 
-	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-router-certs",
+	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-config-router-certs",
 		qdr.SslProfile{
 			Name: types.ServiceClientSecret,
 		})
@@ -697,7 +697,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		}
 	}
 
-	kube.AppendSharedVolume(&volumes, &mounts[qdrouterd], &mounts[configSync], "skupper-router-certs", "/etc/skupper-router-certs")
+	kube.AppendSharedVolume(&volumes, &mounts[qdrouterd], &mounts[configSync], "skupper-router-certs", "/etc/skupper-config-router-certs")
 
 	van.Transport.Volumes = volumes
 	van.Transport.VolumeMounts = mounts
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -63,7 +63,7 @@ func InteriorListener(options types.SiteConfigSpec) qdr.Listener {
 		Name:             "interior-listener",
 		Role:             qdr.RoleInterRouter,
 		Port:             types.InterRouterListenerPort,
-		SslProfile:       types.InterRouterProfile,
+		SslProfile:       types.InterRouterProfile, //The skupper-internal profile needs to be filtered by the config-sync sidecar, in order to avoid deleting automesh connectors
 		SaslMechanisms:   "EXTERNAL",
 		AuthenticatePeer: true,
 		MaxFrameSize:     options.Router.MaxFrameSize,
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -567,7 +567,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		AuthenticatePeer: true,
 	})
 
-	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-shared-certs",
+	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-router-certs",
 		qdr.SslProfile{
 			Name: types.ServiceClientSecret,
 		})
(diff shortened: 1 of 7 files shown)


cmd/config-sync/config_sync.go
```diff
@@ -31,7 +32,7 @@ type ConfigSync struct {
 	vanClient *client.VanClient
 }
 
-const SHARED_TLS_DIRECTORY = "/etc/skupper-router-certs"
+const SHARED_TLS_DIRECTORY = "/etc/skupper-config-router-certs"
```
fgiorgetti, line 35: What do you think of using something that makes it clear it has certs for links only, like: `skupper-router-link-certs` ?

nluaces: You are right about having a more explicit name, but It is going to be also used for all the certificates that the config-sync sidecar downloads for the router, links and httpConnectors/httpListeners (when the tls feature is enabled).

cmd/config-sync/config_sync.go
```diff
@@ -248,6 +251,24 @@ func syncRouterConfig(agent *qdr.Agent, desired *qdr.RouterConfig, c *ConfigSync
...
+	result.AddedSslProfiles = make(map[string]qdr.SslProfile)
+	for _, v1 := range changes.Added {
+		result.Added = append(result.Added, v1)
+		result.AddedSslProfiles[v1.SslProfile] = changes.AddedSslProfiles[v1.SslProfile]
+	}
+
+	for _, v1 := range changes.Deleted {
+		connector, _ := agent.GetConnectorByName(v1)
+
+		if !utils.StringSliceContains([]string{"skupper-internal", "skupper-amqps"}, connector.SslProfile) {
```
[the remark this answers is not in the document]
nluaces: Thanks, I will take a look on that!

cmd/config-sync/config_sync.go
fgiorgetti: You're right. I forgot about the TLS certs. Yes, it sounds clearer to me.

grs (commented):
The overall idea is good and I'm not necessarily opposed to the specific means of filtering. However my initial thought is that something simpler and more direct will also be safer/more intuitive.

One possibility is simply to filter out from deletion list connectors with 'skupper-internal' as the sslprofile name, as that clearly implies connection to routers of the local deployment. That has the advantage of being an easy change.

Perhaps even clearer would be to change the automeshing logic to use an explicit name for the automesh added connectors, following some pattern, e.g. automesh-<host>. However the downside of that is it couples the change to the router to a change to the controller. (Then again this might be a good time for that since there are already many such changes in the pending release).

(In the long term, we could even move the automeshing logic into config-sync in some way, as it would be able to do a better job. But that is certainly for another time).

cmd/config-sync/config_sync.go
```diff
@@ -232,6 +232,8 @@ func syncRouterConfig(agent *qdr.Agent, desired *qdr.RouterConfig, c *ConfigSync
 	}
 
 	differences := qdr.ConnectorsDifference(actual, desired)
+	differences = filterInternalRouterConnectors(differences, agent, SHARED_TLS_DIRECTORY)
```
grs, line 235: Minor point, feel free to ignore. To me, it would be clearer that it is only filtering the deleted connectors if only the deleted list was passed in, e.g. something like:

```differences.Deleted = filterInternalRouterConnectors(differences.Deleted, agent, SHARED_TLS_DIRECTORY)```

Comment (nluaces) [reply cmd/config-sync/config_sync.go:235]: I agree, makes the code much clearer

Comment (grs) [reply cmd/config-sync/config_sync.go:235]: I'd possibly even consider putting the filtering directly into the qdr.ConnectorsDifference to save having to lookup the connector by name again.

**Comment by nluaces** [comment]
> One possibility is simply to filter out from deletion list connectors with 'skupper-internal' as the sslprofile name, as that clearly implies connection to routers of the local deployment. That has the advantage of being an easy change.

@grs I agree with you, and this was the first approach; but after talking with @fgiorgetti, we agreed on filtering by path just in case in the future new sslProfiles similar to skupper-internal where included in the router, in order to avoid side effects (like connectors being deleted). What do you think?

File: cmd/config-sync/config_sync.go (hunk 3 above)
Comment (nluaces) [reply cmd/config-sync/config_sync.go:235]: You are right! I will work on that

**Comment by grs** [comment]
> > One possibility is simply to filter out from deletion list connectors with 'skupper-internal' as the sslprofile name, as that clearly implies connection to routers of the local deployment. That has the advantage of being an easy change.
> 
> @grs I agree with you, and this was the first approach; but after talking with @fgiorgetti, we agreed on filtering by path just in case in the future new sslProfiles similar to skupper-internal where included in the router, in order to avoid side effects (like connectors being deleted). What do you think?

I think any solution here is going to be a little bit of a hack because the automesh 'feature' is a bit of a hack. There is some fragility in using the special 'skupper-internal' sslProfile name as identifying automesh connectors, in that if someone changes the profile name used[1] for the listener the automesh connectors would start getting deleted again. We could mitigate that by (a) adding a comment at [1] and (b) using the constant used there when filtering out the automesh connectors. The advantage is that the scope of the change is small.

Perhaps the most obvious is to add a name to the connectors at [2] and filter on that. E.g. print("  config-sync-ignore-%s", host) then ignore anything with config-sync-ignore as prefix. The only disadvantage of this option is that it affects two different components. Overall the scope is small though, the intent is obvious and it would be fairly robust.

Your proposed change is also ok. Its less fragile than relying on skupper-internal as sslProfile name. To me though, the intent is also less obvious and the scope is a little broader.

It is a trade-off in any case, and probably just a question of personal preference. I am not opposed to the current change.

[1] https://github.com/skupperproject/skupper/blob/master/client/router_create.go#L66
[2] https://github.com/skupperproject/skupper-router/blob/main/scripts/auto_mesh.py#L126

**Comment by nluaces** [comment]
@grs I think that your idea of changing the name of the connectors in order to filter them would be the best option. But because this implies a change in the scripts of the router that are located in a different repo,  do you mind if this is done in a different pull request? Just in case I encounter any unexpected issue at least the automesh problem is fixed.

I changed back to filter the connectors to delete by sslProfile and reduced the number of queries to the router about local connectors. This time I have not filtered by sslProfile `skupper-amqps` as well, becase I have realised that there are no connectors created with that profile, only listeners.

File: client/router_create.go [hunk 4]
```diff
@@ -567,7 +567,7 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 		AuthenticatePeer: true,
 	})
 
-	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-router-certs",
+	routerConfig.AddSimpleSslProfileWithPath("/etc/skupper-shared-certs",
```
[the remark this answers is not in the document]
nluaces: I thought that if the config-sync can not access to the skupper internal certs (because the shared volume is mapped to a different path), would be safer in the sense of making sure that the config-sync can only access to the certs it manages.

On the other hand, I understand that having all the certs that the router needs in the same place is more clear.

If you do not see any issue on sharing the path between volume secrets and the new empty volume, I can change it.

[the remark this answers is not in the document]
nluaces: Done!

fgiorgetti (approved):
Nice!

ajssmith (approved):
lgmt, tried locally, --routers 3 verified service mesh and added/removed site connectors

Issue #713, opened 2022-04-08
Skupper gateway on Docker fails with at least 0.8.2-0.8.7 due to directory permissions


hash-d:
# How to reproduce

* On a skupper-enabled namespace, run `skupper gateway init --type docker`
* Skupper will report the gateway as created: `Skupper gateway 'hostname' created. Use 'skupper gateway status' to get more informaiton`.
* However, the container will keep restarting:

```
$ docker ps
CONTAINER ID        IMAGE                                                                                 COMMAND                 CREATED             STATUS                         PORTS               NAMES
22ab5076c81d        registry/skupper-tech-preview-qdr-rhel8:latest   "/opt/interconnect..."  36 seconds ago      Restarting (1) 7 seconds ago                       hostname
$ 
```

* The logs show that the router is unable to find the configuration file.

```
$ docker logs 22a
No such file or directory: /opt/skupper/config/qdrouterd.json
2022-04-08 16:40:04.168150 +0000 ERROR (error) Configuration: Configuration file was empty
2022-04-08 16:40:04.168211 +0000 MAIN (critical) Router start-up failed: Configuration: Configuration file was empty
qdrouterd: Configuration: Configuration file was empty
```

On the automated testing, the issue presents itself as below:

 ```
=== RUN   TestGateway/local-gateway-docker
=== RUN   TestGateway/local-gateway-docker/local-gateway-setup
2022/04/07 13:24:14 Running Skupper Command Tester scenario: local-gateway-setup
2022/04/07 13:24:14 Running: skupper --namespace public-acceptance-1 --kubeconfig /home/cloud-user/jenkins/kubeconfig.pub.1 gateway init --type docker
2022/04/07 13:25:33 local-gateway-setup has failed, exiting
2022/04/07 13:25:33 STDOUT:
Skupper gateway 'amq-interconnect-builder-273-cloud-user' created. Use 'skupper gateway status' to get more informaiton.
2022/04/07 13:25:33 STDERR:
2022/04/07 13:25:33 Running Skupper Command Tester scenario: local-gateway-teardown
2022/04/07 13:25:33 Running: skupper --namespace public-acceptance-1 --kubeconfig /home/cloud-user/jenkins/kubeconfig.pub.1 gateway unforward tcp-echo-cluster --protocol tcp
=== RUN   TestGateway/local-gateway-podman
--- FAIL: TestGateway (151.74s)
    --- PASS: TestGateway/local-gateway-service (66.45s)
[...]

PR #720, opened 2022-04-13
Skipping container test due to known issue #713

hash-d:
There is a known issue described in #713 that has already been fixed on master.  The issue impacts only container images that specify a user to run the entry point, which is not the case for upstream.

This PR skips the tests affected by that issue, so they do no break testing downstream.

Signed-off-by: Danilo Gonzalez Hashimoto <dhashimo@redhat.com>


hash-d:
@ajssmith, this is the change we discussed yesterday

hash-d:
And @granzoto for review

ajssmith (approved):

PR #736, opened 2022-04-26
Fixed error when a HTTP2 connector/listener was synchronised

nluaces:
- Created ssl profile before the http2 connector/listener.
- Minor refactor: passing functions as parameters to facilitate mocking in tests.
- Changed message when the syncronisation was not finished without errors.

### DIFF
--- a/cmd/config-sync/config_sync.go
+++ b/cmd/config-sync/config_sync.go
@@ -261,6 +261,8 @@ func syncRouterConfig(agent *qdr.Agent, desired *qdr.RouterConfig, c *ConfigSync
 
 func syncSecrets(routerConfig *qdr.RouterConfig, changes *qdr.BridgeConfigDifference, sharedPath string, copyCerts CopyCerts, newSSlProfile CreateSSlProfile, delSslProfile DeleteSslProfile) error {
 
+	log.Printf("Sync profiles: Added %v  Deleted %v", changes.AddedSslProfiles, changes.DeletedSSlProfiles)
+
 	for _, addedProfile := range changes.AddedSslProfiles {
 		if len(addedProfile) > 0 {
 			log.Printf("Copying cert files related to sslProfile %s", addedProfile)
--- a/pkg/qdr/qdr.go
+++ b/pkg/qdr/qdr.go
@@ -448,6 +448,7 @@ type HttpEndpoint struct {
 	EventChannel    bool   `json:"eventChannel,omitempty"`
 	HostOverride    string `json:"hostOverride,omitempty"`
 	SslProfile      string `json:"sslProfile,omitempty"`
+	VerifyHostname  *bool  `json:"verifyHostname,omitempty"`
 }
 
 func convert(from interface{}, to interface{}) error {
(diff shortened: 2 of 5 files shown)


pkg/qdr/qdr.go
```diff
@@ -741,8 +741,9 @@ type TcpEndpointDifference struct {
 }
 
 type HttpEndpointDifference struct {
-	Deleted []HttpEndpoint
-	Added   []HttpEndpoint
+	Deleted          []HttpEndpoint
+	Added            []HttpEndpoint
+	AddedSslProfiles []string
```
grs, line 746: Would this field make more sense on the BridgeConfigDifference?

nluaces: Yes indeed, sslProfiles are independent entities. I will change that. Thanks!

cmd/config-sync/config_sync.go
```diff
@@ -249,23 +255,36 @@ func syncRouterConfig(agent *qdr.Agent, desired *qdr.RouterConfig, c *ConfigSync
 	}
 }
 
-func syncSecrets(configSync *ConfigSync, changes *qdr.BridgeConfigDifference, sharedTlsFilesDir string) error {
-	for _, added := range changes.HttpListeners.Added {
-		if len(added.SslProfile) > 0 {
-			log.Printf("Copying cert files related to HTTP Listener sslProfile %s", added.SslProfile)
-			err := configSync.copyCertsFilesToPath(sharedTlsFilesDir, added.SslProfile, added.SslProfile)
+func syncSecrets(routerConfig *qdr.RouterConfig, changes *qdr.BridgeConfigDifference, sharedPath string, copyCerts func(string, string, string) error, newSSlProfile func(profile qdr.SslProfile) error, delSslProfile func(string, string) error) error {
```
fgiorgetti, line 258: Just a cosmetic thing (so feel free to ignore it), but maybe the last 3
arguments could be represented by a type, or eventually, by an interface.

```
type CopyCerts func(string, string, string) error
type NewSSlProfile func(profile qdr.SslProfile) error
type DelSslProfile func(string, string) error
```

or

```
type SslProfileManager interface {
	Create(profile qdr.SslProfile) error
	Delete(name string) error
	Copy(path string, profilename string, secretname string) error
}
```

nluaces: I like it! it is way better. Thanks!

pkg/qdr/qdr.go
```diff
@@ -750,6 +750,7 @@ type BridgeConfigDifference struct {
 	TcpConnectors  TcpEndpointDifference
 	HttpListeners  HttpEndpointDifference
 	HttpConnectors HttpEndpointDifference
+	SslProfiles    []string
```
grs, line 753: Do we need to track deleted ssl profiles as well? I.e. what happens if a tls enabled service is unexposed then re-exposed?

pkg/qdr/qdr.go
```diff
@@ -836,6 +837,19 @@ func (a *BridgeConfig) Difference(b *BridgeConfig) *BridgeConfigDifference {
 		HttpConnectors: a.HttpConnectors.Difference(b.HttpConnectors),
 		HttpListeners:  a.HttpListeners.Difference(b.HttpListeners),
 	}
+
+	for _, httpConnector := range result.HttpConnectors.Added {
+		if len(httpConnector.SslProfile) > 0 {
+			result.SslProfiles = append(result.SslProfiles, httpConnector.SslProfile)
```
grs, line 843: Is it possible that more than one connector could ever share the same ssl profile? (If so, profiles are only really added if nothing else is using them).

pkg/qdr/qdr.go
```diff
@@ -836,9 +838,68 @@ func (a *BridgeConfig) Difference(b *BridgeConfig) *BridgeConfigDifference {
...
+	var addedSslProfiles []string
+
+	for _, httpConnector := range difference.HttpConnectors.Added {
+		if len(httpConnector.SslProfile) > 0 {
+			addedSslProfiles = append(addedSslProfiles, httpConnector.SslProfile)
+		}
+	}
+
+	for _, httpListener := range difference.HttpListeners.Added {
+		if len(httpListener.SslProfile) > 0 {
```
fgiorgetti, line 859: Is it possible that the httpListener or httpConnector will share the same sslProfile?
If so, don't you need to validate if the sslProfile is already in the slice before adding?
Or each one uses its own sslProfile?

pkg/qdr/qdr.go
```diff
@@ -836,9 +838,68 @@ func (a *BridgeConfig) Difference(b *BridgeConfig) *BridgeConfigDifference {
 		HttpConnectors: a.HttpConnectors.Difference(b.HttpConnectors),
 		HttpListeners:  a.HttpListeners.Difference(b.HttpListeners),
 	}
+
+	result.AddedSslProfiles = checkAddedSslProfiles(result)
```
grs, line 842: I wonder if it would result in simpler, more intuitive code, if we just go through all endpoints for original config and create a map of the ssl profile names, then repeat that for the new config, creating a separate map, then compare the two maps to get the added and deleted ssl profile names?

nluaces: Thanks for this suggestion, I think that it is more clear now; let me know if the solution works for you.

pkg/qdr/qdr.go
```diff
@@ -448,6 +448,7 @@ type HttpEndpoint struct {
 	EventChannel    bool   `json:"eventChannel,omitempty"`
 	HostOverride    string `json:"hostOverride,omitempty"`
 	SslProfile      string `json:"sslProfile,omitempty"`
+	VerifyHostname  *bool  `json:"verifyHostname,omitempty"`
```
nluaces, line 451: @fgiorgetti I have added the `verifyHostname` field as suggested. Thanks!

cc:/@grs @ajssmith

grs (approved):

fgiorgetti (approved):

PR #740, opened 2022-04-27
Integration tests improvements

fgiorgetti:


granzoto (approved):
LGTM

fgiorgetti:
I will make an extra change here and put router to run using `gdb` and to use `trace+` logging level.

jiridanek:
I wonder how much trace+ logging will you get in a test that sends 1gb of to data from iperf3 through skupper.

fgiorgetti:
> I wonder how much trace+ logging will you get in a test that sends 1gb of to data from iperf3 through skupper.

Uncompressed it has 340MB, but after compression it goes down to 23MB (average). The "main" problem is,
that the pod log gets rotated when it reaches 100MB, so for this particular performance scenarios I will need to
use a different strategy.
I will start a log watcher at the moment the router pod goes to running state and start "tailing" the log into a buffer.
Once job is done, if it passed, I will discard the buffer, and if it failed I will save it separetely into file named
`skupper-router.full.log`.

I have already tested it locally and it seems to do the job, so I will update this PR.

fgiorgetti:
FTR: CI time to run integration tests has bumped from 40m to 50m using gdb+trace level.

PR #741, opened 2022-04-27
check for skrouterd before creating resources

ajssmith:


client/gateway.go
```diff
@@ -880,8 +875,16 @@ func (a *GatewayConfig) getBridgeConfig(gatewayName string, routerId string) (*q
 
 func (cli *VanClient) newGateway(ctx context.Context, gatewayName string, gatewayType string, configFile string, owner *metav1.OwnerReference) (string, error) {
 	var err error
+	var routerPath string = ""
 	gatewayResourceName := clusterGatewayName(gatewayName)
 
+	if gatewayType == GatewayServiceType {
+		routerPath, err = exec.LookPath("skrouterd")
+		if err != nil {
+			return "", fmt.Errorf("skrouterd not found, please 'dnf install skupper-router' first or use --type [docker|podman]")
```
fgiorgetti, line 884: Maybe we could provide a URL with skupper-router installation instructions (instead of using dnf only).
It would be even better, if the URL could be customized at build time.

grs (approved):

client/gateway.go
ajssmith: I will create a tracker for this as where skupper-router will be installed "from" is still tbd.

fgiorgetti (approved):

PR #742, opened 2022-04-27
Change `valgrind` router DebugMode to `asan` debug mode

jiridanek:
Second half of the https://github.com/skupperproject/skupper-router/pull/372 change. In the near future, there will be also a `tsan` DebugMode added, which will run the sanitizer equivalent to helgrind.

### DIFF
--- a/cmd/site-controller/README.md
+++ b/cmd/site-controller/README.md
@@ -37,7 +37,7 @@ ConfigMaps allow you to manage a Skupper site using a ConfigMap named `skupper-s
 
 `data:router-console` - (true/false) Set up a Dispatch Router console (not recommended).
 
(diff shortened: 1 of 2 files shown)


client/router_create.go
```diff
@@ -1161,6 +1161,9 @@ sasldb_path: /tmp/skrouterd.sasldb
 			if err != nil && !errors.IsAlreadyExists(err) {
 				return err
 			}
+			if err != nil {
+				fmt.Printf("aaa\n")
```
grs, line 1165: Is this intended?

jiridanek: 🤦🏽 Do you need to ask? I thought I forcepushed it away, but apparently I did not.

grs (approved):

PR #747, opened 2022-04-29
Policy controller and API to share same policies

### LINKED ISSUE #728: Service creation may fail silently when executed right after policy change that allows it.
# How to reproduce

On a script, pipeline or program (not manually, so that the steps occur as close one to the other as possible):
- Give `AllowedServices` permission on a namespace for any services (`"*"`)
- Immediatelly after that, issue `skupper service create hello-world-frontend 8080 --mapping http`
- Check that the service was created (`skupper service status`)

Run that combination several times.  In-between each cycle, remove the policy that allows the services and ensure that the services were removed.

# Expected behavior

The `skupper service create` command should either fail because of missing policies (and then the `skupper service status` command return empty), or succeed in such a way that `skupper service status` would return the created service.

# Actual behavior

The actual test was exactly as described above, but on two namespaces at the same time; in one of them, the policy was actually removed, while on the other the target namespace was removed  from the policy and later reinstated (the policy was edited).

The `skupper service create` command returned a successul response 100% of the time.  However, it not always created the services.  On each test, it failed at a different iteration:

- Test 1 - 1st
- Test 2 - 6st
- Test 3 - 10th
- Test 4 - 4th
- Test 5 - 12th

The low numbers on repetitions till failure indicates that it this situation should occur fairly frequently (whenever policy creation and service creation happen one right after the other).  Actually, the test was created because I saw it several times a day while performing other, unrelated testing.

fgiorgetti:
Fixes #728.

### DIFF
--- a/cmd/service-controller/policy_controller.go
+++ b/cmd/service-controller/policy_controller.go
@@ -5,7 +5,6 @@ import (
 	"log"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/skupperproject/skupper/api/types"
@@ -27,28 +26,21 @@ import (
 
 var (
 	staticPolicyWatchers []*client.ClusterPolicyValidator
-	policyWatcherLock    sync.Mutex
 )
 
 // AddStaticPolicyWatcher all watchers must be defined before
 // the PolicyController is started
 func AddStaticPolicyWatcher(pv *client.ClusterPolicyValidator) {
-	policyWatcherLock.Lock()
-	defer policyWatcherLock.Unlock()
-	if staticPolicyWatchers == nil {
-		staticPolicyWatchers = []*client.ClusterPolicyValidator{}
-	}
 	staticPolicyWatchers = append(staticPolicyWatchers, pv)
 }
 
 type PolicyController struct {
-	name             string
-	cli              *client.VanClient
-	validator        *client.ClusterPolicyValidator
-	informer         cache.SharedIndexInformer
-	queue            workqueue.RateLimitingInterface
-	activeMap        map[string]time.Time
-	staticPolicyLock sync.Mutex
+	name      string
+	cli       *client.VanClient
+	validator *client.ClusterPolicyValidator
+	informer  cache.SharedIndexInformer
+	queue     workqueue.RateLimitingInterface
+	activeMap map[string]time.Time
 }
 
 func (c *PolicyController) loadActiveMap() {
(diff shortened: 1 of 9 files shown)


cmd/service-controller/policy_controller.go
```diff
@@ -24,13 +25,30 @@ import (
...
-	informer  cache.SharedIndexInformer
-	queue     workqueue.RateLimitingInterface
-	activeMap map[string]time.Time
+	name             string
+	cli              *client.VanClient
+	validator        *client.ClusterPolicyValidator
+	informer         cache.SharedIndexInformer
+	queue            workqueue.RateLimitingInterface
+	activeMap        map[string]time.Time
+	staticPolicyLock sync.Mutex
```
grs, line 51: What are the respective roles of this lock and the policyWatcherLock above?

cmd/service-controller/policy_controller.go
```diff
@@ -24,13 +25,30 @@ import (
...
+	policyWatcherLock    sync.Mutex
+)
+
+// AddStaticPolicyWatcher all watchers must be defined before
+// the PolicyController is started
+func AddStaticPolicyWatcher(pv *client.ClusterPolicyValidator) {
+	policyWatcherLock.Lock()
+	defer policyWatcherLock.Unlock()
+	if staticPolicyWatchers == nil {
+		staticPolicyWatchers = []*client.ClusterPolicyValidator{}
```
grs, line 39: I don't believe we need to explicitly create the list here? If there are any elements to be added, append will create if necessary. The only place it seems to be used is in a range operation, which again will be ok on a nil slice.

grs (commented):
I don't properly understand the problem being resolved here, or the design of the locking.

cmd/service-controller/policy_controller.go
fgiorgetti: Actually I need to get rid of them. They were used in an initial approach I used.
The approach I ended up with no longer needs them, I will remove. Thanks for pointing it out.

cmd/service-controller/policy_controller.go
fgiorgetti: True stuff.

fgiorgetti:
> I don't properly understand the problem being resolved here, or the design of the locking.

Basically there is an eventual race between the controller components (definition monitor, policy API, policy controller, ...)
and the CLI. As they were loading the CRs individually, they might be looking at different states.
With this change, once the policy controller detects a policy change, it will update the list used by all instances. Basically
the CLI can still see an "older" state (same as the API has), but not a newer one.

fgiorgetti:
... and actually I will write down another enhancement soon to have all internal controller components calling
the API (via HTTP) directly, to get rid of this static list.

PR #750, opened 2022-05-03
[ci-trigger] Run integration tests with asan instead of gdb

jiridanek:


jiridanek:
it passed, no issues were revealed by doing this

PR #751, opened 2022-05-03
[ci-trigger] Run integration tests with tsan instead of gdb

jiridanek:


jiridanek:
I'll rebase and reopen this when is time to add the "tsan" debug option. That will need some more work, maybe create an array of the allowed values in one place.

PR #755, opened 2022-05-09
Test: An http2 service with TLS feature enabled

nluaces:
- Deploy the http2 service without cert files.
- Expose service with htt2 mapping and enabling TLS feature.
- Append to the deployment the cert files created in previous step.
- Call the service using the certification authority which was used to create service's certification files.

### DIFF
--- a/test/utils/k8s/job.go
+++ b/test/utils/k8s/job.go
@@ -18,7 +18,7 @@ import (
 func GetTestImage() string {
 	testImage := os.Getenv("TEST_IMAGE")
 	if testImage == "" {
-		testImage = "quay.io/nluaces/skupper-tests:latest"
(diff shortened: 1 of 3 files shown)


test/utils/k8s/job.go
```diff
@@ -18,7 +18,7 @@ import (
 func GetTestImage() string {
 	testImage := os.Getenv("TEST_IMAGE")
 	if testImage == "" {
-		testImage = "quay.io/skupper/skupper-tests:master"
+		testImage = "quay.io/skupper/skupper-tests:latest"
```
fgiorgetti, line 21: I believe it is still `master` here.

nluaces: True! Good catch!

granzoto (approved):
Tested. Worked fine. Tks

PR #758, opened 2022-05-10
TCP test setup to vary based on build tag

fgiorgetti:


fgiorgetti:
This PR allows the TCP performance test to run differently based on provided build tags.
When using `-tags=integration` (as done by CI), it will set debug options to the router (gdb, trace+).
When `-tags=performance` is provided, then router debug options won't be set.

test/integration/performance/tcp/tcp.go
```diff
@@ -0,0 +1,165 @@
...
+
+	ENV_SKUPPER_SITES              = "SKUPPER_SITES"
+	ENV_SKUPPER_MAX_FRAME_SIZE     = "SKUPPER_MAX_FRAME_SIZE"
+	ENV_SKUPPER_MAX_SESSION_FRAMES = "SKUPPER_MAX_SESSION_FRAMES"
+	ENV_SKUPPER_MEMORY             = "SKUPPER_MEMORY"
+	ENV_SKUPPER_CPU                = "SKUPPER_CPU"
+)
+
+func TestMain(m *testing.M) {
+	log.Println("BATATA")
```
granzoto, line 35: Banana works better here !!

fgiorgetti: I knew I should have used a fruit instead. :)

granzoto (approved):
LGTM, Nice change

grs (approved):

nluaces (approved):

Issue #768, opened 2022-05-17
TLS test fails when exposing a service with port 443 with a user that has no root permissions 


nluaces:
When the TLS test is executed with a user that has no root user permissions, the test fails because the nginx service can not run properly.

PR #769, opened 2022-05-17
 Change the exposed port to 8443 in the TLS test.

nluaces:
Solves #768


hash-d:
I don't see any further use of plain 443 on the http tests, so lgtm

hash-d (approved):

grs (approved):

PR #771, opened 2022-05-18
improve remove and status

ajssmith:
Handle exceptional conditions to improve removal and status

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -501,7 +501,7 @@ func NewCmdDelete(newClient cobraFunc) *cobra.Command {
 			silenceCobra(cmd)
 			gateways, err := cli.GatewayList(context.Background())
 			for _, gateway := range gateways {
-				cli.GatewayRemove(context.Background(), gateway.GatewayName)
+				cli.GatewayRemove(context.Background(), gateway.Name)
 			}
 			err = cli.SiteConfigRemove(context.Background())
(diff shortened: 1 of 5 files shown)


nluaces (approved):

grs (approved):

cmd/skupper/skupper.go
```diff
@@ -1482,24 +1490,32 @@ func NewCmdStatusGateway(newClient cobraFunc) *cobra.Command {
 			}
 
 			if len(gateways) == 0 {
-				fmt.Println("No gateway definitions found")
+				l := formatter.NewList()
+				l.Item("No gateway definition found on cluster")
+				gatewayType, err := client.GatewayDetectTypeIfPresent()
+				if err == nil {
+					if gatewayType != "" {
+						l.NewChild(fmt.Sprintf(" A gateway of type %s detected on local host. Run 'skupper gateway delete' to remove.", gatewayType))
```
hash-d, line 1498: I'm not sure the recommendation to run `gateway delete` is a good idea.  The empty list might be a transient error talking to the gateway pod, and then the user would be removing locally a gateway that is properly defined on the cluster.

ajssmith: I agree with your suggestion and will update. The need is for a debug gateway command to inspect and identify cluster/host discrepancies which should be the recommended action here.

PR #782, opened 2022-06-01
Handling controller restarts during policy validation

### LINKED ISSUE #753: PolicyAPIClient.IncomingLink() may incorrectly report Policy validation error if executed during service-controller restart
# Description

During a full integration test on a cluster with Policy CRD and an all-allowing policy (all fields with `'*'` or `true`), the HipsterShop test failed with the output below:

```
=== RUN   TestHipsterShop
--- FAIL: TestHipsterShop (53.83s)
    hipstershop.go:60: Creating VANs
    hipstershop.go:83: setting siteConfigSpec to run with Ingress=none
    hipstershop.go:105: assertion failed: error is not nil: Policy validation error: command terminated with exit code 137: unable to create token to public1 cluster
    hipstershop_test.go:46: entering teardown
```

Looking at the code, `PolicyAPIClient.IncomingLink()` calls `get` on the `service-controller` container.  Error `137` indicates a process that was killed with `SIGKILL`.

# How to reproduce

I'll try to create a reproducer that uses the actual code.  Getting the `137` return code should not be difficult.  However, reproducing the exact situation from the HipsterShop that causes the pod to restart and generate the `137` might be impossible.

Meanwhile, I have reproduced only the `137` response with the following process:

- Within the service-controller container, run an infinite loop calling `get policies -h`
  - (the loop must be within the container; do not put the `podman exec` on a loop, as the error will be different)
- Remove the pod with something like `oc delete pod skupper-service-controller-76d867f558-mm2vz`
- The loop will end with error 137

Sample output:

```
$ oc exec -ti skupper-service-controller-76d867f558-cbk6j -- bash
bash-4.4$ while true; do get policies -h > /dev/null; done 
command terminated with exit code 137
$ 
```

# Suggestions

* `PolicyAPIClient.execGet()` calls `get` twice: one with `-h`, and the other with the actual query.  The fix needs to be done on both places, as either process may be the one killed during restart
* The first call to `get` already handles return code 1.  I'd suggest that any return codes above 128 be also handled, with a message that the process was killed by a signal
* When that happens, the error should not be reported as a Policy error.  It should probably be reported as a connectivity error.

With that said, should HipsterShop have failed in the first place, because of a restarting `service-controller` pod?  Would it have, if policies were disabled?  Perhaps, besides fixing the message, a further check should be done on how activating the policies impacts the handling of restarting pods by pre-existing pieces of the code.

# Other notes

As a side note (as this is probably `assert.Assert`'s fault, the error `assertion failed: error is not nil: Policy validation error: command terminated with exit code 137: unable to create token to public1 cluster` looks weird, in that it indicates that `unable to create token to public1 cluster` would be wrapped within `command terminated with exit code 137`, which is not the case.

I'm not sure whether that can be fixed, but it surely confuses things.

fgiorgetti:
Fixes #753

### DIFF
--- a/client/policy_validator.go
+++ b/client/policy_validator.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -9,6 +10,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/skupperproject/skupper/api/types"
 	v1alpha12 "github.com/skupperproject/skupper/pkg/apis/skupper/v1alpha1"
 	"github.com/skupperproject/skupper/pkg/event"
 	"github.com/skupperproject/skupper/pkg/generated/client/clientset/versioned/typed/skupper/v1alpha1"
@@ -348,9 +350,22 @@ func (p *PolicyAPIClient) execGet(args ...string) (*PolicyAPIResult, error) {
 	fullArgs := []string{"get", "policies"}
 	fullArgs = append(fullArgs, args...)
 	fullArgs = append(fullArgs, "-o", "json")
-	out, err := p.cli.exec(fullArgs, p.cli.GetNamespace())
+	var out *bytes.Buffer
+	err = utils.RetryWithContext(ctx, time.Millisecond*100, func() (bool, error) {
+		out, err = p.cli.exec(fullArgs, p.cli.GetNamespace())
+		if err != nil {
+			if _, err := getRootObject(p.cli); err != nil && errors.IsNotFound(err) {
+				return true, notEnabledErr
+			}
+			return false, nil
+		}
+		return true, nil
+	})
 	if err != nil {
-		return nil, fmt.Errorf("Policy validation error: %v", err)
+		if os.IsTimeout(err) {
+			return nil, fmt.Errorf("Policy validation error: %s not ready", types.ControllerDeploymentName)
+		}
+		return nil, notEnabledErr
 	}
 	res := &PolicyAPIResult{}
 	err = json.Unmarshal(out.Bytes(), res)
--- a/client/policy_validator.go
+++ b/client/policy_validator.go
@@ -316,7 +316,8 @@ func (p *PolicyAPIClient) execGet(args ...string) (*PolicyAPIResult, error) {
 	ctx, cn := context.WithTimeout(context.Background(), time.Second*30)
 	defer cn()
 	notEnabledErr := fmt.Errorf("Skupper is not enabled in namespace '%s'", p.cli.Namespace)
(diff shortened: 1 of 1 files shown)


nluaces (approved):

grs (approved):

client/policy_validator.go
```diff
@@ -348,9 +350,22 @@ func (p *PolicyAPIClient) execGet(args ...string) (*PolicyAPIResult, error) {
 	fullArgs := []string{"get", "policies"}
 	fullArgs = append(fullArgs, args...)
 	fullArgs = append(fullArgs, "-o", "json")
-	out, err := p.cli.exec(fullArgs, p.cli.GetNamespace())
+	var out *bytes.Buffer
+	err = utils.RetryWithContext(ctx, time.Millisecond*100, func() (bool, error) {
```
hash-d, line 354: Is there a reason to retry every tenth of second?  I imagine people might put rate limiting on their clusters, and I guess this would generate some 20 requests per second.

client/policy_validator.go
```diff
@@ -348,9 +350,22 @@ func (p *PolicyAPIClient) execGet(args ...string) (*PolicyAPIResult, error) {
...
+				return true, notEnabledErr
+			}
+			return false, nil
+		}
+		return true, nil
+	})
 	if err != nil {
-		return nil, fmt.Errorf("Policy validation error: %v", err)
+		if os.IsTimeout(err) {
+			return nil, fmt.Errorf("Policy validation error: %s not ready", types.ControllerDeploymentName)
```
hash-d, line 366: Any chance capturing and showing the last error that ocurred, in addition to mentioning the timeout?  That might help in investigations

hash-d (approved):
I have added a couple of comments, but those are only suggestions.

That said, I have tested it, and it works fine.  I have merged your changes to my policy testing and added the specific test, and you can check it all here, in case you're interested:

https://github.com/hash-d/skupper/blob/dh-test-fg-issue-753/test/integration/acceptance/custom/hello_policy/issues_test.go#L21

I don't think that should break anything else, but I'll run a full regression on the policy tests and report.

hash-d:
Full policy regression passed

client/policy_validator.go
fgiorgetti: I think it can be increased to 1s.

client/policy_validator.go
fgiorgetti: I was thinking about it as well. I think that after 30s if it is still failing, it should be useful.
I will include it.

client/policy_validator.go
fgiorgetti: Actually 1s does not feel right, as it is being done twice here. So I will use 200ms which is a value that still feels
ok from a user experience standpoint.

PR #793, opened 2022-06-15
Simplifies incoming links state changes

### LINKED ISSUE #789: Policies: a pre-existing token may be able to activate a link while IncomingLink is not allowed
# Description

The error happens on the test `TestPolicies/tests/testLinkPolicy/previously-created-token`, but not always.  An example (with generated debug dumps) can be found on [the latest CI run for the policies testing PR](https://app.circleci.com/pipelines/github/skupperproject/skupper/1637/workflows/619525f3-96eb-438d-b052-7d388517cbd8/jobs/8155).

The test basically goes like this:

- Set `AllowIncomingLinks` as true on a namespace, create a token
- Set `AllowIncomingLinks` to false on the same namespace, and try to create a link using the pre-created token

The expectation is that:

1. The link will be created, but
2. The link will be left in inactive state

However, when the issue happens, `skupper link status` reports the link as `up`.

# How to reproduce

A specific test has been created on that PR:  `testLinkIssue789` on `test/integration/acceptance/custom/hello_policy/link_test.go`.  In order to run it, set the environment variable `SKUPPER_TEST_RUN_ISSUE_TESTS` before running `go test ./test/integration/acceptance/custom/hello_policy/ -tags policy -v -count 1 -timeout 720m -run //testLinkIssue789`.

This test is _generally_ able to reproduce the issue, but not always.  You may increase the number of times the loop runs (currently defaults to 30), or run it several times, until the issue is reproduced.

fgiorgetti:
Fixes #789

### DIFF
--- a/cmd/service-controller/policy_controller.go
+++ b/cmd/service-controller/policy_controller.go
@@ -193,11 +193,13 @@ func (c *PolicyController) validateIncomingLinkStateChanged() {
 		event.Recordf(c.name, "[validateIncomingLinkStateChanged] error validating policy: %v", res.Error())
 		return
 	}
-	c.adjustListenerState("validateIncomingLinkStateChanged", "interior-listener", res.Allowed(), client.InteriorListener)
-	c.adjustListenerState("validateIncomingLinkStateChanged", "edge-listener", res.Allowed(), client.EdgeListener)
-}
+	source := "validateIncomingLinkStateChanged"
+	allowed := res.Allowed()
+	listeners := map[string]func(options types.SiteConfigSpec) qdr.Listener{
+		"interior-listener": client.InteriorListener,
+		"edge-listener":     client.EdgeListener,
+	}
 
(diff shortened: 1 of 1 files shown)


nluaces (approved):

grs (approved):

hash-d:
Since this error is intermittent, I can't confirm this is fixed.  However, it was happening ~40% of the time on the Policies CI, and I have run it three times pointing to your image, without it occurring.  So I'm pretty confident (~80%) that this is now good.

I have also ran policy testing including issue testing locally, and it passed.

hash-d (approved):
Since this error is intermittent, I can't confirm this is fixed. However, it was happening ~40% of the time on the Policies CI, and I have run it three times pointing to your image, without it occurring. So I'm pretty confident (~80%) that this is now good.

I have also ran policy testing including issue testing locally, and it passed.

PR #794, opened 2022-06-15
Basic integration test fix

### LINKED ISSUE #765: Race condition on removal of skupper in TestBasic
#731 introduced the deferred function `Delete(ctx context.Context, t *testing.T)` on `test/integration/acceptance/custom/basic/basic.go`, which removes Skupper from the namespaces with two operations:

    assert.Assert(t, pub1Cluster.VanClient.SiteConfigRemove(ctx))
    assert.Assert(t, pub1Cluster.VanClient.RouterRemove(ctx))

That seems to be causing the following intermittent error on CI test runs:

    --- FAIL: TestBasic/interiors-ingress-nodeport (41.11s)
        basic.go:356: Testing: Connecting two interiors with ingress=nodeport
        basic.go:126: assertion failed: error is not nil: Skupper not installed in 'private-basic-1': deployments.apps "skupper-router" not found

The error has been seen on both policy and non-policy testing (ie, with and without the Policy CRD installed).

fgiorgetti:
Fixes #765

### DIFF
--- a/test/integration/acceptance/custom/basic/basic.go
+++ b/test/integration/acceptance/custom/basic/basic.go
@@ -121,10 +121,12 @@ func (r *BasicTestRunner) Delete(ctx context.Context, t *testing.T) {
 	pub1Cluster, _ := r.GetPublicContext(1)
 	prv1Cluster, _ := r.GetPrivateContext(1)
 	if err := pub1Cluster.VanClient.SiteConfigRemove(ctx); err != nil {
-		_ = pub1Cluster.VanClient.RouterRemove(ctx)
+		t.Logf("error removing site config: %v", err)
+		t.Logf("removing router - err: %v", pub1Cluster.VanClient.RouterRemove(ctx))
 	}
 	if err := prv1Cluster.VanClient.SiteConfigRemove(ctx); err != nil {
-		_ = prv1Cluster.VanClient.RouterRemove(ctx)
+		t.Logf("error removing site config: %v", err)
+		t.Logf("removing router - err: %v", prv1Cluster.VanClient.RouterRemove(ctx))
 	}
 	waitNoPods := func(componentSelector string, cluster *base.ClusterContext) error {
 		return utils.RetryWithContext(ctx, time.Second, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


test/integration/acceptance/custom/basic/basic.go
```diff
@@ -120,10 +120,12 @@ func (r *BasicTestRunner) Delete(ctx context.Context, t *testing.T) {
...
-	assert.Assert(t, pub1Cluster.VanClient.RouterRemove(ctx))
-	assert.Assert(t, prv1Cluster.VanClient.SiteConfigRemove(ctx))
-	assert.Assert(t, prv1Cluster.VanClient.RouterRemove(ctx))
+	if err := pub1Cluster.VanClient.SiteConfigRemove(ctx); err != nil {
+		_ = pub1Cluster.VanClient.RouterRemove(ctx)
+	}
+	if err := prv1Cluster.VanClient.SiteConfigRemove(ctx); err != nil {
+		_ = prv1Cluster.VanClient.RouterRemove(ctx)
+	}
 	waitNoPods := func(componentSelector string, cluster *base.ClusterContext) error {
```
hash-d, line 129: I don't see the reason for the new `if`, if the `Assert`s are gone.  This way, `pub1Cluster.VanClient.RouterRemove` will never be tried if `pub1Cluster.VanClient.SiteConfigRemove` fails.

Removing the `if`, in the case that `pub1Cluster.VanClient.SiteConfigRemove` fails, would you get a second chance of getting rid of the skupper installation with `pub1Cluster.VanClient.RouterRemove`?  Or would that not be sufficient, and that's the reason for the `if`?

It's good that the `Assert` statements are gone, so that an otherwise working test does not fail on a cleanup task.  However, I think we need some logs in case of errors there.  Otherwise, it may be difficult to understand why the next test fails when the removal does not succeed.

fgiorgetti: Actually if SiteConfigRemove fails, the RouterRemove will be performed, and only if it fails.
If the controller is running properly, just removing the site-config should be enough. That's 
exactly what skupper delete does in fact.

fgiorgetti: Ok, I will add it.

hash-d (approved):
This error occurred 3 times in the past 10 CI runs; I have re-ran it three times pointing to this branch, and the error did not occur.  As it is intermittent, I can't state it is fixed, but these runs give me a ~65% confidence it is.  That said, knowing the root cause for the error and reviewing the fix provided, I'm convinced it is fixed.

PR #804, opened 2022-06-28
Get logs for all job runs, instead of only the last one

hash-d:
Often, the last one will be still running when the test timeout
kicks in, and the logs will be empty or incomplete, where the
previous runs will have information on what cause them to fail.

Signed-off-by: Danilo Gonzalez Hashimoto <dhashimo@redhat.com>

### DIFF
--- a/test/utils/k8s/job.go
+++ b/test/utils/k8s/job.go
@@ -175,24 +175,49 @@ func AssertJob(t *testing.T, job *batchv1.Job) {
 	}
 }
 
-// TODO: only latest or all?
 func GetJobLogs(ns string, kubeClient kubernetes.Interface, name string) (string, error) {
+	return GetJobsLogs(ns, kubeClient, name, false)
+}
+
+// Returns the logs of the pods related to a given job.  If 'all' is true, all runs will be
+// returned; failures to get individual logs will be reported on the output, but not cause
+// the function to fail.
+//
+// If 'all' is false, only the last run's logs will be returned.
+func GetJobsLogs(ns string, kubeClient kubernetes.Interface, name string, all bool) (string, error) {
 	pods, err := kube.GetPods(fmt.Sprintf("job=%s", name), ns, kubeClient)
 	if err != nil {
 		return "", err
+	}
+	if len(pods) == 0 {
+		return "", fmt.Errorf("No pods found for job %s on namespace %s", name, ns)
 	}
 	sort.Slice(pods, func(i, j int) bool {
 		return pods[i].CreationTimestamp.Time.Before(pods[j].CreationTimestamp.Time)
 	})
 	var fullLogs string
-	for _, pod := range pods {
-		fullLogs += fmt.Sprintf("\n# %v - %v:\n", pod.Name, pod.CreationTimestamp)
-		log, err := kube.GetPodContainerLogs(pod.Name, pod.Spec.Containers[0].Name, ns, kubeClient)
+	if all {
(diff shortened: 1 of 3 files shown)


test/utils/k8s/job.go
```diff
@@ -175,22 +176,50 @@ func AssertJob(t *testing.T, job *batchv1.Job) {
...
 	pods, err := kube.GetPods(fmt.Sprintf("job=%s", name), ns, kubeClient)
 	if err != nil {
 		return "", err
 	}
-	// Consider just the latest pod
-	latestPod := pods[0]
-	for _, pod := range pods {
-		if pod.CreationTimestamp.After(latestPod.CreationTimestamp.Time) {
-			latestPod = pod
-		}
```
fgiorgetti, line 187: This must be preserved. The pods returned by the `List` method are not sorted by creation timestamp.

test/utils/k8s/job.go
```diff
@@ -175,22 +176,50 @@ func AssertJob(t *testing.T, job *batchv1.Job) {
...
+					ns,
+					err,
+				)
+			}
+			fullLogs += log
+			fullLogs += "\n"
+		}
+	} else {
+		// Just the last one
+		pod := pods[len(pods)-1]
```
fgiorgetti, line 216: It should use the latestPod here instead.

test/utils/k8s/job.go
```diff
@@ -175,22 +176,50 @@ func AssertJob(t *testing.T, job *batchv1.Job) {
...
 func GetJobLogs(ns string, kubeClient kubernetes.Interface, name string) (string, error) {
+	return GetJobsLogs(ns, kubeClient, name, false)
+}
+
+// Returns the logs of the pods related to a given job.  If 'all' is true, all runs will be
+// returned; failures to get individual logs will be reported on the output, but not cause
+// the function to fail.
+//
+// If 'all' is false, only the last run's logs will be returned.
+func GetJobsLogs(ns string, kubeClient kubernetes.Interface, name string, all bool) (string, error) {
```
fgiorgetti, line 187: It might be better to return a map indexed by pod name with value being the logs,
in the former `GetJobLogs` you keep it as is and just consume the first element of
the returned map.

I believe it might work for the long term as future jobs can, eventually, save each log
to an individual file.

test/utils/k8s/job.go
hash-d: I'm sorting the list (see line 195):

	sort.Slice(pods, func(i, j int) bool {
		return pods[i].CreationTimestamp.Time.Before(pods[j].CreationTimestamp.Time)
	})

test/utils/k8s/job.go
hash-d: If map indexed by podname, we lose the timing information...  It could be a slice of structs that have the logs + meta info (timing, podname, etc), and already in the proper order, instead.

On your suggestion, I'm leaning toward creating a third function, then, that returns the map or slice.  GetJobLogs would do as  you say, and GetJobsLogs would call this new function and return the logs as a string, just like GetJobsLogs does today.  That would make that preparation, but still require little changes on the job itself.

On this PR, I just want to get proper information for my investigation; I wanted to make as little changes to the test as possible, before I have root cause for the issue I'm investigating.

fgiorgetti: Sounds fair.

fgiorgetti (approved):

granzoto (approved):
Nice change

Issue #811, opened 2022-07-05
Configure private registry for images


nluaces:
Feature request to configure skupper to access images from a private/disconnected image registry, assuming that the image names remain the same.

PR #812, opened 2022-07-05
Add new flag to specify an image registry

nluaces:
Solves #811


client/image_utils.go
```diff
@@ -95,3 +97,24 @@ func GetConfigSyncImageName() string {
 func GetConfigSyncImagePullPolicy() string {
 	return getPullPolicy(ConfigSyncPullPolicyEnvKey)
 }
+
+func SetImageRegistry(url *url.URL) error {
+
+	path := []string{url.String(), RouterImageName}
+	err := os.Setenv(RouterImageEnvKey, strings.Join(path, "/"))
```
grs, line 104: Instead of setting the individual image values if the registry is specified, what about changing the implementation of the various GetXXXImageName() methods to check if an env var setting the registry has been set (only if the full image name is not already specified) and use that if so?  

If wanted you can still keep the cli option, and it would just set that new env var for registry. (Personally I think I would perhaps let users set the registry env var themselves and not try to do it through the cli options).

fgiorgetti: Agreed. That would then simply change what the "default" value is, if the existing env vars are not set, right?

grs: Exactly.

cmd/skupper/skupper.go
```diff
@@ -420,6 +421,18 @@ installation that can then be connected to other skupper installations`,
 				}
 			}
 
+			if routerCreateOpts.ImageRegistry != "" {
+				urlImageRegistry, err := url.Parse(routerCreateOpts.ImageRegistry)
+				if err != nil {
+					return fmt.Errorf("The specified image registry is not valid: %s", err)
```
fgiorgetti, line 427: This validation does not work properly. If you have a registry running on a custom port,
you usually set it to, in example: `10.0.0.1:5000`. The url.Parse expects the protocol to be defined.

Maybe it can be added temporarily as a prefix but removed from the final value to be used.

fgiorgetti:
One general comment. Should it be added to the SiteConfig as well? I am wondering if someone sets it
during `skupper init`, but while performing a `skupper update` the registry information will be lost.

fgiorgetti:
> I agree, if we do add a flag to init, we should do the same for update.

Yes, I think it is better.

client/image_utils.go
nluaces: Agreed. @ajssmith what do you think about removing the cli option? Is it possible that users are already used to modify the current env vars? In that case, maybe new users would need some kind of awareness of these env vars through the `help` command?

cmd/skupper/skupper.go
nluaces: Thanks! I will try to parse the url using a regexp instead.

grs (approved):

fgiorgetti (approved):

grs (approved):
very nice!

fgiorgetti (approved):
LGTM.

PR #821, opened 2022-07-12
Properly restoring service ports and selector

### LINKED ISSUE #817: Skupper unexpose is not cleaning up correctly services
When you expose a service as a skupper service, it changes the selector annotations to match the router pod. However, when issuing the `unexpose` command, it does not remove them, making the service unusable until they are manually removed.

fgiorgetti:
Fixes #817


nluaces (approved):
Looks good to me, thanks!

PR #825, opened 2022-07-25
Fix redis not available for cartservice on HipsterShop test

hash-d:
Cartservice depends on Redis, via Skupper.  However, the test would only expose the Redis service after it detected that all deployments were up.  By chance, on most of the tests the check for deployments and the cartservice retries from CrashLoopBackOff would be in sync, and the test would pass.

However, that was dependent on a multitude of factors from the environment the test was running on, causing the test to have flaky results.

This PR moves the Skupper exposing of Redis to right after its deployment, allowing the cartservice to connect to Redis earlier, and avoiding the test failures.

### DIFF
--- a/test/integration/examples/custom/hipstershop/hipstershop.go
+++ b/test/integration/examples/custom/hipstershop/hipstershop.go
@@ -142,8 +142,13 @@ func DeployResources(t *testing.T, testRunner base.ClusterTestRunner) {
 
 	t.Logf("Deploying microservices on %s", prv1.Namespace)
 	assert.Assert(t, k8s.CreateResourcesFromYAML(prv1.VanClient, RESOURCES_PRV1))
+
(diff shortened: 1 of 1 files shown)


fgiorgetti:
@hash-d It looks good to me from an integration test point of view.
What do you think of proposing a similar change to the example itself?

fgiorgetti (approved):

ajssmith (approved):
lgtm, thanks

we can investigate whether adding an init container to card service in the example would help address the issue.

hash-d:
@fgiorgetti, I tried to make the change without affecting the examples, as much as possible.  My rationale is that the examples need to be simple, and can afford to wait for the pieces to converge into the end state.  Adding this change to the example would mean one additional step on it, making it (slightly?) more complex.

I guess @ajssmith's proposal would be able to avoid changes on the example's steps, which would be ideal.  Not sure what's the proper forum for that discussion, though.

PR #827, opened 2022-07-26
Improved codegen scripts to work under gopath

fgiorgetti:


scripts/update-codegen.sh
```diff
@@ -18,6 +18,21 @@ set -o errexit
...
+SCRIPT_ROOT=$(cd `dirname "${BASH_SOURCE[0]}"`/.. && pwd)
+GOPATH_NEW="${SCRIPT_ROOT}/_go_tmp"
+GOPATH="${GOPATH_NEW}"
+TMP_DEST="${1:-}"
+mkdir ${GOPATH_NEW}
+
+cleanup() {
+    if [[ -n "${GOPATH_NEW}" && -d "${GOPATH_NEW}" ]]; then
+        chmod -R 755 ${GOPATH_NEW}
+        rm -rf "${GOPATH_NEW}"
```
grs, line 31: Does this mean that the code generator will be downloaded every time you run the script? Is that what we want?

grs (commented):
Would using --output-base option to generate-groups.sh avoid the need for copying? And would that simplify these scripts?

fgiorgetti:
> Would using --output-base option to generate-groups.sh avoid the need for copying? And would that simplify these scripts?

Certainly. I missed this flag. Let me update it. Thanks!

scripts/update-codegen.sh
```diff
@@ -18,17 +18,27 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+SCRIPT_ROOT=$(cd `dirname "${BASH_SOURCE[0]}"`/.. && pwd)
+[[ $# -eq 1 ]] && VERIFY_ONLY=true || VERIFY_ONLY=false
+TMP_DEST="${1:-${SCRIPT_ROOT}/_tmp_dest}"
```
grs, line 23: What about having an optional argument indicating the output base, and using SCRIPT_ROOT or current directory if none is provided? That way you would not need to copy but could still, if wanted, generate into a temp location.

fgiorgetti: If the `--output-base` is set, then code will be generated under:
`<output-base-dir>/github.com/skupperproject/skupper/pkg/{apis,generated}`.

Therefore using `${SCRIPT_ROOT}` would cause it to create this entire tree under the project's root dir.

If we do not provide the `--output-base` at all, then it will perform the updates straight into:
`${GOPATH}/src/github.com/skupperproject/skupper/pkg/`, which would work for people working
under `$GOPATH/src`, but not for people working outside of it.

grs: What if we also use `--trim-path-prefix github.com/skupperproject/skupper/`?

fgiorgetti: Let me give it a try.

grs (approved):

nluaces (approved):

scripts/update-codegen.sh
fgiorgetti: Keeping as is because `--trim-path-prefix` is not available with v0.17.0.

PR #828, opened 2022-07-27
Switch command

fgiorgetti:
This PR introduces a new `switch` command that will be used to control the target
platform to be manipulated by the `skupper` CLI.

Along with the `switch` command, it also introduces a new environment variable
named `SKUPPER_PLATFORM` and a new CLI option `--platform` that can be
used to override the currently defined platform.

At this point, both the `switch` command and the `--platform` option have been
marked as hidden.

In order to support different platforms through the CLI, it also introduces the
`SkupperClient` interface, which must be implemented by each supported
platform so that the supported commands can be executed properly.

The interface is now used by the existing command functions. The reason for
this choice was that the existing functions can be reused as well as the common
command definitions, flags and eventual generic behavior can be preserved, while
platform specific flags and implementation details can be customized as needed.

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -389,10 +389,6 @@ installation that can then be connected to other skupper installations`,
 
 	cmd.Flags().IntVar(&routerCreateOpts.Routers, "routers", 0, "Number of router replicas to start")
 
-	cmd.Flags().IntVar(&routerCreateOpts.Router.MaxFrameSize, "xp-router-max-frame-size", types.RouterMaxFrameSizeDefault, "Set  max frame size on inter-router listeners/connectors")
-	cmd.Flags().IntVar(&routerCreateOpts.Router.MaxSessionFrames, "xp-router-max-session-frames", types.RouterMaxSessionFramesDefault, "Set  max session frames on inter-router listeners/connectors")
-	hideFlag(cmd, "xp-router-max-frame-size")
-	hideFlag(cmd, "xp-router-max-session-frames")
 	cmd.Flags().SortFlags = false
 
 	// platform specific flags
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -42,7 +42,6 @@ type SkupperSiteClient interface {
 	UpdateFlags(cmd *cobra.Command)
 	Version(cmd *cobra.Command, args []string) error
 	RevokeAccess(cmd *cobra.Command, args []string) error
-	SkupperClientCommon
 }
 
 type SkupperServiceClient interface {
@@ -56,7 +55,6 @@ type SkupperServiceClient interface {
 	ExposeArgs(cmd *cobra.Command, args []string) error
 	ExposeFlags(cmd *cobra.Command)
 	Unexpose(cmd *cobra.Command, args []string) error
-	SkupperClientCommon
 }
 
 type SkupperDebugClient interface {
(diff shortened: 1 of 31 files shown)


pkg/config/local.go
```diff
@@ -0,0 +1,90 @@
...
+		decoder := yaml.NewDecoder(bytes.NewReader(data))
+		if err = decoder.Decode(p); err != nil {
+			return fmt.Errorf("error decoding %s: %v", PlatformConfigFile, err)
+		}
+	}
+	return nil
+}
+
+var (
+	Platform string
```
grs, line 61: What's the significance of this variable? (Otherwise all looks good!)

fgiorgetti: It is bound by a CLI option introduced here named `--platform` (similar to `--kubeconfig` or `--context`).

fgiorgetti:
One important point to mention, specially for the `init` command, is that due to
the approach used here, where only the common init flags have been preserved in
the original command definition, but the `kubernetes` specific flags have been
moved into its concrete implementation, the order in which the fields are showing
up in the `skupper help init` has changed.

As an alternative we can put all init flags under the kubernetes implementation for
now so that the original display order is preserved.

grs (approved):
Looks good to me

nluaces:
Looks good to me, thanks!

fgiorgetti:
Splitted `SkupperClient` interface into smaller domain based subsets.

grs (approved):

fgiorgetti:
@nluaces do you mind verifying if deprecated changes still look good for you?

cmd/skupper/skupper.go
```diff
@@ -316,171 +357,47 @@ func asMap(entries []string) map[string]string {
...
-	cmd.Flags().StringVar(&routerCreateOpts.ConfigSync.Memory, "config-sync-memory", "", "Memory request for config-sync pods")
-	cmd.Flags().StringVar(&routerCreateOpts.ConfigSync.CpuLimit, "config-sync-cpu-limit", "", "CPU limit for config-sync pods")
-	cmd.Flags().StringVar(&routerCreateOpts.ConfigSync.MemoryLimit, "config-sync-memory-limit", "", "Memory limit for config-sync pods")
-
-	cmd.Flags().DurationVar(&LoadBalancerTimeout, "timeout", types.DefaultTimeoutDuration, "Configurable timeout for the ingress loadbalancer option.")
 
+	cmd.Flags().IntVar(&routerCreateOpts.Router.MaxFrameSize, "xp-router-max-frame-size", types.RouterMaxFrameSizeDefault, "Set  max frame size on inter-router listeners/connectors")
+	cmd.Flags().IntVar(&routerCreateOpts.Router.MaxSessionFrames, "xp-router-max-session-frames", types.RouterMaxSessionFramesDefault, "Set  max session frames on inter-router listeners/connectors")
+	hideFlag(cmd, "xp-router-max-frame-size")
+	hideFlag(cmd, "xp-router-max-session-frames")
```
nluaces, line 395: This flags were also removed in the PR #877 and default values where assigned in the router create options:
```
routerCreateOpts.Router.MaxFrameSize = types.RouterMaxFrameSizeDefault
routerCreateOpts.Router.MaxSessionFrames = types.RouterMaxSessionFramesDefault
```

fgiorgetti: True. I have already added the default values, but missed this. Thanks!

nluaces: Thanks! but do you still need to assign the default values, right? 
Like here: https://github.com/skupperproject/skupper/blob/e3aa6ea29c73251b826a6dba3cbb4011ef797507/cmd/skupper/skupper.go#L368

nluaces:
@fgiorgetti I only have one comment, but the rest of the changes look good to me! 👍🏻

Issue #832, opened 2022-08-01
Support publishNotReadyAddresses=True set on a Service


nluaces:
## context:

On the headless service spec, field publishNotReadyAddresses must be set to true to propagate SRV DNS records for its Pods for the purpose of peer discovery. 

> "Headless" (without a cluster IP) Services are also assigned a DNS A or AAAA record, depending on the IP family of the Service, for a name of the form my-svc.my-namespace.svc.cluster-domain.example. Unlike normal Services, this resolves to the set of IPs of the Pods selected by the Service. [1]

>Because A or AAAA records are not created for Pod names, hostname is required for the Pod's A or AAAA record to be created. A Pod with no hostname but with subdomain will only create the A or AAAA record for the headless Service (default-subdomain.my-namespace.svc.cluster-domain.example), pointing to the Pod's IP address. Also, Pod needs to become ready in order to have a record unless publishNotReadyAddresses=True is set on the Service. [1]

## approach:

* provide an option for a target to allow to create connections before it is ready.

[1]https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

PR #833, opened 2022-08-01
support service spec publishNotReadyAddress 

nluaces:
Solves #832

### DIFF
--- a/pkg/kube/podtargetresolver.go
+++ b/pkg/kube/podtargetresolver.go
@@ -58,8 +58,8 @@ func (o *PodTargetResolver) List() []string {
 
 	for _, p := range pods {
 		pod := p.(*corev1.Pod)
-		isPodEligible = o.skipStatusCheck || (IsPodRunning(pod) && IsPodReady(pod))
-		if isPodEligible && pod.DeletionTimestamp == nil {
+		isPodEligible = o.skipStatusCheck || IsPodReady(pod)
+		if isPodEligible && IsPodRunning(pod) && pod.DeletionTimestamp == nil {
 			event.Recordf(BridgeTargetEvent, "Adding pod for %s: %s", o.address, pod.ObjectMeta.Name)
 			targets = append(targets, pod.Status.PodIP)
 		} else {
--- a/pkg/kube/podtargetresolver.go
+++ b/pkg/kube/podtargetresolver.go
@@ -33,9 +33,10 @@ const (
 )
 
 type PodTargetResolver struct {
-	address  string
-	informer cache.SharedIndexInformer
-	stopper  chan struct{}
+	address         string
+	informer        cache.SharedIndexInformer
+	stopper         chan struct{}
+	skipStatusCheck bool
 }
 
 func (o *PodTargetResolver) Start() error {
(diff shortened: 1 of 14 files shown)


pkg/kube/misc.go
```diff
@@ -47,11 +47,17 @@ func GetServiceInterfaceTarget(targetType string, targetName string, deducePort
 			return nil, fmt.Errorf("Could not read deployment %s: %s", targetName, err)
 		}
 	} else if targetType == "statefulset" {
+		targetSvc, err := GetService(targetName, namespace, cli)
```
grs, line 50: Is this right? Should it not take the service name from the statefulset spec?

pkg/kube/misc.go
```diff
@@ -66,15 +72,19 @@ func GetServiceInterfaceTarget(targetType string, targetName string, deducePort
 	} else if targetType == "pods" {
 		return nil, fmt.Errorf("VAN service interfaces for pods not yet implemented")
 	} else if targetType == "service" {
+		targetSvc, err := GetServiceTarget(targetName, namespace, cli)
```
grs, line 75: I don't think we want to do this for a target of type service. This would now fail if the service doesn't exist yet, which it would not do before. It also requires read permission for the service if it is in another namespace or else it will fail (though that is true if deduceports is true already).

But I think to begine with, its not clear that there is any value in taking the value of PublishNotReadyAddresses from the service here.

grs (commented):
I think this may be solving the 'wrong' problem... I think the issue is that we want an option for a service target that will not require pods to be ready before we add a connector for it (i.e. a way to configure the middle test in https://github.com/skupperproject/skupper/blob/master/pkg/kube/podtargetresolver.go#L58 to be skipped)

pkg/kube/misc.go
nluaces: Indeed, thanks!

pkg/kube/misc.go
nluaces: You are right, the value PublishNotReadyAddresses is related to services with statefultsets, probably not useful in this case. Thanks!

nluaces:
@grs thank you for your help! I was wrong thinking that the main purpose was to support that value across the service sync...and what it is needed is to support that functionality within skupper by adding connectors in the router for services that are not ready yet. I will work on that. Thanks again!

nluaces:
-

pkg/kube/podtargetresolver.go
```diff
@@ -53,9 +54,12 @@ func (o *PodTargetResolver) Close() {
 func (o *PodTargetResolver) List() []string {
 	pods := o.informer.GetStore().List()
 	var targets []string
+	var isPodEligible bool
+
 	for _, p := range pods {
 		pod := p.(*corev1.Pod)
-		if IsPodRunning(pod) && IsPodReady(pod) && pod.DeletionTimestamp == nil {
+		isPodEligible = o.skipStatusCheck || (IsPodRunning(pod) && IsPodReady(pod))
```
grs, line 61: I *think* that if skipStatusCheck is true, we still want to ensure IsPodRunning, but don't care about IsPodReady?

fgiorgetti (approved):
Other than that comment raised by @grs, it looks good to me.
I have tested after rebasing on top of master and it worked ok.

grs (approved):
Nice work! :shipit:

PR #850, opened 2022-08-10
Updated go to 1.17

### LINKED ISSUE #420: Update to use go 1.16

fgiorgetti:
Fixes #420


grs (approved):

nluaces (approved):
lgtm, thanks!

Issue #849, opened 2022-08-09
Provide a HTTP API that revokes access to all Sites


michael-s-crawford:
Expose the existing `skupper revoke-access` CLI functionality via the Skupper Console (HTTP) APIs, for example:

`curl --location --request DELETE 'https://skupper-host/revokeaccess'`

PR #851, opened 2022-08-10
Adding HTTP API for revoke-access

michael-s-crawford:
Resolves https://github.com/skupperproject/skupper/issues/849

`revoke-access` is only (currently) available via the Skupper CLI. This change will make it available via the HTTP API also, for example, in response to the following request:

`curl --location --request POST 'https://skupper-host/revokeaccess'`

### DATA response before hitting revokeaccess endpoint
image

### revokeaccess request
image

### DATA response after hitting revokeaccess endpoint
image

### DIFF
--- a/cmd/service-controller/sites.go
+++ b/cmd/service-controller/sites.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/skupperproject/skupper/client"
+)
+
+const (
+	SiteManagement string = "SiteManagement"
+)
+
+type SiteManager struct {
+	cli *client.VanClient
+}
+
+func newSiteManager(cli *client.VanClient) *SiteManager {
+	return &SiteManager{
+		cli: cli,
+	}
+}
+
+func (m *SiteManager) revokeAccess() error {
+	err := m.cli.RevokeAccess(context.Background())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func serveSites(m *SiteManager) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodDelete {
+			err := m.revokeAccess()
+			if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/service-controller/sites.go
```diff
@@ -0,0 +1,43 @@
...
+	err := m.cli.RevokeAccess(context.Background())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func serveSites(m *SiteManager) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodDelete {
```
grs, line 34: I would prefer POST as the method for triggering this action (as to me it is not really deleting an exposed resource).

cmd/service-controller/sites.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/skupperproject/skupper/client"
+)
+
+const (
+	SiteManagement string = "SiteManagement"
+)
+
+type SiteManager struct {
```
grs, line 14: As we already have some site related data served up, I would prefer to call this something else. Ideally more explicitly tied to the revoking of access, e.g. name the file revokeaccess.go, call this type AccessRevoker or something along those lines.

grs (changes requested):
Thank you for this PR! I've made a couple of inline comments.

michael-s-crawford:
> Thank you for this PR! I've made a couple of inline comments.

Thanks @grs I've updated accordingly. Note I called the file `revoke_access.go` to match other files in the directory

Rebuilt the Docker image and tested on my environment

cmd/service-controller/sites.go
michael-s-crawford: Changed to `POST`

cmd/service-controller/sites.go
michael-s-crawford: Done

grs (approved):

fgiorgetti (approved):
LGTM! Thanks for the PR.

PR #855, opened 2022-08-16
Kube resources iface

fgiorgetti:


api/types/client.go
```diff
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"time"
 
+	v12 "k8s.io/api/apps/v1"
```
grs, line 9: why v12 rather than appsv1?

api/types/client.go
```diff
@@ -319,4 +321,54 @@ type VanClientInterface interface {
...
 	NetworkStatus() ([]*SiteInfo, error)
+	ServiceManager(namespace string) Services
+	ConfigMapManager(namespace string) ConfigMaps
+	DeploymentManager(namespace string) Deployments
+	SecretManager(namespace string) Secrets
+	StatefulSetManager(namespace string) StatefulSets
+}
+
+type Services interface {
+	GetService(name string, options *v1.GetOptions) (*corev1.Service, bool, error)
```
grs, line 332: I have been trying to reduce dependencies on the client package, particularly from within the pkg tree. I would personally prefer to see these new interfaces introduced elsewhere.

I have also been trying to contain dependence on the kubernetes API a little better. This PR takes things in a different direction, namely making kubernetes resource a central aspect of the entire project. That is not necessarily wrong, but it does give me some concern. Should StatefulSets really be so central a concept?

pkg/kube/deployments.go
```diff
@@ -45,17 +45,17 @@ func GetDeploymentOwnerReference(dep *appsv1.Deployment) metav1.OwnerReference {
...
-		err = cli.AppsV1().Deployments(namespace).Delete(name, &metav1.DeleteOptions{})
+		err = cli.DeleteDeployment(dep, &metav1.DeleteOptions{})
 	}
 	return err
 }
 
 // TODO, pass full client object with namespace and clientset
-func GetDeployment(name string, namespace string, cli kubernetes.Interface) (*appsv1.Deployment, error) {
-	existing, err := cli.AppsV1().Deployments(namespace).Get(name, metav1.GetOptions{})
+func GetDeployment(name string, cli types.Deployments) (*appsv1.Deployment, error) {
```
grs, line 57: These changes - while they may be justified - seem to me to essentially invert the role of the kube package. It is no longer code tied to kubernetes, but potentially a more generic layer above the interface behind which an implementation using the kubernetes API server would sit. Is this intended? What do we see as the role going forward for the kube package?

api/types/client.go
```diff
@@ -319,4 +321,54 @@ type VanClientInterface interface {
...
+	DeleteService(svc *corev1.Service, options *v1.DeleteOptions) error
```
grs, line 333: Is there a specific need to expose DeleteOptions, GetOptions? Likewise for ListOptions, though there we clearly may need to request different types of list, is that the right level of abstraction?

api/types/client.go
fgiorgetti: That was my IDE, I have just switched to appsv1 as it really looks way better.

api/types/client.go
fgiorgetti: > I have been trying to reduce dependencies on the client package, particularly from within the pkg tree. I would personally prefer to see these new interfaces introduced elsewhere.

I agree that `pkg` tree should not depend on the client package and I have just moved the
Kubernetes implementation of the new interfaces into `pkg/kube`.

> I have also been trying to contain dependence on the kubernetes API a little better. This PR takes things in a different direction, namely making kubernetes resource a central aspect of the entire project. That is not necessarily wrong, but it does give me some concern.

The idea is to use the kubernetes models as an internal way to represent Skupper resources. I made it this way, because almost everything is tied to ConfigMaps, Secrets and Deployments (mainly). It sounded to me like a cleaner/safer approach to promote reuse of existing logic.

As an example, decoupling the existing `VanClientInterface` implementation from the kubernetes models, sounds to me like a bigger risk than simply providing appropriate mechanisms to store and manipulate the kubernetes resources that are already broadly manipulated by CLI and by the Controller.

> Should StatefulSets really be so central a concept?

I have added it mainly because of the existing logic that deals with headless services, but actually I am not sure it will be needed in different platforms.

Same thing for DaemonSets. I have not added generic CRUD interface to manipulate DaemonSets.
Do you think it might be needed?

pkg/kube/deployments.go
fgiorgetti: The way I see it is that the `kube` package is still responsible for providing mechanisms to manipulate
common kubernetes resources, but it delegates the CRUD operations to a concrete implementation which
might not necessarily perform that against a Kubernetes cluster.

api/types/client.go
fgiorgetti: @grs I got rid of GetOptions and DeleteOptions already. Now I am going to improve on the ListOptions.

pkg/kube/storage.go
```diff
@@ -0,0 +1,321 @@
...
+	GetDeployment(name string) (*appsv1.Deployment, bool, error)
+	DeleteDeployment(dep string) error
+	ListDeployments(options *types.ListFilter) ([]appsv1.Deployment, error)
+	CreateDeployment(dep *appsv1.Deployment) (*appsv1.Deployment, error)
+	UpdateDeployment(dep *appsv1.Deployment) (*appsv1.Deployment, error)
+	IsOwnedDeployment(dep *appsv1.Deployment) bool
+}
+
+type Secrets interface {
+	GetSecret(name string) (*corev1.Secret, bool, error)
```
nluaces, line 47: I'm not suggesting to change anything, I'm just curious about the following points:
- does the boolean show more information that can not be deducted by the existence of an error? In case of an absence, we still have an error message
- in go we have something equivalent to an `Either<Error, Object>` as the result value of the functions. I wonder if following a structure like `Option<Object>` (a value or an empty value in case of absence) for get functions would be useful.

fgiorgetti: That is a good question. It was based on an existing Services interface used by the controller.

I kept the boolean return parameter to allow the implementation to evaluate the error and eventually
return the error as nil on not founds, to avoid the caller from parsing it itself, as a "not found" error could
be represented differently on each implementation.

But since the subject itself (Secret) would be nil in such case, in spite of an error being returned, I believe it ended
up being just a convenience approach to check whether or not it was found without testing if the subject is not nil.

I think it is simpler to just remove the bool return parameter. Wdyt @nluaces and @grs?

nluaces: I don't have a strong opinion about this; I think we could keep the boolean parameter if it's used in several places.

nluaces (approved):
Big change!
it has improved the readability, thanks!

I also think that this is a good input for an ADR. Like a brief advice for future collaborators to use the these new interfaces instead of the KubeClient library directly. What do you think?

fgiorgetti:
> Big change! it has improved the readability, thanks!
> 
> I also think that this is a good input for an ADR. Like a brief advice for future collaborators to use the these new interfaces instead of the KubeClient library directly. What do you think?

If it gets approved/merged, I think it is a great idea to document and explain that those k8s resources should be managed
through the Interfaces and not directly using k8s api.

fgiorgetti:
Closing in favor of an upcoming approach focused on Skupper entities

PR #859, opened 2022-08-23
Mongo replicaset fix

hash-d:
Mongo tests were failing because replicaset leader election was not complete by the time the test runs.

The best practice is to ensure that the number of voting members on a replicaset is odd; we had two members.

- Add third mongodb replica to the replicaset
- Give each member a different priority on elections
- Wait for elections to be settled before starting actual tests
- Add logging to help relating failures to items in tcpdump, and to inspect Mongo status

The new member on the replicaset mirrors what's done on https://github.com/skupperproject/skupper-example-mongodb-replica-set (and it actually uses a YAML from that repo).


fgiorgetti (approved):
LGTM :shipit:

PR #860, opened 2022-08-23
add new flag to make the timeout for the loadbalancer IP configurable

### LINKED ISSUE #807: timeout for loadbalancer resolution should be configurable
Invoking `skupper init` with the default loadbalancer option waits until the loadbalancer ip or host is available. At present it gives up after a hardcoded duration. We should make that waiting time configurable.

nluaces:
Fixes #807

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -270,7 +270,7 @@ type TargetInfo struct {
 }
 
 type VanClientInterface interface {
-	RouterCreate(ctx context.Context, options SiteConfig, loadbalancertimeout int) error
+	RouterCreate(ctx context.Context, options SiteConfig) error
 	RouterInspect(ctx context.Context) (*RouterInspectResponse, error)
 	RouterInspectNamespace(ctx context.Context, namespace string) (*RouterInspectResponse, error)
 	RouterRemove(ctx context.Context) error
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -96,7 +96,6 @@ type SiteConfigSpec struct {
 	Router              RouterOptions
 	Controller          ControllerOptions
 	ConfigSync          ConfigSyncOptions
-	LoadBalancerTimeout int
 }
 
 const (
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -96,6 +96,7 @@ type SiteConfigSpec struct {
 	Router              RouterOptions
 	Controller          ControllerOptions
 	ConfigSync          ConfigSyncOptions
+	LoadBalancerTimeout int
 }
 
 const (
@@ -269,7 +270,7 @@ type TargetInfo struct {
 }
 
 type VanClientInterface interface {
-	RouterCreate(ctx context.Context, options SiteConfig) error
+	RouterCreate(ctx context.Context, options SiteConfig, loadbalancertimeout int) error
 	RouterInspect(ctx context.Context) (*RouterInspectResponse, error)
 	RouterInspectNamespace(ctx context.Context, namespace string) (*RouterInspectResponse, error)
 	RouterRemove(ctx context.Context) error
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -1062,6 +1062,9 @@ sasldb_path: /tmp/skrouterd.sasldb
 	initialConfig := qdr.AsConfigMapData(van.RouterConfig)
 	kube.NewConfigMap(types.TransportConfigMapName, &initialConfig, nil, nil, siteOwnerRef, van.Namespace, cli.KubeClient)
 
+	currentContext := getCurrentContextOrDefault(ctx)
+	deadline, _ := currentContext.Deadline()
+
 	if options.Spec.RouterMode == string(types.TransportModeInterior) {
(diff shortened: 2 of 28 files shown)


api/types/client.go
```diff
@@ -269,7 +270,7 @@ type TargetInfo struct {
 }
 
 type VanClientInterface interface {
-	RouterCreate(ctx context.Context, options SiteConfig) error
+	RouterCreate(ctx context.Context, options SiteConfig, loadbalancertimeout int) error
```
fgiorgetti, line 273: Maybe you could just set the timeout value at `SiteConfig.Spec.LoadBalancerTimeout`
during skupper init or during site-controller initialization and avoid changing this signature.

client/router_create.go
```diff
@@ -1189,7 +1195,11 @@ sasldb_path: /tmp/skrouterd.sasldb
 					log.Printf("Failed to retrieve route %q: %s", types.ClaimRedemptionRouteName, err.Error())
 				}
 			} else if options.Spec.IsIngressLoadBalancer() {
-				err = cli.appendLoadBalancerHostOrIp(types.ControllerServiceName, van.Namespace, &cred)
+				var timeout = loadbalancertimeout
+				if timeout <= 0 {
+					timeout = types.DefaultTimeout
+				}
```
fgiorgetti, line 1201: Same here, if you set the timeout into the `options SiteConfig` variable, you won't need this.

cmd/site-controller/controller.go
```diff
@@ -253,7 +253,7 @@ func (c *SiteController) checkSite(key string) error {
 			log.Println("Initialising skupper site ...")
 			siteConfig, _ := c.vanClient.SiteConfigInspect(context.Background(), configmap)
 			siteConfig.Spec.SkupperNamespace = siteNamespace
-			err = c.vanClient.RouterCreate(context.Background(), *siteConfig)
+			err = c.vanClient.RouterCreate(context.Background(), *siteConfig, types.DefaultTimeout)
```
fgiorgetti, line 256: Here is another place to set the `*siteConfig.Spec.LoadBalancerTimeout`.

cmd/skupper/skupper.go
```diff
@@ -476,6 +481,8 @@ installation that can then be connected to other skupper installations`,
 	cmd.Flags().StringVar(&routerCreateOpts.ConfigSync.CpuLimit, "config-sync-cpu-limit", "", "CPU limit for config-sync pods")
 	cmd.Flags().StringVar(&routerCreateOpts.ConfigSync.MemoryLimit, "config-sync-memory-limit", "", "Memory limit for config-sync pods")
 
+	cmd.Flags().IntVar(&LoadBalancerTimeout, "timeout", 120, "Configurable timeout in seconds for the ingress loadbalancer option")
```
fgiorgetti, line 484: instead of using the literal 120, use the constant DefaultTimeout

nluaces: Right, thanks!

api/types/client.go
nluaces: I did not want to store that information in the skupper internal yaml, because I thought that was not relevant in that file. But probably is a better outcome than changing the signature.

grs: I don't think this makes sense as a setting in the skupper-site configmap. Logically, it is really a configuration option for the CLI, not the site. The ideal solution for declarative use would be something more adaptive, or at least asynchronous. So to me, the signature change feels 'cleanest' for a simple improvement on current behaviour.

fgiorgetti: Considering that all fields in the SiteConfig spec are persisted, it makes sense to remove it and simply use as a flag.

grs: @nluaces actually I think that is a great idea. Setting a deadline on the context seems a very good approach to allowing a timeout to be specified when using the cli. That same approach could eventually be applied more uniformly.

fgiorgetti: :+1: that is something that we should explore more for request scoped contexts for sure.

client/router_create.go
```diff
@@ -1094,7 +1094,7 @@ sasldb_path: /tmp/skrouterd.sasldb
 					if err == nil {
 						host := kube.GetLoadBalancerHostOrIP(service)
 
-						var timeout = loadbalancertimeout
+						var timeout = ctx.Value(types.Timeout).(int)
```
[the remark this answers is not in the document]
nluaces: sure!

api/types/client.go
```diff
@@ -96,6 +96,7 @@ type SiteConfigSpec struct {
 	Router              RouterOptions
 	Controller          ControllerOptions
 	ConfigSync          ConfigSyncOptions
+	LoadBalancerTimeout int
```
grs, line 99: This is not needed any more is that right?

grs (approved):
I like it a lot!

api/types/client.go
nluaces: Correct! it is removed now, thanks!

fgiorgetti (approved):

PR #862, opened 2022-08-24
Added support for dumping namespace policies

### LINKED ISSUE #698: Add an option to get policies that dumps all policies that apply to the respective namespace
Suggestion for future enhancement would be to have an "inspect" or similar that enumerates the policies applied to the namespace (e.g. or of all the policies defined for the site applied to this namespace). This would also help for namespaces not in the allow list to review what is being applied.

_Originally posted by @ajssmith in https://github.com/skupperproject/skupper/pull/668#discussion_r827154987_

fgiorgetti:
Fixes #698

### DIFF
--- a/client/policy_validator.go
+++ b/client/policy_validator.go
@@ -289,6 +289,39 @@ func (p *ClusterPolicyValidator) ValidateImportService(serviceName string) *Poli
 	return res
 }
 
+func (p *ClusterPolicyValidator) Dump() *PolicyDump {
+	dump := &PolicyDump{
+		AllowIncomingLinks:            allowedBy{},
+		AllowedOutgoingLinksHostnames: allowedBy{},
+		AllowedExposedResources:       allowedBy{},
+		AllowedServices:               allowedBy{},
+	}
+	policies, err := p.LoadNamespacePolicies()
+	if err != nil {
+		log.Printf("error loading policies: %v", err)
+		return dump
+	}
+	for _, policy := range policies {
+		// Incoming links
+		if policy.Spec.AllowIncomingLinks {
+			dump.AllowIncomingLinks.Add("true", policy.ObjectMeta.Name)
+		}
+		// Outgoing link hostnames
+		for _, hostname := range policy.Spec.AllowedOutgoingLinksHostnames {
+			dump.AllowedOutgoingLinksHostnames.Add(hostname, policy.ObjectMeta.Name)
+		}
+		// Services
+		for _, service := range policy.Spec.AllowedServices {
+			dump.AllowedServices.Add(service, policy.ObjectMeta.Name)
+		}
+		// Exposed resources
+		for _, resource := range policy.Spec.AllowedExposedResources {
+			dump.AllowedExposedResources.Add(resource, policy.ObjectMeta.Name)
+		}
+	}
+	return dump
+}
+
 type PolicyAPIClient struct {
 	cli *VanClient
 }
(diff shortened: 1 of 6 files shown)


nluaces (approved):
lgtm, thanks!

ajssmith (approved):
lgtm

does debug dump include polices?

fgiorgetti:
> lgtm
> 
> does debug dump include polices?

Good point. I will include it too.

client/policy_validator.go
```diff
@@ -300,6 +333,19 @@ type PolicyAPIResult struct {
 	Error     string   `json:"error"`
 }
 
+type allowedBy map[string][]string
```
grs, line 336: Seems odd to me to have an unexported type for exported fields on an exported struct. Is that deliberate?

client/policy_validator.go
```diff
@@ -300,6 +333,19 @@ type PolicyAPIResult struct {
...
+
+func (a allowedBy) Add(key, policy string) {
+	a[key] = append(a[key], policy)
+}
+
+type PolicyDump struct {
```
grs, line 342: Minor personal nit pick, not crucial, PolicyInfo seems nicer and more informative than PolicyDump.

cmd/service-controller/console_server.go
```diff
@@ -417,6 +417,7 @@ func (server *ConsoleServer) listen() {
 	r.Handle("/policy/service/{name}", authenticated(server.policies.service()))
 	r.Handle("/policy/incominglink", authenticated(server.policies.incomingLink()))
 	r.Handle("/policy/outgoinglink/{hostname}", authenticated(server.policies.outgoingLink()))
+	r.Handle("/policy/dump", authenticated(server.policies.dump()))
```
grs, line 420: Again, 'dump' to me feels a bit odd in the resource path. Maybe /policy/all? Again though, this is a minor personal nit pick, not necessarily an issue.

cmd/service-controller/policies.go
```diff
@@ -125,3 +126,38 @@ func (p *PolicyManager) outgoingLink() http.Handler {
 		}
 	})
 }
+
+func (p *PolicyManager) writeAllowedByInfo(tw *tabwriter.Writer, title string, allowedInfo map[string][]string) {
```
grs, line 130: Why is this a method on PolicyManager? It doesn't seem to use any members of that struct. If the allowedBy type was exported, this method could perhaps be defined on that?

Even if not, perhaps this could be moved into that package and have a WriteTable() method o the PolicyDebug struct?

grs (approved):

client/policy_validator.go
fgiorgetti: Intention was just to prevent it to be used externally, but I agree it feels odd.
I will make it an exported type.

client/policy_validator.go
fgiorgetti: :+1:

cmd/service-controller/console_server.go
fgiorgetti: Policy all or policy list might be more meaningful. With that, I also need to change the
command to `get policies list` (or all). 
@grs @ajssmith @nluaces which one sounds better?

cmd/service-controller/policies.go
fgiorgetti: It is certainly misplaced.
I will move it into the renamed `PolicyInfo` type as a `String()` (stringer implementation).

grs (approved):

hash-d:
I like that we'll have that both on `get` and on the dump.  It would be nice to have a `skupper` subcommand to access it, but running `get` within the pod is not too much trouble.  I haven't reviewed the code or tested it in depth, but I think this will be invaluable in testing and support.

hash-d (approved):

ajssmith (approved):

PR #867, opened 2022-09-02
Fixed MongoDB integration test with container not found

fgiorgetti:


fgiorgetti:
It prevents the following error:
```
error: unable to upgrade connection: container not found ("mongo-a")
```

Which was happening because the pods were not yet ready when skupper was trying to create the replicaset.

hash-d:
@fgiorgetti, as we discussed offline, I'll add this to my https://github.com/hash-d/skupper/tree/dh-mongo-replicaset-fix, so we put both fixes in a single pull

fgiorgetti:
Closing this one as @hash-d will incorporate it in his other PR

hash-d:
Fix incorporated:  https://github.com/skupperproject/skupper/commit/5c60cad3d77929abdbee6e9fa634bbd1c50dc621

PR #868, opened 2022-09-02
Enable tls for http1 and tcp protocols

nluaces:
pending to be able to create tcpConnectors/tcpListeners with sslProfiles in the router.


nluaces:
closed due the obsolete status of the branch, and because the changes in the http connector/listener will come in the future.

PR #880, opened 2022-09-14
Environment variable to skip debug mode on integration tests

fgiorgetti:


rafaelvzago:
It worked on QE suite. It's good to go.

rafaelvzago:
One change that it's needed:

1. HelloWorld CLI has the debug hardcoded, and we need to remove this parameter since the constraints handle all the config.

Issue #888, opened 2022-09-20
Arm64 support for the CLI


aureamunoz:
Hello!
I have a Mac M1 and would like to install skupper, would it be possible to have the corresponding package? Currently it seems only amd is supported

Thank you in advance!

ted-ross:
I assume we will also need companion issues to produce aarch64 images for skupper-router and config-sync.

grs:
arm64 images is tracked by https://github.com/skupperproject/skupper/issues/884

lucacalcaterra:
@grs @ted-ross at the moment can i use arm64 images ? On quay.io seems missing arm images

grs:
No, we do not as yet have any arm64 images published.

PR #889, opened 2022-09-20
add support for mac on arm64

grs:


grs:
I believe this is all we need to produce an arm64 cli build for macs as requested in issue #888 ?

nluaces:
This is old (2020) issue:

> The problem is that darwin-arm64 does not support internal linking, which means that Go has to invoke the C linker, which means that you need to have a cross-linker from your host system to darwin-arm64. The Go linker is invoking the C linker, but it's not a cross-linker, so it fails.

source: https://github.com/golang/go/issues/39033

Is it possible that it has been fixed at this point?

nluaces:
Never mind, Seems like it is fixed since 1.16:

> Darwin and iOS
> Go 1.16 adds support of 64-bit ARM architecture on macOS (also known as Apple Silicon) with GOOS=darwin, GOARCH=arm64. Like the darwin/amd64 port, the darwin/arm64 port supports cgo, internal and external linking, c-archive, c-shared, and pie build modes, and the race detector.

https://go.dev/doc/go1.16#darwin

nluaces (approved):

grs:
Yes, I tried out the options and it built ok. Resulting binary seems to run ok also on Mac on M1.

fgiorgetti (approved):
I can't test, but it looks right to me.

PR #890, opened 2022-09-20
Container client with podman using libpod

fgiorgetti:
This PR introduces a client interface for manipulating containers (see: `./client/container`),
with our own entities representing container related objects.

A podman implementation is also provided (see: `./client/podman`) and it uses libpod restful
clients so that we can manipulate container objects through either the CLI or through a running
container, like a future implementation of service-controller.

The client stubs have not been added to the repository, they are generated if the target directory
does not exist `./client/generated/libpod`.

To generate the clients, you should run: `make generate-client` and in case you need to
regenerate or update them, run: `make force-generate-client`.

CI has been adjusted to generate the client stubs before starting the build process.


fgiorgetti:
> LGTM. Does this tie us tightly to any particular version of podman?

It is using libpod version 4.0.3 at the moment, which seems to work just fine with 4.1.x.
I still need to validate it against 4.2.x as well, but I don't think it will keep us tightly tied,
because we should be able to provide a different implementation of the `container.Client`
interface if we find it needed.

Thanks for reviewing it.

fgiorgetti:
> I still need to validate it against 4.2.

I have just tested against 4.2 and it seems to work just fine as well.

nluaces (approved):

PR #893, opened 2022-09-26
support for latest ingress api

grs:
If v1beta1 of ingress is not available, this patch uses v1

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -30,11 +30,12 @@ var defaultRetry = wait.Backoff{
 
 // A VAN Client manages orchestration and communications with the network components
 type VanClient struct {
-	Namespace     string
-	KubeClient    kubernetes.Interface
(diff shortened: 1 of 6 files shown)


fgiorgetti (approved):
It seems to work just fine with 1.22+ (v1) and older versions (v1beta1).

One additional change that could be incorporated is the `ingressClassName`
spec field, since the `kubernetes.io/ingress.class` is considered deprecated
(although nginx still supports it).

https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress

grs:
> One additional change that could be incorporated is the `ingressClassName` spec field, since the `kubernetes.io/ingress.class` is considered deprecated (although nginx still supports it).

Done in second commit

fgiorgetti (approved):
LGTM!

Issue #841, opened 2022-08-03
Skupper images run as root


ted-ross:
Kubernetes platforms configured to prevent containers from running as root can't deploy Skupper because the Skupper images run as root by default.

jiridanek:
> Kubernetes platforms configured to prevent containers from running as root

Doesn't that describe every default installation of OpenShift? Why it did not bother anyone who uses OpenShift long time before now?

PR #937, opened 2022-11-02
Issue #841: Skupper images run as root

ajssmith:
Setup user for container and define PodSecurityContext for deployments

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -1081,17 +1081,17 @@ sasldb_path: /tmp/skrouterd.sasldb
 	}
 
 	runAsNonRoot := true
-	podSecurityContext := corev1.PodSecurityContext{
+	securityContext := corev1.SecurityContext{
 		RunAsNonRoot: &runAsNonRoot,
 	}
 	if options.Spec.RunAsUser > 0 {
-		podSecurityContext.RunAsUser = &options.Spec.RunAsUser
+		securityContext.RunAsUser = &options.Spec.RunAsUser
 	}
 	if options.Spec.RunAsGroup > 0 {
-		podSecurityContext.RunAsGroup = &options.Spec.RunAsGroup
+		securityContext.RunAsGroup = &options.Spec.RunAsGroup
 	}
 
-	dep, err := kube.NewTransportDeployment(van, siteOwnerRef, &podSecurityContext, cli.KubeClient)
+	dep, err := kube.NewTransportDeployment(van, siteOwnerRef, &securityContext, cli.KubeClient)
 	if err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


nluaces (approved):

Dockerfile.config-sync
```diff
@@ -9,9 +9,14 @@ RUN go mod download
...
 RUN make
 
-
 FROM registry.access.redhat.com/ubi8-minimal
 
+# Create user and group and switch to user's context
+RUN microdnf -y install shadow-utils \
+&& microdnf clean all
+RUN useradd --uid 10000 runner
+USER 10000
```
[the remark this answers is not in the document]
ajssmith: A nice feature of OpenShift. On other platforms, containers are running as root. Hardening best practices suggest running as non-root user.

fgiorgetti (commented):
It seems to work for me (locally), but it would be nice to test as well against
multiple clusters and versions to make sure it is OK.

@granzoto I believe you have done similar evaluation recently.
Could you validate this one as well?

grs (approved):

cmd/site-controller/deploy-watch-all-ns.yaml
```diff
@@ -135,3 +135,11 @@ spec:
...
       - name: site-controller
         image: quay.io/skupper/site-controller:master
+        securityContext:
+          capabilities:
+            drop:
+            - ALL
+          runAsNonRoot: true
+          allowPrivilegeEscalation: false
+          seccompProfile:
+            type: RuntimeDefault
```
fgiorgetti, line 145: The `seccompProfile` field does not work with OKD 3.11.
All other fields seem to be working just fine.

Do you think it is better to keep it as is and just drop a note
somewhere explaining to remove it if running on 3.11 ?

client/router_create.go
```diff
@@ -1079,7 +1079,19 @@ sasldb_path: /tmp/skrouterd.sasldb
 			}
 		}
 	}
-	dep, err := kube.NewTransportDeployment(van, siteOwnerRef, cli.KubeClient)
+
+	runAsNonRoot := true
+	podSecurityContext := corev1.PodSecurityContext{
```
fgiorgetti, line 1084: One comment/question here. The `corev1.PodSecurityContext` type is not
the same one we are setting at the YAML files, which uses `v1.SecurityContext` type,
and that one is defined at the container level.

Should the same fields be set on both places?

fgiorgetti:
Just some more context:
https://sdk.operatorframework.io/docs/best-practices/pod-security-standards/#how-should-i-configure-my-operators-and-operands-to-comply-with-the-criteria

cmd/site-controller/deploy-watch-all-ns.yaml
ajssmith: Lean towards keeping as is and will add a comment to indicate version considerations

client/router_create.go
ajssmith: Agree and will update

granzoto (approved):
LGTM

PR #949, opened 2022-11-10
Add readiness probe to router

grs:
Add liveness and readiness probes to service controller Add readiness probe to config sync (ready once initial download of any certs is complete)

### DIFF
--- a/cmd/service-controller/controller.go
+++ b/cmd/service-controller/controller.go
@@ -527,6 +527,7 @@ func (c *Controller) initialiseServiceBindingsMap() (map[string][]int, error) {
 	c.ports.inuse(int(types.InterRouterListenerPort))
 	c.ports.inuse(int(types.ConsoleDefaultServicePort))
 	c.ports.inuse(9090) // currently hardcoded in config
+	c.ports.inuse(9191) // currently hardcoded in config
 	return allocations, nil
 
 }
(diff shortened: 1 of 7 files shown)


ajssmith (approved):
lgtm

built images and observed readiness checks succeeding

cmd/service-controller/controller.go
```diff
@@ -527,6 +527,7 @@ func (c *Controller) initialiseServiceBindingsMap() (map[string][]int, error) {
 	c.ports.inuse(int(types.InterRouterListenerPort))
 	c.ports.inuse(int(types.ConsoleDefaultServicePort))
 	c.ports.inuse(9090) // currently hardcoded in config
+	c.ports.inuse(9191) // currently hardcoded in config
```
fgiorgetti, line 530: Should 8182 be added here as well?

pkg/kube/containers.go
```diff
@@ -35,6 +35,37 @@ func ContainerForController(ds types.DeploymentSpec) corev1.Container {
 		ImagePullPolicy: GetPullPolicy(ds.Image.PullPolicy),
 		Name:            types.ControllerContainerName,
 		Env:             ds.EnvVar,
+		Ports:           ds.Ports,
+		StartupProbe: &corev1.Probe{
```
fgiorgetti, line 39: Maybe, in an separate PR/Issue, this could also be done as part of a `skupper update` to existing deployments.

fgiorgetti (approved):
A few comments only.
But it works OK and looks good.

nluaces (approved):

pkg/kube/containers.go
grs: Good point, updated now

PR #952, opened 2022-11-17
Enable TLS with custom certs

nluaces:
#### Changes in the CLI:
- The `enable-tls` flag in `create` and `expose` commands is now deprecated and substituted by `generate-tls-secrets` that works exactly the same.
- Expose command new flags: `tls-cert` for the secret name that stores server certificates and `tls-trust` for the secret name that stores the CA file
- Create command new flag: `tls-cert`
- Bind command new flag: `tls-trust`

#### Additional notes:
- New secrets with certificates generated by Skupper will be deleted once the service is unexposed/deleted as before. However, secrets with custom certificates will not be deleted from k8s.
- When un-exposing a service with custom certificates, their sslProfiles will not be deleted, to benefit using the same certs for different services in the same cluster.
- In order to work with custom secrets, they must be created with the same name in all the clusters included in the VAN.
- Created Annotations `tls-trust` and  `tls-cert`


cmd/service-controller/controller.go
```diff
@@ -905,3 +847,31 @@ func (c *Controller) updateServiceBindings(required types.ServiceInterface, port
 	}
 	return nil
 }
+
+func (c *Controller) getSecret(name string) (*corev1.Secret, error) {
+	return c.vanClient.KubeClient.CoreV1().Secrets(c.vanClient.Namespace).Get(name, metav1.GetOptions{})
+}
+
+func (c *Controller) getConfigMap() (*corev1.ConfigMap, error) {
```
grs, line 855: Very minor point, really just a personal preference, but getConfigMap() is a little ambiguous. Something like getRouterConfigMap() or getTransportConfigMap() would be more immediately obvious. However it is quite obvious when you look at the definition, so not a real problem.

grs (approved):

pkg/kube/qdr/tls_handler.go
```diff
@@ -0,0 +1,148 @@
...
+type ExistsSslProfileFunc func(string) (bool, error)
+type RemoveSslProfileFunc func(string) error
+
+type TlsSupport struct {
+	Address       string
+	Credentials   string
+	CertAuthority string
+}
+
+func EnableTlsSupport(support TlsSupport, getSecret GetSecretFunc, getConfigMap GetConfigMapFunc, newSecret NewSecretFunc, addSslProfile AddSslProfileFunc, existsSslProfile ExistsSslProfileFunc) error {
```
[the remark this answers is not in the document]
nluaces: I think it is a great idea, I'll work on an update. Thanks!

client/client.go
```diff
@@ -118,3 +122,31 @@ func (cli *VanClient) GetIngressDefault() string {
 	}
 	return types.IngressRouteString
 }
+
+func (cli *VanClient) GetSecret(name string) (*corev1.Secret, error) {
```
ajssmith, line 126: Suggest not introducing the wrappers to the client package. Is it necessary?

fgiorgetti (commented):
I noticed that if I have 2 linked sites and the "secret" does no exist on the other side,
the config sync will fail to create the sslProfile as it does not exist there, but it defined
the tcp/http listener and connector referring to the sslProfile that does not exist.

With that, if the router pod is deleted, it won't start again, as there is a config file issue.
So I think it might be better to log an error when service sync tries to define the service
with TLS and the respective secret does not exist, and do not create the
listener/connector in the skupper-internal configmap.

api/types/types.go
```diff
@@ -467,8 +469,8 @@ type ServiceInterface struct {
 	Labels                   map[string]string        `json:"labels,omitempty" yaml:"labels,omitempty"`
 	Targets                  []ServiceInterfaceTarget `json:"targets" yaml:"targets,omitempty"`
 	Origin                   string                   `json:"origin,omitempty" yaml:"origin,omitempty"`
-	EnableTls                bool                     `json:"enableTls,omitempty"`
 	TlsCredentials           string                   `json:"tlsCredentials,omitempty"`
+	TlsCertAuthority         string                   `json:"certAuthority,omitempty"`
 	PublishNotReadyAddresses bool                     `json:"publishNotReadyAddresses,omitempty"`
 }
```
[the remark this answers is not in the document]
nluaces: I updated the code to check first if the secrets in the desired definition exist before creating the ingress/egress bridges. If the secret does not exist, the error will be shown with the `skupper debug events` command. To solve this, the user should create the secret in the cluster, delete the service in that cluster, and the service sync will create the service again successfully.

cmd/skupper/skupper.go
```diff
@@ -458,7 +463,14 @@ func NewCmdExpose(skupperCli SkupperServiceClient) *cobra.Command {
 	cmd.Flags().StringVar(&(exposeOpts.Address), "address", "", "The Skupper address to expose")
 	cmd.Flags().IntSliceVar(&(exposeOpts.Ports), "port", []int{}, "The ports to expose on")
 	cmd.Flags().StringSliceVar(&(exposeOpts.TargetPorts), "target-port", []string{}, "The ports to target on pods")
-	cmd.Flags().BoolVar(&exposeOpts.EnableTls, "enable-tls", false, "If specified, the service will be exposed over TLS (valid only for http2 and tcp protocols)")
+	cmd.Flags().BoolVar(&exposeOpts.GeneratedCerts, "enable-tls", false, "If specified, the service will be exposed over TLS (valid only for http2 and tcp protocols)")
```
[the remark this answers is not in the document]
nluaces: `enable-tls` it is now deprecated and it will be replaced by `generate-tls-secrets`; I was thinking on keeping the deprecated one for now. What do you think?

cmd/skupper/skupper.go
```diff
@@ -585,6 +597,7 @@ func NewCmdService() *cobra.Command {
 }
 
 var serviceToCreate types.ServiceInterface
+var createSvcWithGeneratedTlsCerts bool
```
ajssmith, line 600: Could this field be added to types.ServiceInterface or is there a reason to separate?

nluaces: Actually, I could use `GeneratedCerts` instead, and it is already included in types.ServiceInterface. Thanks for noticing, I will update this.

pkg/kube/qdr/tls_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	RemoveSslProfile(string) error
+}
+
+type TlsServiceSupport struct {
+	Address       string
+	Credentials   string
+	CertAuthority string
+}
+
+func EnableTlsSupport(support TlsServiceSupport, tlsManager TlsManagerInterface) error {
```
ajssmith, line 31: Suggest that this be part of TlsManagerInterface as it appears to be associated to the abstraction

pkg/kube/qdr/tls_handler.go
```diff
@@ -0,0 +1,206 @@
...
+		err = checkAndIncludeSslProfile(sslProfile, tlsManager)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func DisableTlsSupport(credentials string, tlsManager TlsManagerInterface) error {
```
ajssmith, line 78: Also suggest adding to TlsManagerInterface

pkg/kube/qdr/tls_handler.go
```diff
@@ -0,0 +1,206 @@
...
+			_, err := cli.CoreV1().Secrets(namespace).Get(service.TlsCertAuthority, metav1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("SslProfile %s for service %s does not exist in this cluster", service.TlsCertAuthority, service.Address)
+			}
+		}
+	}
+	return nil
+}
+
+func includeSslProfile(sslProfile qdr.SslProfile, tlsManager TlsManagerInterface) error {
```
ajssmith, line 118: Suggest removing this wrapper func as calling tlsManager.AddSslProfile directly accomplishes same

ajssmith (commented):
This looks good. A few comments to consider, thnx

fgiorgetti:
When using `--tls-cert` or `--tls-trust` it would be nice to validate if the provided secrets
contain the expected keys: `tls.key`, `tls.crt` and `ca.crt`, otherwise users will only notice
that it is not working by going through the router logs.

fgiorgetti:
Another issue I noticed. If I create a service, lets's say: `skupper service create my-service 8080 --generate-tls-secrets`,
it will create a new secret named `skupper-tls-my-service`.

Next if I try to reuse that secret for another service, using:
`skupper service create my-new-service 8080 --tls-cert skupper-tls-my-service`, it works ok.

But then, if I delete the original `my-service` the secret goes away.

I believe it needs to ensure no other service is using the secret before removing it.

cmd/skupper/skupper.go
```diff
@@ -666,6 +692,8 @@ func NewCmdBind(skupperClient SkupperServiceClient) *cobra.Command {
 	}
 	cmd.Flags().StringVar(&protocol, "protocol", "", "The protocol to proxy (tcp, http or http2).")
 	cmd.Flags().StringSliceVar(&targetPorts, "target-port", []string{}, "The port the target is listening on (you can also use colon to map source-port to a target-port).")
+	cmd.Flags().StringVar(&tlsCertAuthority, "tls-trust", "", "K8s secret name with the CA to expose the service over TLS (valid only for http2 and tcp protocols)")
```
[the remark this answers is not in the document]
nluaces: Nice catch! It will be fixed in my next update.

pkg/kube/qdr/tls_handler.go
nluaces: Done!

pkg/kube/qdr/tls_handler.go
nluaces: Done!

pkg/kube/qdr/tls_handler.go
nluaces: Done!

fgiorgetti (commented):
A few minor comments only, but overall it seems to be working just fine.

api/types/types.go
```diff
@@ -186,6 +186,8 @@ const (
 	StatusAnnotationKey         string = InternalQualifier + "/status"
 	GatewayQualifier            string = InternalQualifier + "/gateway"
 	IngressOnlyQualifier        string = BaseQualifier + "/ingress-only"
+	TlsCertQualifier            string = BaseQualifier + "tls-cert"
+	TlsTrustQualifier           string = BaseQualifier + "tls-trust"
```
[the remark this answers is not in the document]
nluaces: Thanks! Updated in 89c3650e1df68b9ef5fb7007349c0dd74bd03525

client/serviceinterface_create.go
```diff
@@ -22,44 +19,16 @@ func (cli *VanClient) ServiceInterfaceCreate(ctx context.Context, service *types
...
-			}
-
-			err = qdr.AddSslProfile(serviceSecret.Name, cli.Namespace, cli.KubeClient)
-			if err != nil {
-				return err
-			}
-
+		tlsSupport := kubeqdr.TlsServiceSupport{Address: service.Address, Credentials: service.TlsCredentials}
+		tlsManager := &kubeqdr.TlsManager{KubeClient: cli.KubeClient, Namespace: cli.Namespace}
+		err = tlsManager.EnableTlsSupport(tlsSupport)
```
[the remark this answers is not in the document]
nluaces: Good observation, thanks! Updated in 62275c73066a8c3e39cdc3670c0e78598fda37da

client/serviceinterface_remove.go
```diff
@@ -37,7 +35,17 @@ func (cli *VanClient) ServiceInterfaceRemove(ctx context.Context, address string
...
-					handleServiceCertificateRemoval(address, cli)
+					if len(service.TlsCredentials) > 0 {
+						tlsManager := qdr.TlsManager{KubeClient: cli.KubeClient, Namespace: cli.Namespace}
+						serviceList, err := cli.ServiceInterfaceList(ctx)
+						if err != nil {
+							return err
+						}
+
+						tlsSupport := qdr.TlsServiceSupport{Address: service.Address, Credentials: service.TlsCredentials}
+						tlsManager.DisableTlsSupport(tlsSupport, serviceList)
```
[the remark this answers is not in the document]
nluaces: Updated in 62275c73066a8c3e39cdc3670c0e78598fda37da

fgiorgetti (approved):

ajssmith (approved):

PR #966, opened 2022-12-07
Support for disabling service ingress

grs:


nluaces (approved):

cmd/skupper/skupper.go
```diff
@@ -615,12 +624,17 @@ func NewCmdCreateService(skupperClient SkupperServiceClient) *cobra.Command {
...
+			if err != nil {
+				return err
+			}
 
 			return skupperClient.Create(cmd, args)
 		},
 	}
 	cmd.Flags().StringVar(&serviceToCreate.Protocol, "protocol", "tcp", "The mapping in use for this service address (tcp, http, http2)")
 	cmd.Flags().StringVar(&serviceToCreate.Aggregate, "aggregate", "", "The aggregation strategy to use. One of 'json' or 'multipart'. If specified requests to this service will be sent to all registered implementations and the responses aggregated.")
+	cmd.Flags().StringVar(&serviceIngressMode, "ingress-mode", "", "Controls whether a kubernetes service is created locally for the skupper service. Valid values are Always (default), Never or IfNoLocalTargets.")
```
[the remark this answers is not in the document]
grs: Fixed

api/types/types.go
```diff
@@ -480,6 +490,17 @@ func IsOfLocalOrigin(origin string) bool {
 	return origin == "" || origin == "annotation"
 }
 
+func (s *ServiceInterface) SetIngressMode(mode string) error {
+	if mode == string(ServiceIngressModeAlways) {
```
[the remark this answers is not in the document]
grs: Fixed

fgiorgetti (approved):

ajssmith (approved):
lgtm, re-checked

cmd/skupper/skupper.go
```diff
@@ -459,6 +466,7 @@ func NewCmdExpose(skupperCli SkupperServiceClient) *cobra.Command {
 	cmd.Flags().IntSliceVar(&(exposeOpts.Ports), "port", []int{}, "The ports to expose on")
 	cmd.Flags().StringSliceVar(&(exposeOpts.TargetPorts), "target-port", []string{}, "The ports to target on pods")
 	cmd.Flags().BoolVar(&exposeOpts.EnableTls, "enable-tls", false, "If specified, the service will be exposed over TLS (valid only for http2 and tcp protocols)")
+	cmd.Flags().StringVar(&(exposeOpts.IngressMode), "ingress-mode", "", "Controls whether a kubernetes service is created locally for the exposed service. Valid values are Always (default) or Never.")
```
pwright, line 469:
```suggestion
	cmd.Flags().StringVar(&(exposeOpts.IngressMode), "ingress-mode", "", "Controls whether a Kubernetes service is created in the site namespace when you use Skupper to expose a service.  'Always' (default) and  'Never' are valid values.")
```
@grs I'm not sure whether my suggestion is valid? How could we describe the use case where users would use this option.
eg If you are using something like Istio, you might want to prevent Skupper from creating and clobbering services.

cmd/skupper/skupper.go
```diff
@@ -615,12 +624,17 @@ func NewCmdCreateService(skupperClient SkupperServiceClient) *cobra.Command {
...
+			if err != nil {
+				return err
+			}
 
 			return skupperClient.Create(cmd, args)
 		},
 	}
 	cmd.Flags().StringVar(&serviceToCreate.Protocol, "protocol", "tcp", "The mapping in use for this service address (tcp, http, http2)")
 	cmd.Flags().StringVar(&serviceToCreate.Aggregate, "aggregate", "", "The aggregation strategy to use. One of 'json' or 'multipart'. If specified requests to this service will be sent to all registered implementations and the responses aggregated.")
+	cmd.Flags().StringVar(&serviceIngressMode, "ingress-mode", "", "Controls whether a kubernetes service is created locally for the skupper service. Valid values are Always (default) or Never.")
```
ssorj, line 637: --ingress-mode doesn't quite fit something with "always" and "never" for me.  Do you conceive of it having any other modes?

Perhaps --create-kube-service (always|never)?   I know it's a bit verbose.  Maybe --create-ingress (always|never).

grs: > Do you conceive of it having any other modes?

Possibly. The original PR had a 'IfNoLocalTargets' option as well, but I removed that for simplicity, and as that is a somewhat speculative mode.

grs: Maybe `--enable-access-from-target-site`? 'Determines whether access to the Skupper service is enabled in the site the target is exposed in'. (@pwright is this any clearer in your view?)

pwright: @grs yes, or `create-service-in-target-site` ? 'Determines whether a Skupper service is created in target site'

ssorj: I like that `--enable-access-from-target-site` is not Kube-specific.  "Determines whether access to the Skupper service is enabled in the site the target is exposed in."  The "exposed in" is ambiguous for me.  Where the target is running?

grs: Yes, 'where the target is running' is better, but we will soon have the ability to have targets in a namespace other than that which skupper is running in. In those cases we will still create the service in the skupper site namespace. So... what about: "Determines whether access to the Skupper service is enabled in the site the target was exposed through." ?? Any clearer?

ssorj: Yes, that is clearer for me.

grs: I went with enable-ingress rather than enable-access, as there are actually two flags involved (one on skupper expose one on skupper service create) and access seemed more vague for the second. Wdyt?

PR #967, opened 2022-12-07
Prevents definition monitor from updating stale service

fgiorgetti:


nluaces (approved):

hash-d (commented):
I've ran my changed HelloWorld that was able to detect this problem 25 times in a row, without any errors, using the code from the PR, so it looks good.

I'll let a longer test running for master, then open a PR to include that test on master.

hash-d:
On Friday, I've rebased and left it running; it ran 27 times before failing on my local (but I think that was some environment problem).  Today, it ran another 60 times without failure.  I'll rebase it again and submit the PR.

PR #968, opened 2022-12-12
Generate K8s events

nluaces:
This change consist of generating k8s events about resources created by Skupper in the cluster or any other event that could provide value to the user:
- Create/Delete services
- Create/Delete links (imply the creation of secrets)
- Service Sync: Connection established
- Policy: Service is now allowed
- Service is exposed
- Tokens: Connect/Disconnect

The controller has been refactored to use a component that implements `EventHandlerInterface`, and the recorder was removed from the VanClient:
```mermaid
classDiagram

class EventHandlerInterface

<<interface>> EventHandlerInterface

EventHandlerInterface <|.. DefaultEventLogger

EventHandlerInterface <|.. SkupperEventRecorder

EventHandlerInterface : +RecordWarningEvent()

EventHandlerInterface : +RecordNormalEvent()

class DefaultEventLogger{

-store *EventStore

}

class SkupperEventRecorder{

+ EventRecorder record.EventRecorder

+ Source *v1.Service
}

```

The k8s event recorder is configured by default. In order to disable it and use the DefaultStore,  it would be necessary to specify `--enable-skupper-events=false` in `skupper init`.

### DIFF
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -126,11 +126,6 @@ var ControllerPolicyRule = []rbacv1.PolicyRule{
 		APIGroups: []string{"networking.k8s.io"},
 		Resources: []string{"ingresses"},
 	},
-	{
-		Verbs:     []string{"watch", "create", "patch"},
-		APIGroups: []string{""},
-		Resources: []string{"events"},
-	},
 }
 
 // Certifcates/Secrets constants
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -126,6 +126,11 @@ var ControllerPolicyRule = []rbacv1.PolicyRule{
 		APIGroups: []string{"networking.k8s.io"},
 		Resources: []string{"ingresses"},
 	},
+	{
+		Verbs:     []string{"watch", "create", "patch"},
+		APIGroups: []string{""},
+		Resources: []string{"events"},
+	},
 }
 
 // Certifcates/Secrets constants
--- a/client/client.go
+++ b/client/client.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"k8s.io/client-go/tools/record"
 	"time"
 
 	openshiftapps "github.com/openshift/client-go/apps/clientset/versioned"
@@ -39,6 +40,7 @@ type VanClient struct {
 	DynamicClient   dynamic.Interface
 	DiscoveryClient *discovery.DiscoveryClient
 	LinkHandler     domain.LinkHandler
+	EventRecorder   record.EventRecorder
 }
 
 func (cli *VanClient) GetNamespace() string {
@@ -112,6 +114,7 @@ func NewClient(namespace string, context string, kubeConfigPath string) (*VanCli
 	} else {
 		c.Namespace = namespace
 	}
+	c.EventRecorder = kube.NewEventRecorder(c.Namespace, c.KubeClient)
 	c.DynamicClient, err = dynamic.NewForConfig(restconfig)
 	if err != nil {
 		return c, err
--- a/client/client.go
+++ b/client/client.go
@@ -113,7 +113,6 @@ func NewClient(namespace string, context string, kubeConfigPath string) (*VanCli
 	} else {
 		c.Namespace = namespace
 	}
-	c.EventRecorder = kube.NewEventRecorder(c.Namespace, c.KubeClient)
 	c.DynamicClient, err = dynamic.NewForConfig(restconfig)
 	if err != nil {
 		return c, err
--- a/client/client.go
+++ b/client/client.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"k8s.io/client-go/tools/record"
 	"time"
 
 	openshiftapps "github.com/openshift/client-go/apps/clientset/versioned"
@@ -40,7 +39,7 @@ type VanClient struct {
 	DynamicClient   dynamic.Interface
 	DiscoveryClient *discovery.DiscoveryClient
 	LinkHandler     domain.LinkHandler
-	EventRecorder   record.EventRecorder
+	EventRecorder   kube.SkupperEventRecorder
 }
 
 func (cli *VanClient) GetNamespace() string {
--- a/client/client.go
+++ b/client/client.go
@@ -39,7 +39,6 @@ type VanClient struct {
 	DynamicClient   dynamic.Interface
 	DiscoveryClient *discovery.DiscoveryClient
 	LinkHandler     domain.LinkHandler
-	EventRecorder   kube.SkupperEventRecorder
 }
 
 func (cli *VanClient) GetNamespace() string {
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -1242,11 +1242,13 @@ sasldb_path: /tmp/skrouterd.sasldb
 		}
 	}
 
-	///k8s kube config add option
+	var disableEvents = options.Spec.DisableSkupperEvents
 	err = kube.AddEventRecorderPermissions(van.Namespace, ownerRefs, cli.KubeClient, types.ControllerServiceAccountName)
 	if err != nil {
-		cli.EventRecorder.Enabled = false
+		disableEvents = true
 	}
+
+	cli.EventRecorder = kube.NewEventRecorder(van.Namespace, cli.KubeClient, disableEvents)
 
 	return nil
 }
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -466,11 +466,6 @@ func ClusterRoles() []*rbacv1.ClusterRole {
 				Resources: []string{"nodes"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
-			{
-				Verbs:     []string{"watch", "create", "patch"},
-				APIGroups: []string{""},
-				Resources: []string{"events"},
-			},
 		},
 	})
 	return clusterRoles
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -1240,6 +1240,7 @@ sasldb_path: /tmp/skrouterd.sasldb
 	var disableEvents = options.Spec.DisableSkupperEvents
 	err = kube.AddEventRecorderPermissions(van.Namespace, ownerRefs, cli.KubeClient, types.ControllerServiceAccountName)
 	if err != nil {
+		log.Printf("Failed to add permissions for the event recorder: %s\n", err)
 		disableEvents = true
 	}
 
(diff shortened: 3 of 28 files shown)


pkg/kube/events.go
```diff
@@ -0,0 +1,39 @@
...
+	eventRecorder := eventBroadcaster.NewRecorder(
+		scheme.Scheme,
+		v1.EventSource{
+			Component: types.ControllerServiceName})
+	return eventRecorder
+}
+
+func RecordWarningEvent(namespace string, reason string, message string, recorder record.EventRecorder, cli kubernetes.Interface) {
+	event.Recordf(reason, message)
+	deployment, _ := cli.CoreV1().Services(namespace).Get(types.ControllerServiceName, metav1.GetOptions{})
```
fgiorgetti, line 31: Since the deployment variable is always the same, why not retrieving it in the `NewEventRecorder` instead?
One issue here, is that the variable is named `deployment` but it is actually retrieving a `Service`.

nluaces: Nice catch! I will change the name and I will try to not get the service each time that a event needs to be sent.

api/types/types.go
```diff
@@ -119,6 +119,11 @@ var ControllerPolicyRule = []rbacv1.PolicyRule{
 		APIGroups: []string{"networking.k8s.io"},
 		Resources: []string{"ingresses"},
 	},
+	{
```
grs, line 122: It would be nice to not require this permission (in which case of course events would not be published this way). Not sure what is the simplest way to do that though... could perhaps try to a create a separate role and then  if that succeeds add a role binding for that (and if it fails just disable events).

client/router_create.go
```diff
@@ -446,6 +446,11 @@ func ClusterRoles() []*rbacv1.ClusterRole {
 				Resources: []string{"nodes"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{
+				Verbs:     []string{"watch", "create", "patch"},
```
grs, line 450: Out of curiosity, why do we need a cluster scoped permission for events?

nluaces: I will remove this. This pr only send events from the service-controller so it won't be necessary to have those permissions. Thanks!

nluaces:
As discussed, this events will be sent by default, but it would be helpful to have some sort of config value to disable this option. It will be necessary to check in the initialization the privileges regarding managing k8s events.

There is another issue related to checking privileges in Skupper; so, if no one disagrees, I will postpone that change. However, I will include the configurable option in this pr.

cc: @grs @ajssmith @fgiorgetti

api/types/types.go
nluaces: Updated!

client/router_create.go
```diff
@@ -466,6 +466,11 @@ func ClusterRoles() []*rbacv1.ClusterRole {
...
```
grs, line 470: Do we need this cluster scoped permission?

client/router_create.go
```diff
@@ -1237,6 +1242,14 @@ sasldb_path: /tmp/skrouterd.sasldb
 		}
 	}
 
+	var disableEvents = options.Spec.DisableSkupperEvents
+	err = kube.AddEventRecorderPermissions(van.Namespace, ownerRefs, cli.KubeClient, types.ControllerServiceAccountName)
+	if err != nil {
+		disableEvents = true
```
grs, line 1248: I feel we should at least log the error here. Alternatively we could return the error and therefore require events to be explicitly disabled. Or we could flip the default for events, and require them to be explicitly enabled, and again return the error if this was not possible.

client/client.go
```diff
@@ -39,6 +39,7 @@ type VanClient struct {
 	DynamicClient   dynamic.Interface
 	DiscoveryClient *discovery.DiscoveryClient
 	LinkHandler     domain.LinkHandler
+	EventRecorder   kube.SkupperEventRecorder
```
grs, line 42: Is this the right place to hold the recorder? At present it is intended to be used from the service controller. Would having it as a member of the controller be better?

cmd/service-controller/policy_controller.go
```diff
@@ -35,12 +36,13 @@ func AddStaticPolicyWatcher(pv *client.ClusterPolicyValidator) {
...
-	informer  cache.SharedIndexInformer
-	queue     workqueue.RateLimitingInterface
-	activeMap map[string]time.Time
+	name          string
+	cli           *client.VanClient
+	validator     *client.ClusterPolicyValidator
+	informer      cache.SharedIndexInformer
+	queue         workqueue.RateLimitingInterface
+	activeMap     map[string]time.Time
+	eventRecorder kube.SkupperEventRecorder
```
grs, line 45: This struct already has the cli, so there will be a recorder on that (though see comment above) and at this level, which feels redundant and perhaps error prone?

pkg/kube/events.go
```diff
@@ -0,0 +1,115 @@
...
+
+	eventRecorder := SkupperEventRecorder{
+		EventRecorder: kubeEventRecorder,
+		Source:        service,
+		Disabled:      disabled,
+	}
+	return eventRecorder
+}
+
+func RecordWarningEvent(reason string, message string, recorder SkupperEventRecorder) {
```
grs, line 57: What if we defined an interface under pkg/event with RecordWarning/RecordNormal functions, and then could create two implementations of that, one that just does what we do now, and one that records the event through kube as well, each with a distinct factory function.

grs (commented):
I think we should remove the DefaultStore and related interface in pkg/event if it is not being used, to avoid accidental use of it. (Alternatively we could alter it to be of a new interface type that could be initialised to use kube events or not. I.e. keep the package level static variable as the recorder. However I think in general an explicit variable elsewhere is probably clearer so removing this old approach would be better).

client/router_create.go
fgiorgetti: I like the idea of kubernetes events being generated by default. And considering that, I agree that
if events cannot be generated, users must be aware of that. A warning message on stdout or at least
an error message logged during the service controller startup.

cmd/service-controller/controller.go
```diff
@@ -176,6 +176,22 @@ func NewController(cli *client.VanClient, origin string, tlsConfig *tls.Config,
...
 
+	siteCreationTime := uint64(time.Now().UnixNano()) / uint64(time.Microsecond)
+	configmap, err := kube.GetConfigMap(types.SiteConfigMapName, cli.Namespace, cli.KubeClient)
+	if err == nil {
+		siteCreationTime = uint64(configmap.ObjectMeta.CreationTimestamp.UnixNano()) / uint64(time.Microsecond)
+	}
+	siteConfig, _ := controller.vanClient.SiteConfigInspect(context.TODO(), configmap)
+
+	var ttl time.Duration
+	var disableEvents = true
```
fgiorgetti, line 187: In the RouterCreate the default is set to false.
Shouldn't it be set to false by default here as well?

cmd/service-controller/services.go
```diff
@@ -298,7 +298,8 @@ func serveServices(m *ServiceManager) http.Handler {
 				if err != nil {
 					http.Error(w, err.Error(), http.StatusInternalServerError)
 				} else {
-					event.Recordf("Service %s exposed", options.GetServiceName())
+					message := fmt.Sprintf("Service %s exposed", options.GetServiceName())
```
fgiorgetti, line 301: Should an event be generated in the error case above?

pkg/kube/events.go
fgiorgetti: Another kinda related thing. Why not making `RecordWarningEvent` and `RecordNormalEvent` methods of `SkupperEventRecorder` ? That would eliminate the need of passing the SkupperEventRecorder instance
every time.

pkg/service_sync/messaging.go
```diff
@@ -1,6 +1,9 @@
 package service_sync
 
 import (
+	"fmt"
+	vanClient "github.com/skupperproject/skupper/client"
```
fgiorgetti, line 5: One concern I have with tying `service_sync` with `VanClient` is that on Podman
sites we won't have the event recorder available. So what if we have an
event logger interface and, in this case, the kubernetes implementation uses the
event recorder, but let it open for other platforms to use what suits them better?

This is not a problem for this PR, but something to discuss together considering that
podman is already in place (although service sync is not yet implemented for podman)
and eventually other platforms will be supported soon.

Maybe an EventLogger interface could be added under `pkg/domain` ?

client/client.go
nluaces: There are two k8s events that are sent outside the service-controller, in the service-sync; when the connection is established with the sender or the receiver.

grs: service sync is logically part of the service controller. Perhaps we could pass in an event handler when creating the service sync instance?

pkg/kube/events.go
grs: Agreed, that was implied by having an interface.

client/router_create.go
nluaces: Updated in 1b5d487c84b9ec4ead384eedf377225bd13e1936

client/router_create.go
nluaces: Updated in 0a62b447a69ae400b1b0326680a65a6e2f13bbde

client/client.go
nluaces: Updated in 16e762d1840cea04e30984d9be7b9ea58e89acdc

cmd/service-controller/controller.go
nluaces: You were right, but it is not going to be necessary because I've removed that flag.

cmd/service-controller/services.go
nluaces: True, updated in f21636ed7a402bec9cd6af48558a79b2497110a4

cmd/service-controller/policy_controller.go
nluaces: It was removed from the VanClient, now the PolicyController is instantiated with a `eventHandler` that comes from the controller.

grs (approved):
This looks really good, great work Noe!

client/router_create.go
```diff
@@ -1237,6 +1237,11 @@ sasldb_path: /tmp/skrouterd.sasldb
 		}
 	}
 
+	err = kube.AddEventRecorderPermissions(van.Namespace, ownerRefs, cli.KubeClient, types.ControllerServiceAccountName)
```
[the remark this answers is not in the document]
nluaces: Yes, you are correct. It is fixed now. Thanks!

fgiorgetti (approved):
Looks great to me! :shipit:

ajssmith (approved):
lgtm2

PR #970, opened 2022-12-14
Expose router feature dataConnectionCount with flag on link create co…

mgoulish:
…mmand

This allows user to control data connection count  (number of parallel data links) on inter-router connections between skupper sites. -- using a flag on the link create command.

This is my first Skupper PR, somebody please let me know what I missed...

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -17,6 +17,7 @@ type ConnectorCreateOptions struct {
 	SkupperNamespace string
 	Name             string
 	Cost             int32
+        Dcc              int
 }
 
 type ConnectorRemoveOptions struct {
@@ -29,6 +30,7 @@ type LinkStatus struct {
 	Name        string
 	Url         string
 	Cost        int
+	Dcc         int
 	Connected   bool
 	Configured  bool
 	Description string
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -168,6 +168,7 @@ const (
 	TokenGeneratedBy            string = BaseQualifier + "/generated-by"
 	SiteVersion                 string = BaseQualifier + "/site-version"
 	TokenCost                   string = BaseQualifier + "/cost"
+	TokenDcc                    string = BaseQualifier + "/dcc"
 	UpdatedAnnotation           string = InternalQualifier + "/updated"
 	AnnotationExcludes          string = BaseQualifier + "/exclude-annotations"
 	LabelExcludes               string = BaseQualifier + "/exclude-labels"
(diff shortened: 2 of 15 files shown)


api/types/client.go
```diff
@@ -17,6 +17,7 @@ type ConnectorCreateOptions struct {
 	SkupperNamespace string
 	Name             string
 	Cost             int32
+	Dcc              int
```
grs, line 20: Dcc is a bit opaque. Perhaps ConCount or something along those lines?

api/types/types.go
```diff
@@ -168,6 +168,7 @@ const (
 	TokenGeneratedBy            string = BaseQualifier + "/generated-by"
 	SiteVersion                 string = BaseQualifier + "/site-version"
 	TokenCost                   string = BaseQualifier + "/cost"
+	TokenDcc                    string = BaseQualifier + "/dcc"
```
grs, line 171: As above, particularly for the actual annotation key, which is user visible, lets have something a bit less cryptic.

cmd/skupper/skupper_link.go
```diff
@@ -34,6 +34,7 @@ func NewCmdLinkCreate(skupperClient SkupperLinkClient, flag string) *cobra.Comma
 	}
 	cmd.Flags().StringVarP(&connectorCreateOpts.Name, flag, "", "", "Provide a specific name for the link (used when deleting it)")
 	cmd.Flags().Int32VarP(&connectorCreateOpts.Cost, "cost", "", 1, "Specify a cost for this link.")
+	cmd.Flags().IntVarP(&connectorCreateOpts.Dcc, "dataConnectionCount", "", 1, "How many inter-router data connections")
```
grs, line 37: Other multi word flags use hyphens rather than camel case, so this so follow that pattern for consistency. I think the default should be 0, and that should be treated as unset, in which case a default would be applied by some other part of the code.

pkg/utils/configs/config.go
```diff
@@ -38,6 +38,7 @@ connector {
     name: {{.Name}}-connector
     host: {{.Host}}
     port: {{.Port}}
+    dataConnectionCount: {{.Dcc}}
```
grs, line 41: We should only set this if it is non zero. That way we have a way to revert to the router default if desired.

grs (commented):
This looks mostly fine to me on the whole, apart from the minor issues. However in general I think we would not expect users to need to set this and we should try to set it to the optimal value (or if there are clear trade-offs we should document those clearly and perhaps offer options in terms of those tradeoffs)

pkg/utils/configs/config.go
```diff
@@ -38,9 +44,8 @@ connector {
     name: {{.Name}}-connector
     host: {{.Host}}
     port: {{.Port}}
-    dataConnectionCount: {{.Dcc}}
     role: {{.Role}}
-    cost: {{.Cost}}
+    cost: {{.Cost}} {{if .ConCount}}{{printf "\n    dataConnectionCount:" }} {{.ConCount}}{{end}}
```
grs, line 48: Noting the comment above, I don't think this is the right way to handle the issue. Use `{{- if .ConCount}} ... {{- end}}` as in QdRouterConfig() method.

PR #973, opened 2022-12-15
Draft: update procedure enhancement proposal

fgiorgetti:
# Skupper update enhancement proposal

This is a draft proposal for an eventual enhancement to the update procedure used by Skupper.

It introduces a standard mechanism that allows update tasks to be implemented individually,
delegating the filtering and execution of those tasks to a generic processor function.

First you need to implement what are the update tasks to be performed.
Here is the `update.Task` interface defined at `pkg/update/types.go`:

```go
// Task is an update task that can be filtered and executed by the update process
type Task interface {
	// Info describes the update task
	Info() string
	// AppliesTo determines whether it should be executed or not
	AppliesTo(siteVersion string) bool
	// Version returns the version it has been introduced
	Version() string
	// Priority determines how tasks within the same version will be sorted
	Priority() Priority
	// Platforms contains all platforms this task applies to
	Platforms() []types.Platform
	// Run method is where the update task is done
	Run() Result
}
```

When a `Task` is executed, it returns a result. Its result contains
an eventual error and a `Stop` flag, which is used by the update process
to determine whether or not it has to be stopped.

Here is how the `update.Result` type looks:

```go
// Result contains an eventual error and a Stop flag to control the update process
type Result struct {
	Err  error
	Stop bool
}
```

Basically you have to write all your update tasks, either generically or using a platform based flavor.

Next the platform (kubernetes or podman) client must call the
`update.RegisterTask` function defined at `pkg/update/processor.go` to
provide all (pre-initialized) task implementations.

Once all tasks have been registered, the client must call the
`update.Process` function, providing the current site version to it.

Example:

```go
	update.RegisterTask(&all.UpdateDeployments{})
	update.RegisterTask(&v0_8_0.AddConfigSync{})
	update.RegisterTask(&v0_8_0.MultiplePorts{})
	update.RegisterTask(&v1_3_0.AddVault{})
	update.RegisterTask(&v1_3_0.UpdateAddVflowCollectorKube{})
	update.RegisterTask(&v1_3_0.PodmanController{})
	update.RegisterTask(&v0_7_0.Claims{})
	update.RegisterTask(&all.UpdatePodmanContainers{})
	update.RegisterTask(&v0_8_0.PodmanSwitch{})
    update.Process(siteVersion)
```

The update process will perform the following activities:

* Filter valid tasks per platform and run `AppliesTo(siteVersion)`
* Sort tasks by: version and priority
* Iterate through filtered tasks and execute the `Run()` method
* If any of the tasks return a result with `Stop` flag set to true, the
  update process is terminated

A few fake update task implementations have been defined under
`pkg/update/{all,v0.7.0,v0.8.0,v1.3.0}` and a binary demonstrating
how to use them is defined at `main/main.go`.

Feel free to evaluate, try it and share some feedback.

### DIFF
--- a/main/kube/all/deployments.go
+++ b/main/kube/all/deployments.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 
 	"github.com/skupperproject/skupper/api/types"
+	"github.com/skupperproject/skupper/main/kube"
 	"github.com/skupperproject/skupper/pkg/update"
-	"github.com/skupperproject/skupper/pkg/update/shared"
 	"github.com/skupperproject/skupper/pkg/utils"
 )
 
-type UpdateDeployments struct{}
+type UpdateDeployments struct {
+	Common *kube.KubeTask
+}
 
 func (u *UpdateDeployments) Version() string {
 	return "*"
@@ -35,10 +37,10 @@ func (u *UpdateDeployments) Platforms() []types.Platform {
 }
 
 func (u *UpdateDeployments) Run() update.Result {
-	if shared.RestartController {
+	if u.Common.RestartController {
 		fmt.Println("Restarting controller")
 	}
-	if shared.RestartRouter {
+	if u.Common.RestartRouter {
 		fmt.Println("Restarting router")
 	}
 	return update.Result{}
--- a/main/kube/kubetask.go
+++ b/main/kube/kubetask.go
@@ -1,0 +1,14 @@
+package kube
+
+import (
+	"github.com/skupperproject/skupper/client"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+type KubeTask struct {
+	Cli               *client.VanClient
+	RestartRouter     bool
+	RestartController bool
+	Router            *appsv1.Deployment
+	Controller        *appsv1.Deployment
+}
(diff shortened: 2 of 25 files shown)


grs (commented):
I like the idea; the current approach is certainly not sustainable. I'm not too keen on the shared data though. Would be nicer to have that somehow passed in. Also its not clear how the tasks would get the context they need (kube client etc).

One other aspect to consider is that I believe we should be aiming to have the site reconfigure itself as necessary based on changes to the configuration. Once that is done, though it will I believe still be useful to have some sort of explicit upgrade mechanism, a lot of smaller, simpler aspects would be automatically handled.

fgiorgetti:
> I like the idea; the current approach is certainly not sustainable. I'm not too keen on the shared data though. Would be nicer to have that somehow passed in. Also its not clear how the tasks would get the context they need (kube client etc).

I have updated the sample, removed the shared data and added a more "realistic" usage for
kubernetes and podman clients.

Notice that for the "vault" task, I have prepared an abstract implementation under `pkg/update/v1_3_0` and
each platform provides their own implementation for the `Run()` method.

fgiorgetti:
> One other aspect to consider is that I believe we should be aiming to have the site reconfigure itself as necessary based on changes to the configuration. Once that is done, though it will I believe still be useful to have some sort of explicit upgrade mechanism, a lot of smaller, simpler aspects would be automatically handled.

Do you think it is worth doing it based on the existing skupper-site or should we focus on preparing a shared
model that could work across platforms (similar to the ansible inventory from skupper-ansible)? But anyway,
I definitely agree we need a common configuration strategy as well.

Issue #972, opened 2022-12-14
Skupper exposed services change ClusterIP constantly


muzcategui1106-gs:
There seems to be an issue with the skupper-router or the skupper-service controller in which ClusterIPs are constantly changing. This creates issues down the line as we have automation that depends on the ClusterIPs of the services. 

**Issue I am facing**
exposed services cluster IPs are constantly changing (every couple of hours or so)

**Kubernetes Distribution**
Openshift 4.10

**Versions**
Skupper-site-controller:  1.2.0
Skupper-router: 2.2.0
Skupper-service-controller: 1.2.0
config-sync: 1.2.0

**Environment**
Source cluster sits on internal network and destination cluster where the services are exposed is in AWS.  There is connectivity from the internal cluster to the AWS cluster through AWS direct connect and a transit VPC

**How to replicate**
Unfortunately, I am not able to replicate, it just happens sporadically

**What I think is happening**
By definition the ClusterIP can never change. This suggests to me that either the skupper-service-controller is somehow deleting the services and recreating it. The reason could be a connectivity issue in the link between Skupper routers. 

**Supporting information**
I am unable to post this at the moment\

Happy to hear thoughts on things that I can look at. I am happy to provide supporting information through more private channels.

grs:
Try `skupper debug events` and `kubectl get events` in the namespace the problem is occurring in.

How are you selecting the services in question for exposure with skupper? Using the cli? Or annotations (on deployment or service resource)?

muzcategui1106-gs:
the services are being exposed from an internal cluster to the AWS cluster using the annotation on the service

```
skupper.io/proxy=tcp
```

Please see detailed events at the end of this message. Regardless where the problem lies (network or skupper router), It is always safe to assume that something will happen and the links will get temporarily interrupted; eg; networking blips.  What I propose to mitigate this is that there is a configurable option where users can specify a grace period in which the service-controller wont delete services, should the skupper links get interrupted.  This will ensure that even when there are networking blips, issues in the cluster wont propagate longer than the blip due to Services being recreated which could cause issues for other services depending on those exported services.

here are events from both commands you have given me . Please note I have removed names of specific services and replaced with XXX
```
 ./skupper --namespace skupper-services debug events
NAME                         COUNT                                                                                                                                                                                  AGE
ServiceControllerEvent       127                                                                                                                                                                                    14m6s
                             5     service event for skupper-services/XX                                                                                                                             14m6s
[...]

PR #975, opened 2022-12-19
Site ttl

grs:
Make time service-sync waits to expire stale services configurable and increase the default. Should offer resolution for #972

### DIFF
--- a/pkg/service_sync/service_sync.go
+++ b/pkg/service_sync/service_sync.go
@@ -19,6 +19,7 @@ type UpdateHandler func(changed []types.ServiceInterface, deleted []string, orig
 type ServiceSync struct {
 	origin            string
 	version           string
+	ttl               time.Duration
 	handler           UpdateHandler
 	connectionFactory messaging.ConnectionFactory
 	outgoing          chan ServiceUpdate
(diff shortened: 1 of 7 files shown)


pkg/service_sync/service_sync.go
```diff
@@ -36,10 +37,18 @@ type ServiceUpdate struct {
 	definitions map[string]types.ServiceInterface
 }
 
-func NewServiceSync(origin string, version string, connectionFactory messaging.ConnectionFactory, handler UpdateHandler) *ServiceSync {
+func getTtl(ttl time.Duration) time.Duration {
```
fgiorgetti, line 40: Do you think it is useful to enforce a minimum and safe value for the TTL,
to prevent it from struggling with the default 5s regular updates?

grs: Yes, that would make sense. What do you think the minimum should be?

fgiorgetti: Good question. Probably not less than a minute?

grs: Done

ajssmith (approved):
lgtm

Issue #982, opened 2023-01-03
you cannot update a site using a build of skupper


pwright:
If you build the skupper cmd from source, a hash is used instead of a version.

If you try to update a site, you get a message like:

```
$ skupper update
Error: Site (1.2.0) is newer than library (0fe3e54); cannot update
```

ssorj:
A scenario for this: I want to test upgrade from the current released version to the latest nightly.

grs:
The version is deduced from the last tag on the branch in use. We tag releases on the release branches, so master has no tags. If the version is not numeric, we can't determine its relation to a specified numeric version.

You can override the default version by compiling with `make VERSION=<desired-string>`.

PR #983, opened 2023-01-04
Add instructions for building skupper cli

pwright:
addresses #982


grs (approved):
One thing to perhaps point out is that the images used are not built with this. Therefore any changes made to the code those images are built from will not be in use.

Issue #971, opened 2022-12-14
Exposed services should be synced with labels and annotations


muzcategui1106:
Skupper exposed services do not have the source service annotations and labels synced. Annotations and labels usually convey important information about objects that other components running in the cluster may depend on. 

Proposal: Have the source service annotations and labels synced to the exposed services in target clusters

I understand this may not be a straight forward task as it may seem. One of the problems that I see is race conditions where multiple skupper instances exposed the same services but have defined different annotations keys or values.

To handle disparity in annotation/label keys, the target service can contain a combination of all the source annotations and labels keys. When it comes to the values, this may need some more in depth thinking.

fgiorgetti:
I am not sure if that helps you, but if you are exposing a service, a deployment or daemonset, you could add the
`skupper.io/service-labels` annotation to it, with a comma separated list of 'label=value', like: 

```
    skupper.io/service-labels: label1=value1,label2=value2
```

But Skupper does not yet have the same for annotations.

If you are using the CLI you could use `skupper service label` to manage the shared labels.

muzcategui1106:
yes confirmed that the service label work. Sadly due to label size limitations, what would really come useful here is the annotation sync

fgiorgetti:
I believe it can be done similarly to the way that Skupper deals with labels (via annotations and CLI).
We just need to ensure that the skupper annotations are not replicated and we should be good.
@grs @ajssmith @nluaces @muzcategui1106 What do you all think?

grs:
Sounds good to me.

muzcategui1106:
Another thing that I found out, is that the label sync just works when I create the service, if I update the annotation to either add a new label or update the value of one of the labels, it does not seem to sync

muzcategui1106:
Just following up on the above comment, were you able to replicate the issue were label sync does not work when the service is updated? it seems to only work when the service is created

grs:
> Just following up on the above comment, were you able to replicate the issue were label sync does not work when the service is updated? 

Yes, and that should now be fixed by https://github.com/skupperproject/skupper/commit/24b1c888799f2fc52c1e6fd616441152e25940cf

grs:
[...]

PR #985, opened 2023-01-05
Service annotations

grs:


grs:
This follows the approach used by labels, which has the downside that it restricts the value of annotations that can be used (cannot contain ```,``` or ```=```). An alternative would be instead to use some prefix on the annotation keys, that would then be stripped off.

grs:
Relates to #971

nluaces (approved):

ajssmith (approved):
lgtm, tried locally and tests

PR #996, opened 2023-01-24
bump version

mathianasj:
This is to get the newer k8s client apis to have the option of not exposing node ports when a load balancer service type is selected for the skupper and skupper-controller services.


nluaces:
Closing this pull request as it has already been merged in https://github.com/skupperproject/skupper/pull/1046

PR #997, opened 2023-01-24
fix attr check

ajssmith:


grs (approved):

pkg/flow/encoding.go
```diff
@@ -200,8 +201,11 @@ func decode(msg *amqp.Message) []interface{} {
 				m := make(map[string]interface{})
 				if r, ok := record.(map[interface{}]interface{}); ok {
 					for k, v := range r {
-						// to do assert k
-						m[attributeNames[k.(uint32)]] = v
+						if k.(uint32) < uint32(len(attributeNames)) {
+							m[attributeNames[k.(uint32)]] = v
+						} else {
+							log.Println("Attribute out of range for record conversion: ", k.(uint32))
```
hash-d, line 207: As part of vflow, I agree that an unknown attribute type should not cause failures.  Though I'm guessing it might cause problems down the line, if we decode a message only partially.

In any case, it might be good to add something that would make that log line stand out (such as the words 'warning' or 'unexpected').  The panic that was previously happening helped in the sense that it brought attention to the fact that the enums were out of sync.

For this specific case, the problem was on router and service controller that were in theory 'same version'.  In production this would most probably happen for mixed version deployments, or during VAN upgrades?

Also, perhaps adding the word 'flow' to the log would help during investigations; both when pointing to part of the code and when letting the research ignore a problem on a different part of the product ("I don't need to investigate this message in the log, as it's flow specific and my problem is elsewhere")

fgiorgetti (approved):

pkg/flow/encoding.go
ajssmith: Good suggestion. I will update the log message. As the log is associated with the collector container, it is in effect a flow specific message.

PR #999, opened 2023-01-25
add options to support unusual linking

grs:
The `init` command has a hidden ``--router-disable-mutual-tls` option. The `token create` command has a hidden `template` option which if used, should refer to a secret whose contents will be used to construct the token (adding necessary annotations etc). There are also some internal changes that will allow the use of a token with only the ca.crt, and no private key or client cert.


fgiorgetti (approved):
LGTM!

Listener seems to be configured correctly and the token info looks correct as well.

nluaces (approved):

ajssmith (approved):
lgtm2

PR #1018, opened 2023-02-08
Fixed import issue

fgiorgetti:


grs (approved):
Ideally v1 would be qualified (e.g. openshiftappsv1 or oappsv1 or something), but this at least fixes the broken build.

nluaces (approved):

PR #1020, opened 2023-02-09
restore access to console api

grs:
Also adds /sites path to the public api.

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -92,6 +92,7 @@ type SiteConfigSpec struct {
 	SiteTtl             time.Duration
 	EnableConsole       bool
 	EnableFlowCollector bool
+	EnableRestAPI       bool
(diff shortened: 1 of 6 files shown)


ajssmith (approved):
lgtm

nluaces (approved):

fgiorgetti:
If you initialize Skupper on an OpenShift cluster, there is no `route` (or ingress) exposing port 8080.
@grs Do you think it should be exposed externally as well?

fgiorgetti (approved):

grs:
> If you initialize Skupper on an OpenShift cluster, there is no `route` (or ingress) exposing port 8080. @grs Do you think it should be exposed externally as well?

At this point I don't see a need. If needed that can always be done by the user with an edge terminated ingress or a route.

PR #1030, opened 2023-02-14
ADR-0003 Skupper console provides data first

pwright:
@ajssmith I'm thinking something like this would help with the 1.3 release.. 
The idea being to add to release notes:

The new Skupper console does not support linking sites as per [link to this ADR]


pwright:
I believe sometime in the future, we'll be able to link sites from a console (maybe not the skupper console), but regardless, closing this as obsolete

PR #1040, opened 2023-02-23
Adding podman v4 to CircleCI and disabling GHA

fgiorgetti:


fgiorgetti:
With this I am going to stop using GHA for now and get Podman tests to run as part of the CircleCI as well.
Along with this change, I am also updating the executor image used in CircleCI so that we use latest Ubuntu available.

grs (approved):

PR #1044, opened 2023-02-28
CI jobs incorrectly reported as failed

fgiorgetti:


fgiorgetti:
Succeed and CompletionTime are not guaranteed to be written in the expected order.
The original validation was assuming the fact that Succeed is not positive, then it failed,
which can be an incorrect interpretation.

This makes a strict validation that the job has actually failed or completed without errors.

grs (approved):

PR #1047, opened 2023-03-06
Ensure podman endpoint does not allow local and unspecified addresses

fgiorgetti:


fgiorgetti:
This change will be important for upcoming changes to podman sites where we will have 
the flow-collector and the service-controller implementations and both will need to access
the podman endpoint.

grs (approved):

PR #1055, opened 2023-03-10
Draft: Debugging basic test failure

fgiorgetti:


fgiorgetti:
The basic test (impacting TestBasic/test-sync-containers) was being impacted by this headless test
as both were writing to the same token file, and since tests run in parallel, sometimes it was causing
one test to link to the incorrect namespace causing the other to fail.

Now the headless test uses a unique token file and bypasses that CI issue.

nluaces (approved):

grs (approved):

PR #1065, opened 2023-03-20
Adjusting permissions needed by site controller

fgiorgetti:
Site controller is not working on OpenShift as it is trying to grant permissions not currently held.


nluaces (approved):
nice job! thanks!

PR #1068, opened 2023-03-22
Podman commands now verify if site exists before executing

fgiorgetti:
Some commands were not verifying if site exists before executing actions,
giving users inappropriate  error messages. Now this validation is done for
all podman commands in a standard way.

### DIFF
--- a/cmd/skupper/skupper_podman.go
+++ b/cmd/skupper/skupper_podman.go
@@ -18,12 +18,12 @@ var SkupperPodmanCommands = []string{
 }
 
 type SkupperPodman struct {
-	cli     *clientpodman.PodmanRestClient
-	curSite *podman.Site
-	site    *SkupperPodmanSite
-	token   *SkupperPodmanToken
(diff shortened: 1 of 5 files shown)


ajssmith (approved):
lgtm, built and tried locally

grs (approved):

cmd/skupper/skupper_podman.go
```diff
@@ -19,6 +19,7 @@ var SkupperPodmanCommands = []string{
 
 type SkupperPodman struct {
 	cli     *clientpodman.PodmanRestClient
+	curSite *podman.Site
```
nluaces, line 22: Does this new attribute stand for "current site"? In that the case my preference would be `currentSite` because I find it more clear to me, but this is not a blocker.

nluaces (approved):

cmd/skupper/skupper_podman.go
fgiorgetti: True. I will update it.

Issue #865, opened 2022-09-01
How to expose multiple deployments to a single skupper service in a declarative way


rmallam:
Hi,

I am looking to expose multiple deployment resources with a single skupper service. I can do this using the CLI but i want to do this in a declarative way. I tried passing the same skupper service name in annotations for both the deployments , but the one which gets created the last is the only one available as target. 

Example: 
- Deploy a simple httpd webserver with the name httpd and again with name httpd-1
- skupper expose deployment httpd --address skupperhttpd
- skupper expose deployment httpd-1 --address skupperhttpd

This will create a skupper service called skupperhttpd with both the deployments listed in the targets as shown below.

$skupper service status
Services exposed through Skupper:
├─ skupperhttpd (tcp port 8080)
│  ╰─ Targets:
│     ├─ application=httpd name=httpd
│     ╰─ application=httpd-1 name=httpd-1

but if try the annotations to achieve the same, The last created deployment will be the only target and the previous deployments are deleted from targets.

below are my annotations passed.

  annotations:
    skupper.io/address: "skupperhttpd"
    skupper.io/port: "8080"
    skupper.io/proxy: "tcp"

rmallam:
@grs thanks for updating this, i tested the latest release with multiple ports and multiple deployments,  i can see service controller trying to update the service and was not able to fullfill the requests.

grs:
Is the resourceVersion on the service changing continually?

rmallam:
@grs looks like this is because of a different problem, Customer was running 2 replicas of service controller and i could see from their logs that both of them are trying to update the resource definition and are failing to do so causing this error. I will test this again properly and raise a new issue.

PR #1077, opened 2023-03-30
fix the test for port equivalence on service definition updates

grs:


nluaces (approved):

cmd/service-controller/definition_monitor.go
```diff
@@ -197,9 +197,17 @@ func updateAnnotatedServiceDefinition(actual *types.ServiceInterface, desired *t
 	if actual.Origin != "annotation" {
 		return false
 	}
-	if actual.Protocol != desired.Protocol || !reflect.DeepEqual(actual.Ports, desired.Ports) {
```
fgiorgetti, line 200: This seems to have the same effect to me. The new test also passes without changing it.

What seems to be wrong here for me is the fact that, if any of the if statements in the
beginning of the function returns true, the final section of this function, which actually
updates the targets in the desired argument, is not being executed, therefore if ports
AND targets have changed, the desired targets are not being evaluated.

grs: You are right, the new test does not change the order of the ports and the new test logic incorrectly assumes the ports are in a map, rather than a list. Both fixed now.

cmd/service-controller/definition_monitor_test.go
```diff
@@ -1133,6 +1133,63 @@ func TestUpdateAnnotatedServiceDefinition(t *testing.T) {
...
+							8081: 8081,
+							8082: 8082,
+						},
+					},
+				},
+			},
+			desired: types.ServiceInterface{
+				Address:  "five",
+				Protocol: "tcp",
+				Ports:    []int{8080, 8082, 8081},
```
fgiorgetti, line 1158: Try applying this patch to the `definition_monitor_test.go` and you will see the same behavior
reported through #865.

```
diff --git a/cmd/service-controller/definition_monitor_test.go b/cmd/service-controller/definition_monitor_test.go
index ca256ef7..f003a148 100644
--- a/cmd/service-controller/definition_monitor_test.go
+++ b/cmd/service-controller/definition_monitor_test.go
@@ -1155,7 +1155,7 @@ func TestUpdateAnnotatedServiceDefinition(t *testing.T) {
 			desired: types.ServiceInterface{
 				Address:  "five",
 				Protocol: "tcp",
-				Ports:    []int{8080, 8081, 8082},
+				Ports:    []int{8080, 8081, 8082, 8083},
 				Targets: []types.ServiceInterfaceTarget{
 					{
 						Name:     "t2",
@@ -1164,6 +1164,7 @@ func TestUpdateAnnotatedServiceDefinition(t *testing.T) {
 							8080: 8080,
 							8081: 8081,
 							8082: 8082,
+							8083: 8083,
 						},
 					},
 				},
@@ -1186,6 +1187,7 @@ func TestUpdateAnnotatedServiceDefinition(t *testing.T) {
 						8080: 8080,
 						8081: 8081,
 						8082: 8082,
+						8083: 8083,
 					},
 				},
 			},
```

grs: If the ports are actually different, then the two definitions are not compatible and the later one is used as a replacement. If they are compatible, then the targets are merged.

So in the case where you have four ports defined instead of three, then the expected behaviour is to take the second definition as a replacement for the first i.e. the first target will be lost as it is based on an definition deemed incompatible. Likewise if the protocol or some other aspect of the service is different. Only the targets for compatible definitions can be merged.

fgiorgetti: That makes sense! Thank you.

fgiorgetti (approved):

cmd/service-controller/definition_monitor.go
```diff
@@ -197,9 +197,21 @@ func updateAnnotatedServiceDefinition(actual *types.ServiceInterface, desired *t
...
+	if len(actual.Ports) != len(desired.Ports) {
+		return true
+	}
+	ports := map[int]int{}
+	for _, value := range desired.Ports {
+		ports[value] = value
+	}
+	for _, value := range actual.Ports {
+		if ports[value] != value {
+			return false
```
fgiorgetti, line 212: Is it supposed to be true here?
I have looked into this again and I noticed that with false returned here, it prevents annotated services from changing ports.

grs:
@fgiorgetti thank you! You are of course correct. I also added a test case for changing port. Appreciate the sharp eyes and deep review, apologies for my many blunders in this one!

fgiorgetti (approved):

fgiorgetti:
> @fgiorgetti thank you! You are of course correct. I also added a test case for changing port. Appreciate the sharp eyes and deep review, apologies for my many blunders in this one!

Thanks for the patience.

PR #1079, opened 2023-03-31
add receiver for flow address

ajssmith:
required for:

https://github.com/skupperproject/skupper-router/pull/930

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,7 +208,7 @@ commands:
           no_output_timeout: 60m
           command: |
             make build-cmd && sudo install skupper /usr/local/bin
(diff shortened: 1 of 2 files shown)


nluaces (approved):
Looks good to me; just to confirm, the console is the one that will filter by the `.flows` suffix in the beacon address, right?

grs (approved):
Is there any issue with versioning here? e.g. if the local router is from before this change? or if there are older sites in the network?

ajssmith:
@grs for downrev routers, the flow records should arrive on the original address. iiuc, adding the new flow address should not impact collection.

fgiorgetti (approved):
Looks good to me.

PR #1080, opened 2023-03-31
Podman container informer

fgiorgetti:


client/podman/informer.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (e *EventHandlerBase[T]) OnDelete(obj T) {
+	e.Delete(obj)
+}
+
+//
+// Container informer
+//
+
+func NewContainerInformer(cli *PodmanRestClient) *ContainerInformer {
```
grs, line 62: Where does this get used?

fgiorgetti: It will be used as part of the service-controller, which will initially provide event sources to the flow-collector.

grs (approved):

client/podman/informer.go
```diff
@@ -0,0 +1,148 @@
...
+func (i *informerBase) SetResyncPeriod(t time.Duration) {
+	i.resyncPeriod = t
+}
+
+type updateNotification[T any] struct {
+	oldObj T
+	newObj T
+}
+
+type EventHandler[T any] interface {
```
nluaces, line 34: Just a detail regarding naming; the project has an interface called `EventHandlerInterface` that is implemented by components that generate events in skupper or in kubernetes with different methods to the `EventHandler` interface. What should we do in this case? using different names to express a difference in functionality (in this case methods)? I don't have a preference on which interface should be renamed.

fgiorgetti: It makes sense.

I believe it might be clearer to move this new one under `client/container` and rename it to `Informer`,
this way it would be referred to by other packages (like podman or an eventual docker) as `container.Informer`. 

Check it out one more time, please.

ajssmith (approved):

nluaces (approved):

PR #1081, opened 2023-04-03
Generate a Manifest file per release

### LINKED ISSUE #924: skupper containers image versions not available in the skupper releases page.
Hi,

We use a custom container registry to deploy skupper in our environments and having a manifest of all the image versions listed as part of the skupper releases page will be helpful. Given that skupper router version is different, we need to lookup the router version every time.

or 

if we could align the skupper-router container image version with skupper version, that will help to pull images easily from Quay and push them to our custom registry.

nluaces:
This change generates a `manifest.json` file with the images that Skupper uses by default.

This is an example of the content:
```
{
   "images": [
      {
         "Name": "quay.io/skupper/skupper-router:main",
         "Repository": "https://github.com/skupperproject/skupper-router"
      },
      {
         "Name": "quay.io/skupper/service-controller:master",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/config-sync:master",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/flow-collector:master",
         "Repository": "https://github.com/skupperproject/skupper"
      }
   ]
}

```
The file will be published along the artifacts in GitHub's release page.

Fixes #924

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -528,6 +528,8 @@ jobs:
                 tar -zcf "${BASEDIR}/archives/skupper-cli-${VERSION}-$p.tgz" *
               fi
             done
+            make -e generate-manifest
+            cp ${BASEDIR}/manifest.json ${BASEDIR}/archives
             cd ${BASEDIR}
             ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -replace -prerelease -draft ${VERSION} "${BASEDIR}/archives"
 
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 /get
 /config-sync
 /flow-collector
+/manifest
 
 # Test binary, built with `go test -c`
 *.test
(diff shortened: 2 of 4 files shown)


grs:
Is the `manifest` executable supposed to be checked in?

grs:
Normally, executables would be built before use, rather than checked in. Since this is mostly only going to be used when building releases under ci, that doesn't on the face of it look problematic. I.e. just ensure `make build-manifest` is called before generate-manifest (or even, by marking it as a dependency of generate-manifest, as a by product of generate-manifest).

fgiorgetti:
> Normally, executables would be built before use, rather than checked in. Since this is mostly only going to be used when building releases under ci, that doesn't on the face of it look problematic. I.e. just ensure `make build-manifest` is called before generate-manifest (or even, by marking it as a dependency of generate-manifest, as a by product of generate-manifest).

I agree. The `generate-manifest` target should depend on the `build-manifest` to enforce the presence of the `manifest`
binary file. Then the `manifest` binary should be removed from the PR and added into the `.gitignore`.

fgiorgetti (commented):
I would also include the `manifest.json` as part of the .gitignore.

grs (approved):

fgiorgetti (approved):

ajssmith (approved):

PR #1087, opened 2023-04-05
Ignoring skmanage responses from stale routers

fgiorgetti:


fgiorgetti:
This PR fixes an issue observed when you run `skupper status` right after a link is deleted.
Skupper still tries to run skmanage against a router node that is not longer available, causing the
`skupper status` command to show an unexpected error message.

pkg/domain/podman/router.go
```diff
@@ -70,8 +70,9 @@ func (r *RouterEntityManager) QueryConnections(routerId string, edge bool) ([]qd
 		return nil, fmt.Errorf("error querying connections - %w", err)
 	}
 	var connections []qdr.Connection
+	// if the given routerId no longer exists, it is expected to fail
 	err = json.Unmarshal([]byte(data), &connections)
-	if err != nil {
+	if err != nil && routerId == "" {
```
grs, line 75: Would it be clearer to just check that the routerId is specified as the first line of the function, failing with an error regarding the missing id if not?

fgiorgetti: In this function, the router id is optional. If you do not pass one, then the Unmarshal is expected to work.
If you pass one, it can eventually fail (as skmanage might not return a JSON).

Just to give some context. This function is called to help determine all routers in the VAN (interiors and edges),
so prior to calling this function, it queries the `router.node` entities (interiors), then it iterates through each of 
them, querying their connection entities (which include edges) providing the specific router id to skmanage.

There is a delay between the moment the link is removed and the moment you stop seeing the respective 
`router.node` entity returned by the $management, so although we are still seeing the node entity, it is no 
longer available and that's why it is expected to fail.

fgiorgetti: True stuff. It is really not the place to decide whether or not it is an expected error.
I have delegated this decision to the callers, which are passing a specific router id.

pkg/domain/podman/router.go
```diff
@@ -106,13 +106,17 @@ func (r *RouterEntityManager) QueryAllRouters() ([]qdr.Router, error) {
...
+				continue
+			}
 			return nil, fmt.Errorf("error decoding router info from %s - %w", routerToQuery, err)
 		}
 		router.Site = qdr.GetSiteMetadata(records[0].AsString("metadata"))
 
 		// retrieving connections
 		conns, err := r.QueryConnections(routerToQuery, router.Edge)
-		if err != nil {
+		if err != nil && !strings.Contains(err.Error(), "invalid character") {
```
grs, line 119: It is not at all obvious why 'invalid character' is 'expected' here. The issue I think is that any unusual output from the exec will fail to be parsed as json, most likely with this same error. One option that I think would at least be clearer is just to ignore the error here. Alternatively, to skip only expected errors, the QueryConnections() call could include the exec output in the error it returns, That way you can query for the actual expected error here (e.g. no such router or whatever the skmanage command reports).

grs: My inclination would be to return the output from exec in the error if failing to unmarshal, and in this specific case check for 'SendException: RELEASED' in the error and ignore it if so (ideally with a comment explaining that this most likely means the routerId did not exist anymore).

Ignoring the error entirely is not terrible either though.

pkg/domain/podman/router.go
```diff
@@ -106,13 +111,17 @@ func (r *RouterEntityManager) QueryAllRouters() ([]qdr.Router, error) {
...
+				continue
+			}
 			return nil, fmt.Errorf("error decoding router info from %s - %w", routerToQuery, err)
 		}
 		router.Site = qdr.GetSiteMetadata(records[0].AsString("metadata"))
 
 		// retrieving connections
 		conns, err := r.QueryConnections(routerToQuery, router.Edge)
-		if err != nil {
+		if err != nil && err.(*json.SyntaxError) == nil {
```
grs, line 124: Why is this extra check needed? Isn't the special case handled already in QueryConnections()?

pkg/domain/podman/router.go
```diff
@@ -106,13 +111,17 @@ func (r *RouterEntityManager) QueryAllRouters() ([]qdr.Router, error) {
 		var records []qdr.Record
 		err = json.Unmarshal([]byte(rJson), &records)
 		if err != nil {
+			// routerToQuery no longer exists
+			if routerToQuery != "" {
```
grs, line 115: I don't follow the logic here still. When would the router id be ""? And why does the error to be returned include the router id, if it will only be returned if the id is ""?

pkg/domain/podman/router.go
fgiorgetti: This is no longer needed and it has been removed now.

pkg/domain/podman/router.go
fgiorgetti: The logic is not valid. I have updated it to ignore both SendException and Timeout messages.

grs (approved):
Looks good. (Minor suggestion would be to have a function for the strings checks to  avoid duplication and perhaps through the name explain the context).

grs (approved):
:shipit:

PR #1092, opened 2023-04-14
provide error for flag check

ajssmith:


nluaces (approved):

cmd/skupper/skupper_kube_site.go
```diff
@@ -71,6 +71,10 @@ func (s *SkupperKubeSite) Create(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("The minimum value for service-sync-site-ttl is 1 minute")
 	}
 
+	if routerCreateOpts.EnableConsole && !routerCreateOpts.EnableFlowCollector {
```
fgiorgetti, line 74: This whole stuff could be moved to NewCmdInit's function under `cmd/skupper/skupper.go`,
right before the `skupperCli.Create` is called, as it is used by both podman and kube sites.

ajssmith: moved, thanks

PR #1103, opened 2023-05-10
make incoming links optional in the link status command

nluaces:
- New flag `show-remote` to include remote link information in the `link status` command.
- Default timeout for the remote links changed from 2 minutes to 5 seconds.

### DIFF
--- a/cmd/skupper/skupper_link.go
+++ b/cmd/skupper/skupper_link.go
@@ -81,6 +81,7 @@ func NewCmdLinkDelete(skupperClient SkupperLinkClient) *cobra.Command {
 var waitFor int
 var remoteInfoTimeout time.Duration
 var verboseLinkStatus bool
+var showRemoteLinks bool
 
 func allConnected(links []types.LinkStatus) bool {
 	for _, l := range links {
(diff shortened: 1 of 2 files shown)


cmd/skupper/skupper_link.go
```diff
@@ -217,8 +221,9 @@ func NewCmdLinkStatus(skupperClient SkupperLinkClient) *cobra.Command {
 		},
 	}
 	cmd.Flags().IntVar(&waitFor, "wait", 0, "The number of seconds to wait for links to become connected")
-	cmd.Flags().DurationVar(&remoteInfoTimeout, "timeout", types.DefaultTimeoutDuration, "Configurable timeout for retrieving information about remote links")
 	cmd.Flags().BoolVar(&verboseLinkStatus, "verbose", false, "Show detailed information about a link")
+	cmd.Flags().DurationVar(&remoteInfoTimeout, "timeout", time.Second*5, "Configurable timeout for retrieving information about remote links")
+	cmd.Flags().BoolVar(&showRemoteLinks, "show-remote", false, "Show incoming links from other sites")
```
grs, line 226: Could we just use the verbose option to enable showing the incoming links? Else show-incoming-links would be clearer.

nluaces: Currently, the verbose option shows the details of a link (cost, etc). I can change it to show incoming links as a part of the verbose mode when listing all the links, but I also agree with you that `show-incoming-links` flag is clearer.

ajssmith (approved):

PR #1105, opened 2023-05-11
Fix handling of console ingress option for routes

### LINKED ISSUE #1104: Cannot use route as ingress mode for console when main ingress mode is something else.
Options `--ingress=loadbalancer --console-ingress=route --enable-console --enable-flow-collector` do not result in console being exposed via a router as would be expected.

grs:
Specifically fix case when console-ingress option to be set to route when ingress is not option. Fixes #1104

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -29,6 +29,12 @@ import (
 	"github.com/skupperproject/skupper/pkg/utils"
 )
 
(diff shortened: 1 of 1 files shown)


fgiorgetti (approved):
Seems to work just fine on OCP using ingress lb and console-ingress route.

nluaces (approved):

PR #1107, opened 2023-05-11
Controller and collector beacons for podman sites

fgiorgetti:


cmd/service-controller/main.go
```diff
@@ -61,32 +62,43 @@ func main() {
...
-	log.Println("Waiting for Skupper router component to start")
-	_, err = kube.WaitDeploymentReady(types.TransportDeploymentName, namespace, cli.KubeClient, time.Second*180, time.Second*5)
-	if err != nil {
-		log.Fatal("Error waiting for transport deployment to be ready: ", err.Error())
-	}
-
-	// start the controller workers
-	if err = controller.Run(stopCh); err != nil {
-		log.Fatal("Error running controller: ", err.Error())
+	platform := config.GetPlatform()
```
grs, line 70: I wonder if it might be simpler here to have a separate main.go for the podman version, and therefore a separate image?

cmd/service-controller/controller.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	clientpodman "github.com/skupperproject/skupper/client/podman"
```
grs, line 13: I wonder if it might also be clearer to keep these in a distinct file. The current controller does define far too much stuff in the main package. We should move stuff into explicit packages if needed.

pkg/domain/podman/common.go
```diff
@@ -18,7 +18,7 @@ const (
 var (
 	Username                = readUsername()
 	SkupperContainerVolumes = []string{"skupper-services", "skupper-local-server", "skupper-internal", "skupper-site-server", SharedTlsCertificates,
-		types.ConsoleServerSecret, types.ConsoleUsersSecret}
+		types.ConsoleServerSecret, types.ConsoleUsersSecret, types.ClaimsServerSecret}
```
grs, line 21: Is this perhaps a distinct change to adding collector beacons?

cmd/service-controller/main.go
fgiorgetti: That is a good question.
I would say yes, it would be simpler and probably safer, especially for the kubernetes controller.
The only painpoint, or maybe just something to consider, is that it introduces an extra image to be maintained.

Suppose we split it out as a separate process/image, what would be your preference to place it?
* cmd/service-controller-podman, or,
* cmd/podman/service-controller ?   <-- this would be my choice

Considering we change it to be a separate process/image, I believe we should do the same to
the flow-collector process/image. What do you think @grs @ajssmith @nluaces ?

cmd/service-controller/controller.go
fgiorgetti: If we keep a separate package/image to the podman controller, the new `ControllerPodman` could be placed
there and the `collector_beacons.go` could be moved under `pkg/domain/podman` as already suggested by
@ajssmith.

I will separate it as a new process and image, then we can reevaluate it.

pkg/domain/podman/common.go
fgiorgetti: This is just a reference to the podman volumes that will be created during initialization.
It is also used as a pre-step to the initialization, to ensure skupper is able to create everything it needs beforehand.

cmd/service-controller/main.go
fgiorgetti: @grs could you please review this alternative PR, which introduces a new image ( #1120 )?

fgiorgetti:
Closing this one in favor of the separated images approach.

PR #1110, opened 2023-05-16
Fix for ocp-only bug due to runAsUser out of allowed ID range

hash-d:
"spec.containers[0].securityContext.runAsUser: Invalid value: 1000: must be in the ranges:"


fgiorgetti (approved):
LGTM.

PR #1111, opened 2023-05-17
Cherry pick ocp user limit test bug fix

hash-d:
cherry pick for https://github.com/skupperproject/skupper/pull/1110 into 1.4


hash-d:
@ajssmith, @nluaces, I'd appreciate if you could consider the inclusion of this cherry-pick into 1.4.  It fixes an OCP-only test bug, where Hello World would not work under OpenShift due to the use of RunAsUser with a value outside of the range allowed by OCP for the namespace.  It's already merged on master, from where this cherry pick originates.

PR #1112, opened 2023-05-17
purge and log records

ajssmith:
also emulate external processes

### DIFF
--- a/pkg/flow/flow_mem_driver.go
+++ b/pkg/flow/flow_mem_driver.go
@@ -1984,7 +1984,7 @@ func (fc *FlowCollector) reconcileRecords() error {
 						}
 						delete(fc.flowsToProcessReconcile, flowId)
 					} else {
-						if fc.needForExternalProcess(flow, siteId, connector.StartTime, false) {
+						if fc.needForSiteProcess(flow, siteId, connector.StartTime, false) {
 							delete(fc.flowsToProcessReconcile, flowId)
 						}
 					}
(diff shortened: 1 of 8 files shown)


grs (approved):

pkg/flow/flow_mem_driver.go
```diff
@@ -2089,6 +2179,15 @@ func (fc *FlowCollector) reconcileRecords() error {
 		}
 	}
 
+	for _, source := range fc.eventSources {
+		t := time.Now()
+		diff := uint64(t.UnixNano())/uint64(time.Microsecond) - source.EventSourceRecord.LastHeard
+		if diff > 30*oneSecond {
```
fgiorgetti, line 2185: How do you feel about 30 seconds? Do you think using the same value defined
through `--flow-collector-record-ttl` is a bit too much?

fgiorgetti (approved):
LGTM.

nluaces (approved):

pkg/flow/flow_mem_driver.go
ajssmith: record-ttl is 15 minutes default and a number of users have commented that they expect what is rendered in the console to be "real time". For example, if a site is deleted, they expect it to be removed soon after. It seems like anything more than 30-60 seconds would lose the real time feel. Additional user response should help with tuning this behavior.

fgiorgetti: Fine. I believe it is ok to go with it now and address later based on users feedback.

PR #1120, opened 2023-05-23
Podman service controller new image

fgiorgetti:


pkg/domain/podman/credential_handler.go
```diff
@@ -16,6 +17,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var credentialMountInContainer = map[string]string{
+	types.ClaimsServerSecret:  "/etc/service-controller/certs/",
```
grs, line 21: Why do we have a separate cert for claims and site-server in podman?

fgiorgetti: The claims server cert is not yet being used. Better to remove it and just add when claims become
supported?

fgiorgetti:
> Is this a service-controller? Or is it really a flow-collector for podman sites?

At this point it only runs the flow controller piece that provides Processes and Host records
to the flow-collector, but eventually more features will be added to this controller.

So maybe renaming it to skupper-podman-controller would make more sense.

Any other thought on this?

PR #1123, opened 2023-05-23
move claim redemption behind router service

grs:
This avoids confusion in configuring ingress. The changes also include some refactoring, moving code out of the client package and reducing duplication.


client/router_update.go
```diff
@@ -450,6 +456,35 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
...
+					return false, err
+				}
+			}
+		}
+	}
+	if moveClaims {
+		if err := cli.moveClaimsToRouterService(ctx, namespace); err != nil {
+			return false, err
+		}
+		latest, err := cli.KubeClient.AppsV1().Deployments(namespace).Get(context.TODO(), types.TransportDeploymentName, metav1.GetOptions{})
```
fgiorgetti, line 477: One thing I have noticed, while testing the update process, is that after the `regenerateSiteSecret`
completes (called by `moveClaimsToRouterService`), and the router is restarted, the
`skupper-router` deployment gets updated a few times, due to the statuses transitions, so the `latest`
resource obtained here, is not actually the latest, so I faced errors resulting from updates to a modified
skupper-router deployment quite a few times.

I tried adding a `WaitDeploymentReady` after `regenerateSiteSecret` is called and it seems to resolve it.

pkg/site/site.go
```diff
@@ -0,0 +1,829 @@
...
+	if ingress, ok := siteConfig.Data[SiteConfigIngressKey]; ok {
+		result.Spec.Ingress = ingress
+	} else {
+		// check for deprecated key
+		if clusterLocal, ok := siteConfig.Data["cluster-local"]; ok {
+			if clusterLocal == "true" {
+				result.Spec.Ingress = types.IngressNoneString
+			} else {
+				result.Spec.Ingress = types.IngressLoadBalancerString
+			}
```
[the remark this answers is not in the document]
grs: Agreed; done.

client/router_update.go
grs: Good suggestion, I have made that change.

pkg/kube/claims/claim_create.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+type ClaimOptions struct {
+	Name     string
+	Password []byte
+	Expiry   time.Duration
+	Uses     int
+}
+
+type SiteContext interface {
```
ajssmith, line 26: Would it make sense to move this to pkg/kube/site/context.go?

ajssmith (approved):
lgtm. 

Tried locally with revised images. Checked things such as can't connect to self, network status, etc. Note, did not test update.

pkg/kube/claims/claim_create.go
grs: Maybe ultimately. At present I see it as establishing the interface to an external context that this claims package requires (allowing the implementation to be decoupled from it). Different packages may have slightly different interfaces that need to be fulfilled. I do expect that there will be something that satisfies all of them. For now though, my view is that this interface is part of the claims package.

client/router_update.go
```diff
@@ -455,6 +461,35 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
...
+		}
+		latest, err := kube.WaitDeploymentReady(types.TransportDeploymentName, namespace, cli.KubeClient, time.Minute, time.Second)
+		if err != nil {
+			return false, err
+		}
+		if latest.ObjectMeta.ResourceVersion != router.ObjectMeta.ResourceVersion {
+			latest.Spec = router.Spec
+			router = latest
+		}
+		kube.AppendSecretVolumeWithVolumeName(&router.Spec.Template.Spec.Volumes, &router.Spec.Template.Spec.Containers[1].VolumeMounts, types.SiteServerSecret, "claims-cert", "/etc/skupper-internal/")
```
fgiorgetti, line 490: There is one problem here. If, let's say we are updating from 1.4.2 to 1.4.3, this will work fine.
If we need to update from 1.4.3 to 1.4.4, the secret will be appended twice and will cause the
router update to fail.

grs: This code will only run when updating to 1.5.x, and only when the original version is less than 1.5.0. So e.g. if you were at 1.4.2 and used the 1.5.0 client t update, the site would then be 1.5.0. If you then updated to 1.5.1 then the moveClaims flag would be false so you wouldn't append again.

Or am I missing the sequence that would cause the problem? (It would be fairly easy to fix anyway, I just don't yet see how that line would run twice).

fgiorgetti: Alright. My bad. I thought that, in spite of it being clear that applies only to sites using less than 1.5.0,
it could be part of an eventual upcoming patch release. So if this is just going to land as part of 1.5.0,
then it is working just fine.
Sorry for the noises and confusion.

fgiorgetti (approved):

PR #1126, opened 2023-05-25
Podman controller

fgiorgetti:


grs (approved):

pkg/images/images.go
```diff
@@ -4,6 +4,7 @@ const (
 	DefaultImageRegistry       string = "quay.io/skupper"
 	RouterImageName            string = "skupper-router:main"
 	ServiceControllerImageName string = "service-controller:master"
+	ControllerPodmanImageName  string = "controller-podman:master"
```
nluaces, line 7: should we add it into the version command output?

nluaces: I take that back, it is probably better to add it when the update command implementation for podman is ready.

nluaces:
Looks good to me! I tested it on local and I could see my podman site on the console and its services.

nluaces (approved):

ajssmith (approved):
lgtm - parent site succesfully linked

PR #1132, opened 2023-05-31
add state query param

ajssmith:


pkg/flow/record.go
```diff
@@ -499,6 +501,18 @@ func getQueryParams(url *url.URL) QueryParams {
 			default:
 				qp.TimeRangeOperation = intersects
 			}
+		case "state":
+			recordState := v[0]
+			switch recordState {
+			case "all":
```
nluaces, line 507: We could remove the `all` case and rely on the default case: 

```
switch recordState {
			case "active":
				qp.State = active
			case "terminated":
				qp.State = terminated
			default:
				qp.State = all
			}
```

PR #1135, opened 2023-06-05
Using link cost from token if not provided via CLI

### LINKED ISSUE #1109: Cost annotation in a token is ignored by CLI by used by the controller
If a token is generated with the `skupper.io/cost` annotation then used with the CLI, 
the cost value is being ignored, while it is considered when processed by the controller.

fgiorgetti:
Fixes #1109

### DIFF
--- a/api/types/client.go
+++ b/api/types/client.go
@@ -17,7 +17,7 @@ type ConnectorCreateOptions struct {
 	SkupperNamespace string
 	Name             string
 	Cost             int32
-	Yaml             []byte
+	Secret           *corev1.Secret
 }
 
 type ConnectorRemoveOptions struct {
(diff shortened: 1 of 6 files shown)


fgiorgetti:
This PR uses the `skupper.io/cost` annotation of a token, if one provided, only if `--cost` is not
used during `link create`. If `--cost` is passed through `link create`, the cost defined in token is ignored.

grs (approved):
It is not entirely obvious to me that this is what users would (or should) expect when using the cli (or the rest api for that matter). I have no objections if others feel it is though.

nluaces (approved):

fgiorgetti:
@ajssmith thoughts ?

ajssmith (approved):

PR #1140, opened 2023-06-14
prom query via api

ajssmith:


grs:
I may well be doing something wrong, but though the console comes up ok and doesn't give any errors, it doesn't show any metrics (I have a tcp exposed service over which I have made many http calls).

Also, oddly, I connected three sites, all on one cluster. Two were created using the code from this PR, one from master. The console only shows the two sites created using code from this PR. Though the cli status shows the other site connected, and it is routing connections as expected, it does not show up in the console.

As I say, perhaps I am doing something wrong in my testing here though.

ajssmith:
With this change, an update to the console implementation is needed. Not sure that it would explain the behaviors you observed and will investigate.

cmd/flow-collector/handlers.go
```diff
@@ -166,3 +168,45 @@ func (c *Controller) collectorHandler(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "%s", *response.Body)
 	}
 }
+
+func (c *Controller) promqueryHandler(w http.ResponseWriter, r *http.Request) {
+	client := http.Client{}
+
+	urlOut := c.FlowCollector.Collector.PrometheusUrl + "query?" + r.URL.RawQuery
+	proxyReq, err := http.NewRequest(r.Method, urlOut, nil)
```
fgiorgetti, line 176: Don't you need to pass the r.Body and r.Header to the proxyReq?
Without it, POST data won't be passed through or am I missing something?

ajssmith: Presently, the console does not POST to the query api

nluaces (approved):

grs:
On the site not showing up, I tried with 1.4.0-rc3 and that site does show up ok. I'm also now seeing some metrics; not sure what happened to change that. Perhaps I was too impatient. Seems to be working ok now for me.

grs (approved):

PR #1145, opened 2023-06-21
modified error message

nluaces:
Modified error message when trying to expose a statefulset as a headless service.

### DIFF
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -252,7 +252,7 @@ func (cli *VanClient) GetHeadlessServiceConfiguration(targetName string, protoco
 	proxySvcName := targetName + "-proxy"
 	proxyService, err := cli.KubeClient.CoreV1().Services(svcNamespace).Get(context.TODO(), proxySvcName, metav1.GetOptions{})
 	if err == nil && proxyService != nil {
-		return nil, fmt.Errorf("service %s already exists in the cluster, use any other name to expose the statefulset with a headless service", proxySvcName)
+		return nil, fmt.Errorf("To expose the %s statefulset, Skupper must create a '%s' service, and that service already exists in this namespace. To avoid this problem, rename the statefulset or the existing service.", targetName, proxySvcName)
 	}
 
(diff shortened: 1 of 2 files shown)


client/serviceinterface_update.go
```diff
@@ -252,7 +252,7 @@ func (cli *VanClient) GetHeadlessServiceConfiguration(targetName string, protoco
 	proxySvcName := targetName + "-proxy"
 	proxyService, err := cli.KubeClient.CoreV1().Services(svcNamespace).Get(context.TODO(), proxySvcName, metav1.GetOptions{})
 	if err == nil && proxyService != nil {
-		return nil, fmt.Errorf("service %s already exists in the cluster, use any other name to expose the statefulset with a headless service", proxySvcName)
+		return nil, fmt.Errorf("To expose the %s statefulset, Skupper must create a '%s' service, and that service already exists in this namespace. To avoid this problem, rename the statefulset or the existing service.", targetName, proxySvcName)
```
fgiorgetti, line 255: What do you think if instead of: "... rename the statefulset or the existing service",
we had: "... use an alternative address, rename the statefulset or the existing service".

Considering that when you're exposing using the CLI or using annotations, you can also
define an alternative address through `--address` (CLI) or `skupper.io/address` (annotation).

nluaces: I checked that, but skupper uses the target name instead of the address when it creates a headless service. 
```
$ skupper expose statefulset redis --address hello --headless                                                                                                                                                          
Error: To expose the redis statefulset, Skupper must create a 'redis-proxy' service, and that service already exists in this namespace. To avoid this problem, rename the statefulset or the existing service.
```

fgiorgetti: Actually it is using the `spec.serviceName` defined in the statefulset.
But it is throwing an error if `-address` differs from `.spec.serviceName`.

fgiorgetti (approved):

hash-d (approved):
Looks good to me!

Issue #1149, opened 2023-06-21
skupper cli is not enforcing --console-auth option


bryonbaker:
The skupper cli documentation states that `--console-auth=internal` is a mandatory switch if you specify either of the other options: `--console-user` or `--console-password`

However, the following command is accepted:
'skupper init --enable-console --enable-flow-collector --console-password password --console-user admin'
(This command is defaulting to _internal_)

The switch existence either needs to be enforced or update the documentation to state that the default is 'internal'.

Skupper version: 1.4.1
OCP version 12.4.5

pwright:
@bryonbaker I'm not seeing where it says `--console-auth=internal` is a mandatory switch if ....


What i do see is ` Valid only when --console-auth=internal` and I think that's the correct phrase, is there somewhere else you saw 'mandatory'? Should we change to ` Valid only when --console-auth=internal (default)` to avoid folks tripping on this?

Issue #1148, opened 2023-06-21
skupper cli not validating the --console-auth switch


bryonbaker:
When performing a `skupper init`, skupper cli is not validating the authentication option is one of _openshift_, _internal_, or _unsecured_. You can enter any value
E.g.: `skupper init --enable-console --enable-flow-collector --console-auth password --console-user admin`

That command is accepted, and even worse, sets an empty password.

Skupper version: 1.4.1
OCP version: 4.12.5
Browser version: Brave [Version 1.52.117 Chromium: 114.0.5735.90 (Official Build) (64-bit)](https://brave.com/latest/)

PR #1152, opened 2023-06-26
Validate console-auth options

nluaces:
Fixes: 
#1149
#1148


cmd/skupper/skupper.go
```diff
@@ -440,10 +441,26 @@ installation that can then be connected to other skupper installations`,
...
 
+			if routerCreateOpts.AuthMode != "internal" && (len(routerCreateOpts.User) > 0 || len(routerCreateOpts.Password) > 0) {
+				return fmt.Errorf("the --console-auth option must be set to internal")
+			}
+
+			if routerCreateOpts.AuthMode == "internal" && (len(routerCreateOpts.User) == 0 || len(routerCreateOpts.Password) == 0) {
+				return fmt.Errorf("the user and password values need to be configured if the --console-auth is set to internal")
+			}
+
+			if routerCreateOpts.AuthMode == "internal" && len(routerCreateOpts.User) > 0 && len(routerCreateOpts.Password) > 0 && !routerCreateOpts.EnableConsole {
```
nluaces, line 452: I am assuming that if a user wants to set up the user/pass values for the console, the console should be enabled as well. Do you agree @ajssmith ?

cmd/skupper/skupper.go
```diff
@@ -440,10 +441,26 @@ installation that can then be connected to other skupper installations`,
 				return fmt.Errorf("The --enable-flow-collector option must be used with the --enable-console option")
 			}
 
+			if routerCreateOpts.AuthMode != "internal" && (len(routerCreateOpts.User) > 0 || len(routerCreateOpts.Password) > 0) {
+				return fmt.Errorf("the --console-auth option must be set to internal")
```
grs, line 445: Maybe clarify that the reason for this is the presence of username/password?

cmd/skupper/skupper.go
```diff
@@ -440,10 +441,26 @@ installation that can then be connected to other skupper installations`,
...
+			}
+
+			if routerCreateOpts.AuthMode == "internal" && (len(routerCreateOpts.User) == 0 || len(routerCreateOpts.Password) == 0) {
+				return fmt.Errorf("the user and password values need to be configured if the --console-auth is set to internal")
```
grs, line 449: This isn't the case. Specifying the password in the options is an insecure convenience only. The username(s) and password(s) can be directly specified through the skupper-console-users secret also (probably the recommended option), or just the username can be specified in which case the password will be randomly generated.

cmd/skupper/skupper.go
nluaces: That makes sense, thanks!

nluaces: Updated in b2bd80797a311e98cad54e4793baed3b2244ef91

cmd/skupper/skupper.go
nluaces: Skupper behaviour is correct, if the user is set but not the password, it generates a random password that it is not an empty value. Updated in b2bd80797a311e98cad54e4793baed3b2244ef91

cmd/skupper/skupper.go
```diff
@@ -440,10 +441,22 @@ installation that can then be connected to other skupper installations`,
 				return fmt.Errorf("The --enable-flow-collector option must be used with the --enable-console option")
 			}
 
+			if routerCreateOpts.AuthMode != "internal" && (len(routerCreateOpts.User) > 0 || len(routerCreateOpts.Password) > 0) {
+				return fmt.Errorf("for the console to work with this user or password, the --console-auth option must be set to internal")
+			}
+
+			if routerCreateOpts.AuthMode == "internal" && !routerCreateOpts.EnableConsole {
```
grs, line 448: At present I think the restapi option also can use the auth mode. Conceivably you could use it also to secure the collector api even if console was not enabled? I don't think this is necessarily a condition we need to prevent?

nluaces: You are right, should we check that at least --enable-flow-collector is true if the console authentication has any value?

grs: Maybe we could warn, but I wouldn't prevent it for now.

nluaces: Updated in 51473eeb635042d0c3aba392f28729c52875aed2

grs (approved):

cmd/skupper/skupper.go
```diff
@@ -440,10 +441,22 @@ installation that can then be connected to other skupper installations`,
 				return fmt.Errorf("The --enable-flow-collector option must be used with the --enable-console option")
 			}
 
+			if routerCreateOpts.AuthMode != "internal" && (len(routerCreateOpts.User) > 0 || len(routerCreateOpts.Password) > 0) {
+				return fmt.Errorf("for the console to work with this user or password, the --console-auth option must be set to internal")
+			}
+
+			if len(routerCreateOpts.AuthMode) > 0 && !routerCreateOpts.EnableFlowCollector {
+				fmt.Println("The authentication only applies when --enable-console or --enable-flow-collector are enabled")
```
fgiorgetti, line 449: Is it intentional just to show this message and let it move on?

nluaces: It is intentional,  it was agreed in [this conversation.](https://github.com/skupperproject/skupper/pull/1152#discussion_r1251837114)

cmd/skupper/skupper.go
```diff
@@ -440,10 +441,22 @@ installation that can then be connected to other skupper installations`,
 				return fmt.Errorf("The --enable-flow-collector option must be used with the --enable-console option")
 			}
 
+			if routerCreateOpts.AuthMode != "internal" && (len(routerCreateOpts.User) > 0 || len(routerCreateOpts.Password) > 0) {
```
fgiorgetti, line 444: The default mode is internal, although the default value is empty. So with this
change it is forcing the --console-auth to be provided. I believe instead of checking
if the value is not "internal" it should also consider AuthMode as not being empty,
or eventually you could set a default value to `routerCreateOpts.AuthMode`.

nluaces: 1. There is no explicit default value, but if --console-auth is empty, the default is `internal`. 
2. The documentation states that --console-auth is mandatory (#1149). That's why I was forcing the value to be provided.

Probably the best would be to:
- Make an explicit default value aligned with the code
- Eliminate 51473eeb635042d0c3aba392f28729c52875aed2 otherwise it could confuse  users that could be not interested in the console/flow-collector
- Update the documentation accordingly

what do you think?

fgiorgetti: It makes sense to me!

nluaces: Updated in 48a5d32b299a44b2bffb011999c0d6664e2bc1fe

fgiorgetti (approved):

ajssmith:
In testing I tried
$skupper init --enable-console --enable-flow-collector --console-auth password
and got an expected error
Error: the --console-auth option must contain one of these values: [internal unsecured openshift]

When trying
$skupper init --enable-console --enable-flow-collector --console-auth password --console-user fred
Error: for the console to work with this user or password, the --console-auth option must be set to internal

I would have expected the --consol-auth validity check to take precedence. wdyt?

nluaces:
> I would have expected the --consol-auth validity check to take precedence. wdyt?

@ajssmith  I agree with you, I will update this.

nluaces:
@ajssmith  updated in 7493df2188767627508e981bdc3b422e405a75e4

ajssmith (approved):
lgtm, thnx

PR #1165, opened 2023-07-07
Podman flag to control IPv6 usage

### LINKED ISSUE #1164: Skupper podman is trying to use IPv6 by default
This migth cause issues to customers when IPv6 is not enabled.

fgiorgetti:
Fixes #1164


grs (approved):
Only question would be whether there is a way to determine this automatically? This change is an improvement though.

nluaces (approved):

ajssmith (approved):

grs:
Ok, if it is an explicit option on the underlying podman network, then I agree that having the same default is good and I would say we don't need a check in that case.

grs (approved):

PR #1168, opened 2023-07-10
Site controller missing permissions

fgiorgetti:


ajssmith (approved):

grs (commented):
I wonder if it would be better to add a third file that has these, i.e. one to use if you wish to allow the cluster-permissions option. What do you think?

grs:
> I wonder if it would be better to add a third file that has these, i.e. one to use if you wish to allow the cluster-permissions option. What do you think?

Or else perhaps add some comments around 'optional' permissions, to better indicate what is always required versus what is required only for specific configuration choices.

nluaces (approved):

fgiorgetti:
> I wonder if it would be better to add a third file that has these, i.e. one to use if you wish to allow the cluster-permissions option. What do you think?

Considering that this change is only done to the YAML that watches all namespaces (which already requires
cluster level permission), I don't think it is an issue. If someone does not want that, they can use the other YAML
that does not look for cluster level resources.

This change also keeps things similar to what we have on the operator. Thoughts?

fgiorgetti:
> > I wonder if it would be better to add a third file that has these, i.e. one to use if you wish to allow the cluster-permissions option. What do you think?
> 
> Or else perhaps add some comments around 'optional' permissions, to better indicate what is always required versus what is required only for specific configuration choices.

I believe comments work in this case. Do you think we need some sort of comments on the operator as well?

grs (approved):

PR #1171, opened 2023-07-11
Add option to add one edge-host hostAlias to the router

nicolacdnll:
Hi everyone,

In environments where most, if not all, outbound connectivity is blocked, we usually have to has to ask for specific firewall/NAT rules to map an internal IP address and port pair to an external one. When deploying edge Skupper sites in such environments, there are two solutions to still use the `edge-host` domain name that's in the TLS certificate.

The first is to use a local DNS and add a record so that the `edge-host` domain name points to the internal IP. However, this is not always possible because sometime we simply can't create ad hoc DNS servers or add records to the existing DNS.

The second option is to use `/etc/host`. And the suggested way to add entries `/etc/host` with Kubernetes is to use [HostAliases](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/). However, since Skupper doesn't set HostAliases, until now we have been manually editing the `skupper-router` deployment to add the HostAliases section ourselves. And we love to merge this PR merged to do the same by using the proposed new `router-edge-host-alias` flag.

Pending Issue:
What happens if there are multiple core site? Should we (1) allow a full mapping of edge-hosts and IPs in the config map? or (2) remove the cm flag and add the IP as an annotation of the secret exactly like for the `edge-host`? I prefer the second over the first, but this means that the site-controller should react also to secrets so to patch the router deployment with the new aliases.

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -651,10 +651,6 @@ func configureDeployment(spec *types.DeploymentSpec, options *types.Tuning) erro
 	}
 }
 
-// func isEdge(options types.SiteConfigSpec) bool {
-// 	return options.RouterMode == string(types.TransportModeEdge)
-// }
-
 func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId string) *types.RouterSpec {
 	// skupper-router container index
 	// TODO: update after dataplance changes
@@ -1094,7 +1090,7 @@ func (cli *VanClient) GetRouterHostAliasesSpecFromTokens(ctx context.Context) ([
 		return nil, fmt.Errorf("Failed to retrieve connection token: %w", err)
 	}
 	if len(secrets.Items) == 0 {
-		return nil, fmt.Errorf("Connection token not found")
+		return nil, nil
 	} else {
 		for _, s := range secrets.Items {
 			if alias, ok := s.ObjectMeta.Annotations["edge-alias"]; ok {
@@ -1181,14 +1177,11 @@ func (cli *VanClient) RouterCreate(ctx context.Context, options types.SiteConfig
 		ownerRefs = []metav1.OwnerReference{*siteOwnerRef}
 	}
 	var err error
-	isEdge := options.Spec.IsEdge()
-	if isEdge {
-		hostAliases, err := cli.GetRouterHostAliasesSpecFromTokens(ctx)
-		if err != nil {
-			return err
-		}
-		van.Transport.HostAliases = hostAliases
+	hostAliases, err := cli.GetRouterHostAliasesSpecFromTokens(ctx)
+	if err != nil {
+		return err
 	}
+	van.Transport.HostAliases = hostAliases
 	if options.Spec.AuthMode == string(types.ConsoleAuthModeInternal) {
 		config := `
 pwcheck_method: auxprop
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -1083,9 +1083,9 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
 	return van
 }
 
-func (cli *VanClient) GetRouterHostAliasesSpecFromTokens(ctx context.Context) ([]corev1.HostAlias, error) {
+func (cli *VanClient) GetRouterHostAliasesSpecFromTokens(ctx context.Context, namespace string) ([]corev1.HostAlias, error) {
 	hostAliasesMap := make(map[string]map[string]bool)
-	secrets, err := cli.KubeClient.CoreV1().Secrets(cli.GetNamespace()).List(ctx, metav1.ListOptions{LabelSelector: types.SkupperTypeQualifier + "=" + types.TypeToken})
+	secrets, err := cli.KubeClient.CoreV1().Secrets(namespace).List(ctx, metav1.ListOptions{LabelSelector: types.SkupperTypeQualifier + "=" + types.TypeToken})
 	if err != nil {
 		return nil, fmt.Errorf("Failed to retrieve connection token: %w", err)
 	}
@@ -1177,7 +1177,7 @@ func (cli *VanClient) RouterCreate(ctx context.Context, options types.SiteConfig
 		ownerRefs = []metav1.OwnerReference{*siteOwnerRef}
 	}
 	var err error
-	hostAliases, err := cli.GetRouterHostAliasesSpecFromTokens(ctx)
+	hostAliases, err := cli.GetRouterHostAliasesSpecFromTokens(ctx, cli.GetNamespace())
 	if err != nil {
 		return err
 	}
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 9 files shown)


grs (commented):
Thank you for describing an interesting use case!

I like the use of HostAliases, though it does require the deployment to be altered if in place already so would cause the router to restart. This is not an issue with the assumption that the secret is already in place when the deployment is created of course.

The code change you propose is a little specific (e.g. requires secret to be in place at time of site creation, assumes it is the edge host that is to be aliased, assumes a single secret etc).

I wonder whether e.g. we could add a label to the token that gave the IP to alias for the urls in the token? e.g. skupper.io/host-alias="10.0.0.1" or whatever. Even if the code change was much the same as what you have now, it would allow it to be enhanced to be more generic if needed in future?

nicolacdnll:
100% agree; moving this parameter to the token does make sense but we must document that the router will have to be restarted! Will look into it. Thanks for the feedback.

nicolacdnll:
PR updated, here a summary of the changes:

- Removed previous flags
- Included optional flag to force router restart to map inter-router and edge aliases.
- Aliases are now tied to tokens. Aliases from multiple Token are combined together.
- Since updates cause router restarts, updates are only done when the new combined aliases actually differ from the previous ones.

pkg/site/site.go
```diff
@@ -56,24 +56,25 @@ const (
...
+	SiteConfigRouterNodeSelectorKey             string = "router-node-selector"
+	SiteConfigRouterMaxFrameSizeKey             string = "xp-router-max-frame-size"
+	SiteConfigRouterMaxSessionFramesKey         string = "xp-router-max-session-frames"
+	SiteConfigRouterDataConnectionCountKey      string = "router-data-connection-count"
+	SiteConfigRouterIngressHostKey              string = "router-ingress-host"
+	SiteConfigRouterServiceAnnotationsKey       string = "router-service-annotations"
+	SiteConfigRouterPodAnnotationsKey           string = "router-pod-annotations"
+	SiteConfigRouterLoadBalancerIp              string = "router-load-balancer-ip"
+	SiteConfigRouterDisableMutualTLS            string = "router-disable-mutual-tls"
+	SiteConfigRouterForceRestartsForHostAliases string = "router-force-restarts-for-host-aliases"
```
grs, line 77: Is this new site option really needed? Is there a case where the token might have aliases added to it, but you don't want the deployment to be updated?

client/router_create.go
```diff
@@ -1081,6 +1087,63 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
...
 }
 
+func (cli *VanClient) GetRouterHostAliasesSpecFromTokens(ctx context.Context) ([]corev1.HostAlias, error) {
+	hostAliasesMap := make(map[string]map[string]bool)
+	secrets, err := cli.KubeClient.CoreV1().Secrets(cli.GetNamespace()).List(ctx, metav1.ListOptions{LabelSelector: types.SkupperTypeQualifier + "=" + types.TypeToken})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to retrieve connection token: %w", err)
+	}
+	if len(secrets.Items) == 0 {
+		return nil, fmt.Errorf("Connection token not found")
```
grs, line 1097: This doesn't seem like an error, just the absence of any aliases at present.

client/router_create.go
```diff
@@ -1118,6 +1181,14 @@ func (cli *VanClient) RouterCreate(ctx context.Context, options types.SiteConfig
 		ownerRefs = []metav1.OwnerReference{*siteOwnerRef}
 	}
 	var err error
+	isEdge := options.Spec.IsEdge()
+	if isEdge {
```
grs, line 1185: Should this be restricted only to edge sites? (The inter-router-alias annotation would in that case be irrelevant).

client/router_create.go
nicolacdnll: Good point. Will change to `return nil, nil`

client/router_create.go
nicolacdnll: It shouldn't have, removed.

pkg/site/site.go
nicolacdnll: As well, yes yes. We can remove it, no problem at all.

grs (approved):
This looks good to me. Thank you for taking the time to do this! One note is that we are starting tp plan for a next version control plane, using server-side-apply for creating the router deployment in the hope that the deployment can then be customised more directly by yaml snippets from other 'owners' (e.g. using argocd to apply memory and cpu details. In that model the host aliases could also be applied as a partial yaml deployment and if we proceed with that model that will be what we would recommend at that time.

In other words I think this is a good short term addition for the issue you describe, but in the future it may be handled a slightly different way.

nicolacdnll:
Thanks to you for taking the time to review it!

client/router_update.go
```diff
@@ -1084,6 +1086,56 @@ func (cli *VanClient) RouterUpdateAnnotations(ctx context.Context, settings *cor
...
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	return false, nil
+}
+
+func (cli *VanClient) RouterUpdateHostAliases(ctx context.Context, token *corev1.Secret) (bool, error) {
+	hostAliases, err := cli.GetRouterHostAliasesSpecFromTokens(ctx)
```
fgiorgetti, line 1127: This is not working when the `site-controller` is deployed watching ALL namespaces.
You need `GetRouterHostAliasesSpecFromTokens` to use the token's namespace for that, instead of relying on the cli's namespace.

nicolacdnll: Thanks for the review and spotting it, I wasn't even aware of this. The latest commit fixes it.

fgiorgetti (approved):
LGTM! Thank you.

nicolacdnll:
Great!

BTW, I might be wrong but it looks like the build-workflow is failing for something unrelated to this PR, no?

```
=== RUN   TestNetworkStatusWithCluster
W0913 09:40:08.832494  197872 warnings.go:70] tls: failed to find any PEM data in certificate input
    skupper_cluster_test.go:79: Expected to match: 
         ^Sites\:\n╰─\s\[local\]\s[0-9a-f-]*\s-\s.*\n\s{3}URL:\s\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\n\s{3}mode:\s.*\n\s{3}name:\s.*\n\s{3}namespace:\s.*\n\s{3}version:\s.*\n\s{3}╰─\sServices\:\n\s{6}╰─\sname:\s.*\n\s{9}address:\s.*\:\s\d{4}\n\s{9}protocol:\s.*\n\s{9}╰─\sTargets\:\n\s{12}╰─\sname:\s.*
        Got: Unable to retrieve network information: context deadline exceeded
        
        Loading just local information:
        The site configuration is not available: context deadline exceeded
--- FAIL: TestNetworkStatusWithCluster (146.66s)
```

grs:
Yes, the CI failures are not related to this PR. @ajssmith are you ok with this change?

nicolacdnll:
> Yes, the CI failures are not related to this PR. @ajssmith are you ok with this change?

Friendly up.

ajssmith (approved):
lgtm2

PR #1174, opened 2023-07-13
Publish-not-ready-addresses value is not being sync

nluaces:
This pull request unifies the results of the expose command with the combination of create & bind commands:
- publishNotReadyAddresses value is synced over the VAN
- headless services can be created when binding a statefulset to a service


nluaces:
@grs that's correct, but when the expose command is used, the current behaviour is that this value gets synced in all the sites, and I was thinking that having two processes (expose and create+bind) that should end up with the same result but at the end show different things,  can make the user think that there is a mistake somewhere.

grs:
I agree that there should be no inconsistency between expose v. create+bind here. How is the syncing different for expose though?

nluaces:
@grs ah, do you mean to not sync the value in any command, right? that makes sense. But what happens if you expose a service with that value in a different namespace for the target? (I will check it)

grs:
Yes, I don't actually understand how expose could sync the value if bind does not, as the syncing mechanism is (or should be) the same in each case?

grs:
Good analysis! I understand now. I think your fix is correct also. (I think it should perhaps be a property of the target, not the service? but that is not something we should change here.) Thanks for clarifying!

grs (approved):

nluaces:
I like your idea of having this value in the target, I just created #1178 to change it in a future pull request. Thanks!

grs (approved):

fgiorgetti:
One thing I have noticed with the create + bind approach is that on the remote site,
a service ingress is already created at the time we do a `service create` on the origin site,
and the resulting k8s service is not a headless service (it has an IP).

If the --headless option lies on the `service bind` instead of `service create`, then the service sync
should recreate the k8s service on the remote site as a headless one (without an IP), or it won't work.

fgiorgetti (approved):

Issue #816, opened 2022-07-06
Support of Gateway on Mac OSX


hguerrero:
Currently, the launch of docker or podman assumes the machine host is a Linux system. When using docker or podman on a Mac OSX system, the `host` network is not available, so each of the ports needs to be bonded manually, or use `-P` for a random selection. 

The launch script and the init command should be able to detect an OSX system and change the way it is exposing the ports to enable the networking connectivity.

ajssmith:
V2 non-kube on (later) osx should work with host network

PR #1176, opened 2023-07-14
use adminport instead of `host` network

hguerrero:
a first approach to address #816 

Use the admin port mapped explicitly when launching the container.

This fixes the `gateway status` and most of the commands related to binding, as it does not need to expose new ports. 

Missing: 
- Restart the container when a new port needs to be opened when forwarding from Kubernetes.
- Test if this affects podman

### DIFF
--- a/client/gateway.go
+++ b/client/gateway.go
@@ -596,7 +596,7 @@ func (cli *VanClient) setupGatewayConfig(ctx context.Context, gatewayName string
 	}
 	gatewayConfig.Listeners["amqp"] = qdr.Listener{
 		Name: "amqp",
-		Host: "localhost",
+		Host: "0.0.0.0",
 		Port: int32(amqpPort),
 	}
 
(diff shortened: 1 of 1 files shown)


client/gateway.go
```diff
@@ -743,8 +745,8 @@ func (cli *VanClient) gatewayStartContainer(ctx context.Context, gatewayName str
 		"-d",
 		"--name",
 		gatewayName,
-		"--network",
-		"host",
```
fgiorgetti, line 747: I believe this would break `gateway forward` as host ports won't be bound
to the gateway container.

hguerrero: It is already broken in non-Linux systems. IIRC, the recommended way to proceed for Linux systems is using podman sites instead. If it is not the case, then we can add a flag to detect the OS running and keep `--network host` for the ones where it is supported.

hguerrero: @fgiorgetti Added a check to get the host OS and only do port mapping for `windows` and `darwin`.

fgiorgetti (commented):
Along with these changes, I believe it would also be safe to throw an error inside
`NewCmdForwardGateway` and `NewCmdUnforwardGateway` as `skupper gateway forward` and `unforward` won't work without the `host` network mode.

PR #1196, opened 2023-08-07
dp collector lite

ajssmith:
collector for site, router records only for dp status


fgiorgetti:
Other than this minor comment, it looks good to me.
The leader election seems to be working fine, the new configmap also looks right to me.
Tested on latest minikube and minishift (3.11).

grs (approved):
Agree with Fernando that the retry without retrieving the latest configmap is ineffective. Due to the leader election there should be no conflicts but would be clearer to change the code. Other than that looks good to me!

nluaces (approved):

fgiorgetti (approved):

PR #1201, opened 2023-08-11
add regex match

ajssmith:


ajssmith:
Note: regex performance degradation noted, changed to prefix match which is behaving satisfactorily for console

nluaces (approved):

pkg/flow/record.go
```diff
@@ -661,7 +662,7 @@ func matchFieldValue(x interface{}, y string) bool {
 	if x != nil && y != "" {
 		switch x.(type) {
 		case string:
-			return x.(string) == y
+			return strings.HasPrefix(x.(string), y)
```
grs, line 665: I'm lacking the broader context here, but as I understand it this changes matchFieldValue to always match string based values by prefix. Is it possible for this to ever result in an error? E.g. if one site name is a prefix of another site name?

ajssmith: To provide additional context, this change was to support console behavior to provide query filter based on prefix match. In the case of a prefix value matching both a field and a field prefix, this is expected and both results will be provided in the results.

grs (approved):

PR #1203, opened 2023-08-11
Service unexpose fix

### LINKED ISSUE #1202: Unexposing services not working
Unexposing a target service is not working with 1.4.2.

fgiorgetti:
Fixes #1202.

### DIFF
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -329,7 +329,11 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				if (t.Name == targetName || (t.Name == "" && targetName == serviceName)) && (t.Namespace == "" || t.Namespace == namespace) {
+				targetFound := t.Name == targetName
+				unqualifiedServiceFound := t.Name == "" && targetName == serviceName && t.Service == targetName
+				qualifiedServiceFound := t.Service == fmt.Sprintf("%s.%s", targetName, namespace)
+				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
+				if (targetFound || unqualifiedServiceFound || qualifiedServiceFound) && namespaceMatches {
 					modified = true
 				} else {
 					targets = append(targets, t)
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -197,12 +197,12 @@ func (cli *VanClient) ServiceInterfaceBind(ctx context.Context, service *types.S
 		if err != nil {
 			return err
 		}
-		err = validateCrossNamespacePermissions(cli, namespace)
+		_, svcNamespace := kube.GetServiceName(targetName, utils.GetOrDefault(namespace, cli.GetNamespace()))
+		err = validateCrossNamespacePermissions(cli, svcNamespace)
 		if err != nil {
 			return err
 		}
 		deducePorts := len(service.Ports) == 0 && len(targetPorts) == 0
-		svcNamespace := utils.GetOrDefault(namespace, cli.GetNamespace())
 		target, err := kube.GetServiceInterfaceTarget(targetType, targetName, deducePorts, svcNamespace, cli.KubeClient, cli.OCAppsClient)
 		if err != nil {
 			return err
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -329,10 +329,9 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				targetFound := t.Name == targetName && t.Service == ""
-				targetServiceFound := t.Service == targetName
-				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
-				if (targetFound || targetServiceFound) && namespaceMatches {
+				targetFound := (t.Name == targetName && t.Service == "") && t.Namespace == namespace
+				targetServiceFound := t.Service == targetName && t.Namespace == ""
+				if targetFound || targetServiceFound {
 					modified = true
 				} else {
 					targets = append(targets, t)
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -329,6 +329,7 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
+				targetName, namespace = kube.GetServiceName(targetName, namespace)
 				targetFound := t.Name == targetName
 				unqualifiedServiceFound := t.Name == "" && targetName == serviceName && t.Service == targetName
 				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -197,12 +197,12 @@ func (cli *VanClient) ServiceInterfaceBind(ctx context.Context, service *types.S
 		if err != nil {
 			return err
 		}
-		_, svcNamespace := kube.GetServiceName(targetName, utils.GetOrDefault(namespace, cli.GetNamespace()))
-		err = validateCrossNamespacePermissions(cli, svcNamespace)
+		err = validateCrossNamespacePermissions(cli, namespace)
 		if err != nil {
 			return err
 		}
 		deducePorts := len(service.Ports) == 0 && len(targetPorts) == 0
+		svcNamespace := utils.GetOrDefault(namespace, cli.GetNamespace())
 		target, err := kube.GetServiceInterfaceTarget(targetType, targetName, deducePorts, svcNamespace, cli.KubeClient, cli.OCAppsClient)
 		if err != nil {
 			return err
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -329,11 +329,10 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				targetName, namespace = kube.GetServiceName(targetName, namespace)
-				targetFound := t.Name == targetName
-				unqualifiedServiceFound := t.Name == "" && targetName == serviceName && t.Service == targetName
+				targetFound := t.Name == targetName && t.Service == ""
+				targetServiceFound := t.Service == targetName
 				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
-				if (targetFound || unqualifiedServiceFound) && namespaceMatches {
+				if (targetFound || targetServiceFound) && namespaceMatches {
 					modified = true
 				} else {
 					targets = append(targets, t)
--- a/client/serviceinterface_update.go
+++ b/client/serviceinterface_update.go
@@ -331,9 +331,8 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			for _, t := range service.Targets {
 				targetFound := t.Name == targetName
 				unqualifiedServiceFound := t.Name == "" && targetName == serviceName && t.Service == targetName
-				qualifiedServiceFound := t.Service == fmt.Sprintf("%s.%s", targetName, namespace)
 				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
-				if (targetFound || unqualifiedServiceFound || qualifiedServiceFound) && namespaceMatches {
+				if (targetFound || unqualifiedServiceFound) && namespaceMatches {
 					modified = true
 				} else {
 					targets = append(targets, t)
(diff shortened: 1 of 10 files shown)


client/serviceinterface_update.go
```diff
@@ -329,7 +329,11 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				if (t.Name == targetName || (t.Name == "" && targetName == serviceName)) && (t.Namespace == "" || t.Namespace == namespace) {
+				targetFound := t.Name == targetName
+				unqualifiedServiceFound := t.Name == "" && targetName == serviceName && t.Service == targetName
```
grs, line 333: Can you elaborate on the bug this is fixing? What were the args for the expose and the unexpose and what is the actual result v expected result?

fgiorgetti: Actually the `unqualifiedServiceFound` (case) is not related to the issue itself.
It does not seem to be a valid condition to me anymore (or maybe/probably something I am not considering).

As is, meaning, without the latest `t.Service == targetName` comparison, it basically checks if the respective
target.Name is empty, then no other param for the respective target being iterated is validated, it simply 
compares the provided target name and the provided service address. This way if multiple targets have empty
names (which I believe is no longer a valid condition?), all of them would be removed.

Do you know/remember of something I am not considering for this case?
If not, do you think it should be removed?

grs: Can you give an example of a failing interaction?

fgiorgetti: Sure. The latest case: `qualifiedServiceFound := t.Service == fmt.Sprintf("%s.%s", targetName, namespace)` is
actually fixing the related issue.

If you use a latest version of Skupper (1.4.2) and expose an existing k8s service, i.e.:

```
skupper expose service nginx --address nginx-skupper
```

then looking at the targets in the skupper-services ConfigMap, we have this:

```
"targets":[{"name":"nginx.sk1","service":"nginx.sk1","namespace":"sk1"}]
```

As you can see, the exposed k8s service is now qualified with the "target-namespace" option,
which was introduced recently (but it was not actually tied to the unexposed command before this PR).

grs: To me it is the behaviour for expose that seems wrong. If you don't specify a target-namespace for the service then the service as recorded in the target should be exactly what the user specified. The '.sk1' suffix seems wrong to me.

fgiorgetti: I believe you're correct. Possibly it was done like this so that the connector is created using the value
stored in the "target.service" as the target host. I will modify the expose to stop storing the namespace as part
of the service name.

Now two questions.

1. Do you think it is worth adding an skupper update verification on eventual service
definitions with targets containing service names with the namespace suffix and fix that as well?

2. What do you think of the updated condition above (`unqualifiedServiceFound`)? Should it be kept in this PR?

fgiorgetti: Answering my own questions. I believe that for the former 1st question, it is not needed to due a minor
impact, as it would only impact exposed services using the CLI (services exposed using annotations
should not be impacted).

For the 2nd question, I believe that the udpated logic seems right. So I think it should be kept.

grs: Generally, I believe what we want is that the arguments needed for a successful unexpose  should have values that match those provided for those same options on the expose.

In my opinion, the simplest solution is to state that the target namespace is not applicable when the target type is 'service'. It is possible to target a service in a different namespace simply by qualifying the service name provided. With that view we would check if target-namespace was specified for target type service and report an error if so.  Then unexpose only needs to test either the target-name+target-namespace or the servicename (which should be provided in the exact form used for expose).

To me this avoids unnecessary complication.

However if that approach is not desired, I think it would be better to ensure that the namespace of a target is only populated if the target-namespace option was specified. I.e. if it is the current namespace, it would be empty.

fgiorgetti: > Generally, I believe what we want is that the arguments needed for a successful unexpose should have values that match those provided for those same options on the expose.

Agreed. I believe with the latest changes, this is how it is supposed to work now for services as well.

> In my opinion, the simplest solution is to state that the target namespace is not applicable when the target type is 'service'. It is possible to target a service in a different namespace simply by qualifying the service name provided. With that view we would check if target-namespace was specified for target type service and report an error if so. Then unexpose only needs to test either the target-name+target-namespace or the servicename (which should be provided in the exact form used for expose).

I have added one minor change that basically allows you to use it either way. So basically you can expose and unexpose as follows:

```
skupper expose service my-service.target-namespace --address my-skupper-service
skupper unexpose service my-service.target-namespace --address my-skupper-service
```

or

```
skupper expose service my-service --target-namespace target-namespace --address my-skupper-service
skupper unexpose service my-service --target-namespace target-namespace --address my-skupper-service
```

I believe this is a more uniform way to tell the story of exposing resources from different namespaces, as
services can be exposed as you mentioned (just by adding the . namespace) or using --target-namespace,
which makes it similar to how other kind of resources are exposed.

grs: I think the apparent uniformity is something of an illusion. For deployments etc, the target namespace is used to query the api server, which is why using it with those target types requires extra permissions. The way services are handled is very different. (Arguably the target type 'service' is misleading name)

With your changes `skupper expose service foo.bar --target-namespace baz --port 80 --address foo` would cause the baz to be silently ignored.  `skupper expose service east.database.acme.org --port 80 --address db` would be stored as service='east' namespace='database.acme.org' which is not ideal since the latter is not actually a valid namespace.

If we must keep the --target-namespace for services I think we should aim to have the target fields match what was actually specified by the user, without modification, or else we will not necessarily be able to determine that again should there be a reason to do so. I.e. the service field should always hold what was in the <target-name> argument and the namespace field what was in the --target-namespace option.

I increasingly feel that the best option would be to just say that the target-namespace is not compatible with the service target-type (since the mechanism that option actually is designed to invoke is not relevant for the way we handle service targets)

fgiorgetti: I agree that the mixed behavior of exposing/unexposing a qualified service along with using the
--target-namespace flag is obscure and should be rejected.

This behavior is already present today in the expose command, this latest change just added to the
unexpose. So I think this PR should also remove that behavior from the expose command.

> The way services are handled is very different. (Arguably the target type 'service' is misleading name)

True. One interesting thing is that during expose, the target service is being verified in order to determine
the target ports. But when we do a service bind, then the target service name is handled merely as a hostname.
Using 'service' as the target type is really misleading in this case.

> I increasingly feel that the best option would be to just say that the target-namespace is not compatible with the service target-type (since the mechanism that option actually is designed to invoke is not relevant for the way we handle service targets)

I was leaning toward leaving both and reject mixing them by verifying if target service was qualified, but it might
be simpler to simply reject using the --target-namespace for target type == service and do as you suggested.

Once the definition monitor is updated to handle target service names from other namespaces, it will make 
usage seamless between CLI and the declarative approach as well.

cmd/service-controller/policy_controller.go
```diff
@@ -347,15 +347,19 @@ func (c *PolicyController) validateExposeStateChanged() {
 		for _, target := range service.Targets {
 			svcNamespace := utils.GetOrDefault(target.Namespace, c.cli.GetNamespace())
 			targetType := c.inferTargetType(target, svcNamespace)
-			res := c.validator.ValidateExpose(targetType, target.Name)
+			targetName := target.Name
+			if targetType == "service" && svcNamespace != c.cli.GetNamespace() {
```
grs, line 351: If we are preventing the target namespace being specified when target tpe is 'service', do we need this here?

pkg/kube/misc.go
```diff
@@ -70,11 +71,10 @@ func GetServiceInterfaceTarget(targetType string, targetName string, deducePort
 		return nil, fmt.Errorf("VAN service interfaces for pods not yet implemented")
 	} else if targetType == "service" {
 		var svcName, svcNamespace string = GetServiceName(targetName, namespace)
-		var formattedName = fmt.Sprintf("%s.%s", svcName, svcNamespace)
 		target := types.ServiceInterfaceTarget{
-			Name:      formattedName,
-			Service:   formattedName,
-			Namespace: namespace,
+			Name:      svcName,
```
grs, line 75: Since we are not allowing target-namespace in conjunction with service targets, I would think we should just use the specified targetName as the Name and the Service here, not set the namespace and and not try to interpret the targetName in anyway.

client/serviceinterface_update.go
```diff
@@ -329,7 +329,11 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				if (t.Name == targetName || (t.Name == "" && targetName == serviceName)) && (t.Namespace == "" || t.Namespace == namespace) {
+				targetName, namespace = kube.GetServiceName(targetName, namespace)
```
grs, line 332: I think the logic here should be that we remove if targetName == t.Service || ((t.Service == "" && targetName == t.Name) && (namespace == "" || namespace == t.Namespace)). I.e. we don't need to care about the namespace at all if the Service on the target is specified, nor about any interpretation of whether the service is qualified or not.

pkg/service/bindings.go
```diff
@@ -396,11 +396,16 @@ func (sb *ServiceBindings) addServiceTarget(name string, service string, port ma
 		identifier: service,
 		namespace:  namespace,
 	}
+	targetService := service
+	// on annotated services, namespace is not yet available
+	if namespace != "" {
+		targetService = fmt.Sprintf("%s.%s", service, namespace)
```
grs, line 402: As namespace cannot be specified when target is service, we don't need this I don't believe.

cmd/service-controller/policy_controller.go
fgiorgetti: We are preventing the `--target-namespace` flag from being provided through the CLI when the target type is
service, but allowing users to bind a service from another namespace by qualifying the service.

With latest version of Skupper, the namespace field of a target in skupper-services is never empty.

So if we populate the "service" field as a qualified value, the namespace field would be populated incorrectly.

My inclination is to keep the namespace field being populated, as is, and when a service is qualified, we
split it and store the qualified namespace at the namespace field and leave the service field unqualified there.

Do you think it is better to change this current behavior and leave the "namespace" field empty in the targets
when target type is service?

fgiorgetti: I will make this change then you ping you again for another round of review.
And I believe your other questions might be answered once I have it done.

pkg/kube/misc.go
fgiorgetti: We actually need the target namespace here as it is used to deduce ports below.

grs: Deducing the ports is an optional feature. We can still do that without actually recording the namespace in the target structure. It is fine if this only works when the target is the unqualified name and the namespace is the site namespace.

client/serviceinterface_update.go
fgiorgetti: Logic has been updated and feels clearer now.

cmd/service-controller/policy_controller.go
fgiorgetti: Done.

pkg/service/bindings.go
fgiorgetti: It is no longer needed.

client/serviceinterface_update.go
```diff
@@ -329,7 +329,10 @@ func removeServiceInterfaceTarget(serviceName string, targetName string, deleteI
 			modified := false
 			targets := []types.ServiceInterfaceTarget{}
 			for _, t := range service.Targets {
-				if (t.Name == targetName || (t.Name == "" && targetName == serviceName)) && (t.Namespace == "" || t.Namespace == namespace) {
+				targetFound := t.Name == targetName && t.Service == ""
+				targetServiceFound := t.Service == targetName
+				namespaceMatches := t.Namespace == "" || t.Namespace == namespace
```
grs, line 334: I think the logic would be clearer as

`
targetFound := (t.Name == targetName && t.Service == "") && t.Namespace == namespace
targetServiceFound := t.Service == targetName && t.Namespace == ""
if targetFound || targetServiceFound) {

`

client/serviceinterface_update.go
```diff
@@ -197,12 +197,12 @@ func (cli *VanClient) ServiceInterfaceBind(ctx context.Context, service *types.S
 		if err != nil {
 			return err
 		}
-		err = validateCrossNamespacePermissions(cli, namespace)
+		_, svcNamespace := kube.GetServiceName(targetName, utils.GetOrDefault(namespace, cli.GetNamespace()))
```
grs, line 200: I don't understand why we want this change. We only need to check for the extra permissions if a different namespace has been specified. `err = validateCrossNamespacePermissions(cli, namespace)` seems the right way to do that to me. Or am I missing something?

fgiorgetti: If the target is a qualified service name, then the target namespace is part of the targetName,
so we pick the namespace from it, otherwise we use the namespace provided.

client/serviceinterface_update.go
fgiorgetti: Certainly more precise! I will update it.

client/serviceinterface_update.go
grs: But we don't need to do any permission validation in that case.

grs: I don't think we should try to read the service unless the target name is a service in the site namespace. That is what I believe we did before the introduction of the target namespace.

fgiorgetti: I will make it this change for your evaluation. One thing that will happen is that, once ports cannot be
deduced, user will need to provide `--port` when exposing a service from a different namespace. Which
seems reasonable to me. I will push it then let me know if you're ok with it.

fgiorgetti:
> Thank you Fernando! Sorry for all the nit picking!

I appreciate your attention and careful review. Thank you.

nluaces (approved):

PR #1205, opened 2023-08-11
add explicit per-pod type annotation

waldner:
New keys in the `skupper-site` configmap to explicitly set annotations for `service-controller`, `router` and `prometheus` pods.


waldner:
I'm not familiar with CI pipelines used by skupper so I'm not sure whether the failure is caused by the changes or is independent.

grs (approved):
Thank you @waldner !

fgiorgetti (approved):
LGTM.

waldner:
Thanks! I suppose this will go into 1.5 right?

PR #1213, opened 2023-08-17
Prometheus container on podman sites

fgiorgetti:


grs (approved):

cmd/flow-collector/main.go
```diff
@@ -205,7 +205,7 @@ func main() {
 		}
 		flowRecordTtl, _ = time.ParseDuration(os.Getenv("FLOW_RECORD_TTL"))
 		enableConsole, _ = strconv.ParseBool(os.Getenv("ENABLE_CONSOLE"))
-		prometheusUrl = os.Getenv("PROMETHEUS_URL")
+		prometheusUrl = "http://skupper-prometheus:9090/api/v1/"
```
nluaces, line 208: Is the environment variable not going to be used for future developments like adding an external prometheus server? I agree that this should be the value as default.

nluaces (approved):

ajssmith (approved):
lgtm

PR #1214, opened 2023-08-21
reload tls certificates for each connection

grs:


grs:
At present the tls certs/key are loaded once at startup. This means that if they are updated, the new versions arenot picked up. This change causes them to be loaded everytime we create a new AMQP connection.

fgiorgetti (approved):

PR #1218, opened 2023-08-24
Move to quay

hash-d:
This changes Skupper test source code to use auxiliary external test images (such as mongo and redis) from our quay.io/skupper repo, instead of elsewhere.

This is to avoid Dockerhub pull limits, and also to facilitate testing in disconnected clusters.

The change also includes automation to copy the images to and from our repo.


test/images/Makefile
```diff
@@ -0,0 +1,180 @@
...
+	docker.io/maistra/examples-bookinfo-productpage-v1\:0.12.0 \
+	docker.io/maistra/examples-bookinfo-ratings-v1\:0.12.0 \
+	docker.io/maistra/examples-bookinfo-details-v1\:0.12.0 \
+	docker.io/maistra/examples-bookinfo-reviews-v3\:0.12.0 \
+	docker.io/library/mongo\:5.0 \
+	docker.io/nginxinc/nginx-unprivileged\:stable-alpine
+
+# These are the images that are generated by this Makefile (as opposed to
+# created elsewhere and just copied here).  To add a new image, simply create
+# Dockerfile.IMAGENAME and add the IMAGENAME below.
```
fgiorgetti, line 106: New images pushed into quay are private by default.
Maybe just add a note advising users to make image public.

hash-d: @fgiorgetti, I have something for that on line 66.  Do you think I should move or duplicate it here?

Issue #1191, opened 2023-07-27
After deleting a skupper service with the same address as a kubernetes service, the original ports are not restored


qb0C80aE:
# Summary
When creating a skupper service with the same address as an existing kubernetes service name,the original `targetPort` of the service is stored in the `originalTargetPort` annotation.
However, when the skupper service is deleted,the original `targetPort` of the kubernetes service does not get restored.

# Version
1.4.2

# Detail

```
### initial kubernetes service status
$ kubectl get svc app1 -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"name":"app1","namespace":"site1"},"spec":{"ports":[{"name":"someport1","port":5000,"targetPort":20000},{"name":"someport2","port":5001,"targetPort":20001},{"name":"someport3","port":5002,"targetPort":20002}],"selector":{"app":"app1"}}}
  creationTimestamp: "2023-07-27T05:20:39Z"
  name: app1
  namespace: site1
  resourceVersion: "74362"
  uid: 04db53b7-4331-4109-8944-ec0292eed465
spec:
  clusterIP: 172.30.66.244
  clusterIPs:
  - 172.30.66.244
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: someport1
    port: 5000
    protocol: TCP
    targetPort: 20000
  - name: someport2
    port: 5001
    protocol: TCP
    targetPort: 20001
  - name: someport3
    port: 5002
    protocol: TCP
    targetPort: 20002
  selector:
    app: app1
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}

$ skupper service status
No services defined

### expose a service with annotation
$ kubectl annotate service app1 "skupper.io/proxy=http"
service/app1 annotated

$ kubectl get svc app1 -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    internal.skupper.io/originalAssignedPort: 5000:1039,5001:1040,5002:1041
    internal.skupper.io/originalSelector: app=app1
    internal.skupper.io/originalTargetPort: 5000:20000,5001:20001,5002:20002
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #1224, opened 2023-08-26
fix original service restoration

qb0C80aE:
Issue: #[1191](https://github.com/skupperproject/skupper/issues/1191)

As suggested, I also considered iterating through the return value if `GetOriginalTargetPorts()`.
However, its type is map[int]int, so the order is unpredictable.
Since I thought it's better to preserve the previous port names as much as possible, I took the way iterating through the current ports.

```
During the iteration:
1. If the port value exists as a key in the original port map, the port's name is kept unchanged and restored the TargetPort using the value.
2. If the port value doesn't exist as a key in the original port map, there is no need to restore it.

After the iteration:
3. The map keys that haven't been referenced up to this point need to be restored as port entries.
```

The reason why step 2. and 3. above might be necessary is:

Suppose the original service `app1` has only one port, `80:8080`.

[I] If there is no `--port` specified on `expose service`, and `app1` is exposed with the same name as address, the number of entries in service's `ports` and the port values inside that exactly matchs the content of service's `originalTargetPort` annotation. In this case, processing step 1. above is enough.
```
ex.
run `skupper expose service app1 --address app1`
then run `skupper unexpose service app1.site1 --address app1`
```

[2] If `--port` is specified on `expose service`, or `create service`, with the same name address,
the number of ports and port values can be different from the original ones in `app1` (I'm not sure whether accepting this is intended behavior or not).
In this case, steps 2. and 3. become necessary.
ex. 
```
run `skupper expose service app1 --address app1 --port 11111 22222 ...`
then run `skupper unexpose service app1.site1 --address app1`

or

run `skupper service create app1 11111 22222 ...`
then run `skupper service delete app1`
```

In steps 2. and 3., the port names and the order of entries cannot be restored anyway.

### DIFF
--- a/cmd/service-controller/definition_monitor.go
+++ b/cmd/service-controller/definition_monitor.go
@@ -5,7 +5,6 @@ import (
 	jsonencoding "encoding/json"
 	"fmt"
 	"reflect"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


grs (approved):

nluaces (approved):

fgiorgetti:
@qb0C80aE would you mind rebasing (on main) and updating this PR?

qb0C80aE:
not at all, I've rebased it on main.

qb0C80aE:
@fgiorgetti @grs @nluaces Thank you for your time!

grs:
Thank you @qb0C80aE !

Issue #1185, opened 2023-07-20
Modify the default ingress and ingress-hosts values for podman sites


fgiorgetti:
On podman sites, the default ingress is `host`, which requires `--ingress-host` to be provided.
This gives a bad user expirence, as it differs from kubernetes sites where you can simply use `skupper init`.

The suggestion is to switch the default ingress to `none` and also set the default `--ingress-host` (if none provided)
to the fully qualified hostname.

DendoD96:
Hi @fgiorgetti, 
Could you help me understand this statement?

From what I know, on podman sites, ingress option can be set to `none` | `external`, by default the option is set to `''`. 
In the first case skupper does not support linking to the current site and so `--ingress-host` is ignored, while in the other two cases `--ingress-host` is required. So, if we set `ingress` to `none`, why shoud we set the default ingress-host to the FQHN?

Best regards,
Daniele

fgiorgetti:
Fixed by #1230

PR #1230, opened 2023-09-05
Add FQDNs as default ingressHosts in podman.

DendoD96:
Try to implement the enhancement referenced in Issue #1185.

Basically, i get all unicast IP of the machine and for all IPv4 different from loopback one try a DNS reverse resolution. If the resolution fails, i use the IP as ingress host.

### DIFF
--- a/cmd/skupper/skupper_podman_site.go
+++ b/cmd/skupper/skupper_podman_site.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"strings"
 
@@ -81,7 +82,26 @@ func (s *SkupperPodmanSite) Create(cmd *cobra.Command, args []string) error {
 	if routerCreateOpts.Ingress != types.IngressNoneString {
 		// Validating ingress hosts (required as certificates must have valid hosts)
 		if len(site.IngressHosts) == 0 {
-			return fmt.Errorf("At least one ingress host is required")
+			// Get all unicast IP addresses on the machine
+			addresses, err := net.InterfaceAddrs()
+			if err != nil {
+				return fmt.Errorf("Cannot get a default ingress host")
+			}
+			var ingressHosts []string
+			for _, address := range addresses {
+				ipnet, ok := address.(*net.IPNet)
+				if ok && !ipnet.IP.IsLoopback() && ipnet.IP.To4() != nil {
+					ipv4ValidAddress := ipnet.IP.String()
+					// Try a reverse lookup of a valid IPv4 address
+					fqdns, err := net.LookupAddr(ipv4ValidAddress)
+					if err != nil {
+						ingressHosts = append(ingressHosts, ipv4ValidAddress)
+					} else {
+						ingressHosts = append(ingressHosts, fqdns...)
+					}
+				}
+			}
+			site.IngressHosts = append(site.IngressHosts, ingressHosts...)
 		}
 	} else {
 		// If none is set, do not allow any ingress host (ignore those provided via CLI)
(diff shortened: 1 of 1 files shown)


cmd/skupper/skupper_podman_site.go
```diff
@@ -78,7 +79,26 @@ func (s *SkupperPodmanSite) Create(cmd *cobra.Command, args []string) error {
 	if routerCreateOpts.Ingress != types.IngressNoneString {
 		// Validating ingress hosts (required as certificates must have valid hosts)
 		if len(site.IngressHosts) == 0 {
-			return fmt.Errorf("At least one ingress host is required")
+			// Get all unicast IP addresses on the machine
+			addresses, err := net.InterfaceAddrs()
+			if err != nil {
```
fgiorgetti, line 84: Maybe before calling net.InterfaceAddrs, you could also try calling `os.Hostname()`.
If the hostname is returned, it could be used (with no error here).
But if both failed, then the error below can still be returned.

cmd/skupper/skupper_podman_site.go
```diff
@@ -78,7 +79,26 @@ func (s *SkupperPodmanSite) Create(cmd *cobra.Command, args []string) error {
...
+				return fmt.Errorf("Cannot get a default ingress host")
+			}
+			var ingressHosts []string
+			for _, address := range addresses {
+				ipnet, ok := address.(*net.IPNet)
+				if ok && !ipnet.IP.IsLoopback() && ipnet.IP.To4() != nil {
+					ipv4ValidAddress := ipnet.IP.String()
```
fgiorgetti, line 91: I believe that the `ipv4ValidAddress` could always be added to the `ingressHosts` slice.

cmd/skupper/skupper_podman_site.go
```diff
@@ -78,7 +79,26 @@ func (s *SkupperPodmanSite) Create(cmd *cobra.Command, args []string) error {
...
+					// Try a reverse lookup of a valid IPv4 address
+					fqdns, err := net.LookupAddr(ipv4ValidAddress)
+					if err != nil {
+						ingressHosts = append(ingressHosts, ipv4ValidAddress)
```
fgiorgetti, line 95: ... then you could get rid of this and just add the fqdns in case error is nil. Wdyt?

fgiorgetti (commented):
Thanks for the PR! I have a couple of suggestions, but overall it looks good.

cmd/skupper/skupper_podman_site.go
DendoD96: Yes, LGTM.

cmd/skupper/skupper_podman_site.go
DendoD96: I don't know if I understood correctly: i think sometimes hostname could be different from fqdn, soI add both hostname and any fqdn different from it. Let me know.

cmd/skupper/skupper_podman_site.go
```diff
@@ -78,7 +80,35 @@ func (s *SkupperPodmanSite) Create(cmd *cobra.Command, args []string) error {
 	if routerCreateOpts.Ingress != types.IngressNoneString {
 		// Validating ingress hosts (required as certificates must have valid hosts)
 		if len(site.IngressHosts) == 0 {
-			return fmt.Errorf("At least one ingress host is required")
+			var ingressHosts []string
+			// Get hostname of the machine
+			hostname, err := os.Hostname()
+			if err == nil {
+				ingressHosts = append(ingressHosts, hostname)
+			} else {
```
fgiorgetti, line 88: This part should not be in the else. It's better to always run it, as it
would use all IP4 addresses as part of the valid set of ingress-hosts.

fgiorgetti (approved):
LGTM! Thank you @DendoD96 for your contribution.

@ssorj What do you think? This PR will basically use the hostname,
along with all IPv4 addresses and their respective names. Users can
then specify the desired ingress host when creating a new token.

grs (approved):

DendoD96:
@fgiorgetti Thank you for your time. It has been a pleasure for me. Hope to contribute again during my learning Go journey!

fgiorgetti:
@DendoD96 would you mind rebasing and updating this PR?

DendoD96:
@fgiorgetti Done.

PR #1231, opened 2023-09-06
Fix for TestHelloWorldCLIOnPodman runs on OpenShift

hash-d:
This is similar to #1110; OpenShift requires RunAsUser within specific range


hash-d:
Fix validated on CRC and Minikube.  The CircleCI errors that happened on the PR validation are unrelated.

granzoto (approved):
lgtm

Issue #1208, opened 2023-08-15
"Secret already exists" when binding a service right after its creation


hash-d:
While testing a fix for #1153, my local tests failed intermittently with a different error:

    http.go:908: assertion failed: error is not nil: secrets "skupper-tls-nghttp1tls" already exists

or

    http.go:885: assertion failed: error is not nil: secrets "skupper-tls-nghttp2tls" already exists

Looking at the code, two operations are referencing the `types.ServiceInterface` that defines the offending secret name, one right after the other:

    904         err = prv1Cluster.VanClient.ServiceInterfaceCreate(ctx, &http1TlsService)
    905         assert.Assert(t, err)
    906
    907         err = prv1Cluster.VanClient.ServiceInterfaceBind(ctx, &http1TlsService, "deployment", "nghttp1tls", map[int]int{}, "")
    908         assert.Assert(t, err)
    909                                                                                                                                                                                                                                          

The first just creates or updates a `ConfigMap` (and lets some other module detect that creation/change and react?), while the second, which's the one that fails, checks for the existence of a secret before creating it.  

Between each of these steps, however, there are a few other network-bound operations (such as other `Get` operations against the K8S cluster).

My guess is that an asynchronous process saw the new ConfigMap from the first step  and created the secret, between check and creation of secret on the second step.

This would be a test code issue, however I was able to reproduce it using the cli (see below)

# How to reproduce

I've run the following loop on a CRC 4.12 running locally.   That part is key, as the issue seems to depend on how fast one process or the other takes to complete.

    #!/usr/bin/sh -xv
    #
    # Reproducer for error
    #
    #   http.go:885: assertion failed: error is not nil: secrets "skupper-tls-nghttp2tls" already exists
    #
    # on testHttps
    
    skupper init
    oc create deploy hello --image quay.io/skupper/hello-world-backend
    
    i=0
    
    while true
    do
      i=$(( i+1 ))
      echo Try $i
      skupper service create hello 8080 --generate-tls-secrets
      skupper service bind hello deployment hello || { exit 1; }
      skupper service delete hello
    done

On my machine, the error was reproduced after 74 iterations:

    (...)
    + true
    + i=74
    + echo Try 74
    Try 74
    + skupper service create hello 8080 --generate-tls-secrets
    + skupper service bind hello deployment hello
    Secret already exists:  skupper-tls-hello
    Error: secrets "skupper-tls-hello" already exists
    + exit 1
    
# Expected behavior

[...]

PR #1232, opened 2023-09-07
More testHttp flaky test fixes

hash-d:
Two proposed fixes:

- for `the object has been modified`, we wait for the secret `skupper-internal` to be updated after an exposed deployment is updated, to avoid a race between an asynchronous process triggered by the deployment update, and the rest of the test's code
- for `secret already exists`, we wait for the secret to be created before we bind the service.  This is akin to what was done on #1226 for `TestTls`, and it is required due to #1208

### DIFF
--- a/test/integration/examples/http/http.go
+++ b/test/integration/examples/http/http.go
@@ -971,7 +971,7 @@ func setup(ctx context.Context, t *testing.T, r base.ClusterTestRunner) {
 	assert.Assert(t, err)
 
 	//update tls service with cert files
-	_, err = prv1Cluster.VanClient.KubeClient.AppsV1().Deployments(prv1Cluster.Namespace).Update(context.TODO(), nghttp2tcpTlsDepWithCertFiles, metav1.UpdateOptions{})
+	err = updateDeployment(prv1Cluster, ctx, nghttp2tcpTlsDepWithCertFiles)
 	assert.Assert(t, err)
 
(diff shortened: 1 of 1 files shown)


test/integration/examples/http/http.go
```diff
@@ -650,6 +653,73 @@ var h1HeyBaseJob = &batchv1.Job{
...
+			_, err := cctx.VanClient.KubeClient.CoreV1().Secrets(cctx.Namespace).Get(
+				context.Background(),
+				secretName,
+				metav1.GetOptions{},
+			)
+			if err == nil {
+				return true, nil
+			}
+			if errors.IsNotFound(err) {
+				fmt.Printf("TLS Secret %q not yet created.  See #1208\n", secretName)
```
nluaces, line 678: Not sure about printing the number of a github issue in the logs, we can link this pull request to the issue for traceability, increase the priority, etc.

hash-d: @nluaces, I'm good to remove it, if you think that's more appropriate.  The reason I put it there, though, is to emphasize that that part of the code is temporary, and should be removed once #1208 is resolved.   I did the same thing on #1229.

nluaces: If you have done this for other pull requests, it is also fine for me, this is not really important. Thanks.

granzoto (approved):
LGTM

PR #1243, opened 2023-09-29
add envars to emit complete flow records

ajssmith:
To reconcile proxies properly in the collector. Also check whether host is ip address or hostname in order to correlate to connectors.

### DIFF
--- a/pkg/kube/deployments.go
+++ b/pkg/kube/deployments.go
@@ -194,10 +194,6 @@ func NewProxyStatefulSet(image types.ImageDetails, serviceInterface types.Servic
 									Value: "json",
 								},
 								{
-									Name:  "APPLICATION_NAME",
-									Value: "skupper-router",
-								},
-								{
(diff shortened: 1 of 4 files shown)


fgiorgetti (approved):

pkg/kube/deployments.go
```diff
@@ -187,7 +194,15 @@ func NewProxyStatefulSet(image types.ImageDetails, serviceInterface types.Servic
 									Value: "json",
 								},
 								{
-									Name: "NAMESPACE",
+									Name:  "APPLICATION_NAME",
```
grs, line 197: Why do we need the APPLICATION_NAME? (Not saying we don't, just clarifying where it is needed).

grs (approved):

pkg/kube/deployments.go
ajssmith: Agree it is not needed and will remove. It is not a recommended label and it was simply carried over from looking at the transport deployment.

PR #1246, opened 2023-10-02
apply specified labels to all created resources

grs:


ajssmith (approved):
lgtm

Suggest updating help text to indicate labels applied to all resource not just skupper pods:
--labels strings                             Labels to add to skupper pods

grs:
> lgtm
> 
> Suggest updating help text to indicate labels applied to all resource not just skupper pods: --labels strings Labels to add to skupper pods

Yes, I missed that but have fixed it now.

pkg/kube/secrets.go
```diff
@@ -80,6 +81,7 @@ func NewSecret(cred types.Credential, owner *metav1.OwnerReference, namespace st
 	}
 
 	secret = PrepareNewSecret(cred, caSecret, connectJsonHost)
+	secret.ObjectMeta.Labels = cred.Labels
```
nluaces, line 84: In case of secrets generated by the service controller (enabling TLS with generated credentials) this field will end up empty (https://github.com/skupperproject/skupper/blob/06f42e106da2a307f4853925030286b432e77edc/pkg/kube/qdr/tls_handler.go#L147).

One option would be to get the labels of the CA if there are none specified for the credentials, or to make  `tls_handler.go` functions responsible of that.

fgiorgetti:
It looks good to me. I noticed that services created by skupper, as well as the `-proxy` statefulsets (for headless
services), are not being labeled. Is this intentional?

fgiorgetti:
One more question. If a token is created, should the token be labeled with the local site labels or should it be labeled
with the destination site labels (where link created is executed)?

pkg/kube/secrets.go
grs: I've added a commit that adds the site labels to generate service tls secrets.

grs:
Having though further on this, I think that for now at least, leaving resources that are created for specific services 

> It looks good to me. I noticed that services created by skupper, as well as the `-proxy` statefulsets (for headless services), are not being labeled. Is this intentional?

The -proxy stateful sets do get the labels, though indirectly, as they include the labels applied to the router deployment (which gets them).

For services, there is already a way of specifying the labels to be applied so I did not add anything extra there. Labelling becomes complicated quite quickly. This is certainly not a complete solution, but the aim is to help with the case where certain labels must be present in order for the resources to be allowed.

grs:
> One more question. If a token is created, should the token be labeled with the local site labels or should it be labeled with the destination site labels (where link created is executed)?

Tokens can be manually labelled before being applied if needed. Again, I acknowledge this is not a comnsistent or comprehensive solution for labeling, just a pragmatic improvement.

fgiorgetti (approved):
Since services can already be labeled using the CLI and the token is a resource that can be easily annotated, I agree that keeping it simple is the best approach.

PR #1248, opened 2023-10-05
Logging extra policy events

fgiorgetti:
Recording extra policy related events for:

* Unexpected error conditions
* Policy state changes (enabled or disabled)
* Allow links only logged when appropriate (reduce noise)
* Service definition removal event
* Only logging service definition removal when it is still defined 

These new events might be helpful to identify policy related issues.


grs (approved):
:shipit:

PR #1249, opened 2023-10-06
fix update

grs:


fgiorgetti (approved):
I tested the update procedure using a site-controller watching all namespaces.
LGTM.

PR #1253, opened 2023-10-24
update resources for older versions than 1.0.2

### LINKED ISSUE #791: Skupper update does not re-write secrets
I picked up version 1.0.2 so I could use the fix to #785.  Running `skupper update` was successful in updating the router and the service controller, but the secrets (specifically skupper-local-client) were not re-generated.  Therefore, the bug in 1.0.1 remained in effect after the update to 1.0.2.

nluaces:
Fixes #791

### DIFF
--- a/client/router_update.go
+++ b/client/router_update.go
@@ -138,7 +138,6 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 			}
 			if utils.LessRecentThanVersion(originalVersion, "1.3.0") {
 				substituteFlowCollector = true
-				updateSecretsFor102 = utils.LessRecentThanVersion(originalVersion, "1.0.2")
 			}
 			if utils.LessRecentThanVersion(originalVersion, "1.4.0") {
 				addPrometheusServer = true
--- a/client/router_update.go
+++ b/client/router_update.go
@@ -94,6 +94,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 		return false, fmt.Errorf("Site (%s) is newer than library (%s); cannot update", site.Version, version.Version)
 	}
 	renameFor050 := false
+	updateSecretsFor102 := false
 	addClaimsSupport := false
 	addMultiportServices := false
 	addClusterPolicy := false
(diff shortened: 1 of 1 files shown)


client/router_update.go
```diff
@@ -114,6 +115,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 		substituteFlowCollector = utils.LessRecentThanVersion(originalVersion, "1.3.0")
 		addPrometheusServer = utils.LessRecentThanVersion(originalVersion, "1.4.0")
 		moveClaims = utils.LessRecentThanVersion(originalVersion, "1.5.0")
+		updateSecretsFor102 = utils.LessRecentThanVersion(originalVersion, "1.0.2")
```
fgiorgetti, line 118: I am not sure this will be enough. In example, if a site was updated from 1.0.1
to 1.0.2, the connect.json will still have the incorrect host name. So eventually
newer versions (>=1.0.2) might have that same issue as well.

I believe to fix this, we might need an excluvise validation inside the connect.json
from skupper-local-client secret to detect whether the host name is still incorrectly
pointing to `skupper-local-client.<namespace>.svc.cluster.local`, if so, then the
`host` key within the connect.json needs to be updated.

The skupper-router-local TLS certificate itself, does not need to be recreated (I suppose),
as the CN is already being updated.

grs: Though I suspect the likelihood of hitting this for anyone who cares is low (and could be mitigated by a manual fix), it might be as clear and easy to just make the check as Fernando suggests.

nluaces: >  if so, then the host key within the connect.json needs to be updated.

@fgiorgetti does the host key need to be updated to an empty value, like it happens for older versions than 0.5?

nluaces: updated in dc5caa56e104f08bc63868a1ac55c0b778a5b3ce

nluaces:
waiting for #1257

client/router_update.go
```diff
@@ -135,6 +138,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 			}
 			if utils.LessRecentThanVersion(originalVersion, "1.3.0") {
 				substituteFlowCollector = true
+				updateSecretsFor102 = utils.LessRecentThanVersion(originalVersion, "1.0.2")
```
grs, line 141: What is the relevance of 1.3.0 here?

grs (approved):

client/router_update.go
nluaces: You are right, I will change this

fgiorgetti (approved):

ajssmith (approved):

PR #1254, opened 2023-10-25
update golang version to 1.20

nluaces:
Also removed network status from the cluster tests given that is going to be obsolete soon.


fgiorgetti (approved):

grs (approved):
:shipit:

PR #1255, opened 2023-10-30
Compress the collector response

bartoval:
We are not currently compressing server responses.

gzip

This serves as an illustrative instance of payload compression applied to the 'process' API. 

The response should include a 'Content-Encoding: gzip' header. param, when it is enabled

PS: It's important to note that the response generated by this API may be influenced by the need to correct data within the 'addresses' field. Nevertheless, it presents an excellent opportunity to assess the benefits of compression.

### DIFF
--- a/cmd/flow-collector/main.go
+++ b/cmd/flow-collector/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


cmd/flow-collector/main.go
```diff
@@ -445,7 +474,7 @@ func main() {
 	log.Printf("COLLECTOR: server listening on %s", addr)
 	s := &http.Server{
 		Addr:    addr,
-		Handler: mux,
+		Handler: enableGzipMiddleware(mux),
```
fgiorgetti, line 477: @bartoval I believe this could be simplified by using: `Handler: handlers.CompressHandler(mux)` instead.
You'd also need this extra import: `"github.com/gorilla/handlers"`.

Let me know if it works for you.

bartoval: Thank you for your suggestion @fgiorgetti . Much better now 🕺

fgiorgetti (approved):

ajssmith (approved):

PR #1256, opened 2023-11-02
Name of cert token based on site UID

### LINKED ISSUE #1154: Tokens created with type cert have a static name
If a token is created using `skupper token create --token-type cert my-token.yaml`,
skupper is setting the `Secret` name to `skupper`. With that if you generate multiple
secrets, for different clusters, and apply them all to a given namespace, only one link
will be defined, as secret `skupper` will be replaced.

It would be good to use a distinct secret name, probably based on the site-id.

DendoD96:
Try to resolve #1154 adding site UID to the token secret name.

### DIFF
--- a/client/connector_token_create.go
+++ b/client/connector_token_create.go
@@ -23,6 +23,19 @@ func (cli *VanClient) checkNotEdgeAndGetVersion(ctx context.Context, namespace s
 	return current.GetSiteMetadata().Version, nil
 }
 
+func (cli *VanClient) getSiteConfig(ctx context.Context, namespace string) (*types.SiteConfig, error) {
+	// get the host and port for inter-router and edge
+	siteConfig, err := cli.SiteConfigInspect(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	if siteConfig == nil {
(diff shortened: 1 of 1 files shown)


client/connector_token_create.go
```diff
@@ -125,7 +133,11 @@ func (cli *VanClient) ConnectorTokenCreateFile(ctx context.Context, subject stri
 	if !res.Allowed {
 		return res.Err()
 	}
-	secret, localOnly, err := cli.ConnectorTokenCreate(ctx, subject, "")
+	siteConfig, err := cli.getSiteConfig(ctx, cli.Namespace)
+	if err != nil {
+		return err
+	}
+	secret, localOnly, err := cli.ConnectorTokenCreate(ctx, fmt.Sprintf("%s-%s", subject, siteConfig.Reference.UID), "")
```
grs, line 140: Ideally we would separate the name used in the secret yaml from the subject used in the client certificate within that secret.

DendoD96: LGTM. I can move the concatenation of subject and site UID into `ConnectorTokenCreate`, but the function is called at several points in the code (e.g. `newGateway`  and `GatewayGenerateBundle`). The alternative could be to change the signature of the function.

PR #1260, opened 2023-11-06
User and Logout APIs for Auth basic and Openshift OAuth Proxy

bartoval:
**Note**: Logout is not part of the Basic Auth model, but it does not pose any issues in our case.

To the the behaviour
1) call the /user api
2) call the /logout
3) back to the user /api and re-authenticate 
OPENSHIFT AUTH

https://skupper-sk-openshift-login.skupper-4-153f1de160110098c1928a6c05e19444-0000.au-syd.containers.appdomain.cloud/api/v1alpha1/user/
https://skupper-sk-openshift-login.skupper-4-153f1de160110098c1928a6c05e19444-0000.au-syd.containers.appdomain.cloud/api/v1alpha1/logout/

AUTH BASIC: user: admin - password: admin

https://skupper-sk-internal-login.skupper-4-153f1de160110098c1928a6c05e19444-0000.au-syd.containers.appdomain.cloud/api/v1alpha1/user/
https://skupper-sk-internal-login.skupper-4-153f1de160110098c1928a6c05e19444-0000.au-syd.containers.appdomain.cloud/api/v1alpha1/logout/

### DIFF
--- a/cmd/flow-collector/main.go
+++ b/cmd/flow-collector/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -45,6 +46,11 @@ type connectJson struct {
 	Host   string    `json:"host,omitempty"`
 	Port   string    `json:"port,omitempty"`
 	Tls    tlsConfig `json:"tls,omitempty"`
+}
+
+type UserResponse struct {
+	Username string `json:"username"`
+	AuthMode string `json:"authType"`
 }
 
 var onlyOneSignalHandler = make(chan struct{})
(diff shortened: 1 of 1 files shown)


cmd/flow-collector/main.go
```diff
@@ -176,6 +237,7 @@ func main() {
 
 		flowRecordTtl = siteConfig.Spec.FlowCollector.FlowRecordTtl
 		enableConsole = siteConfig.Spec.EnableConsole
+		authMode = siteConfig.Spec.AuthMode
```
bartoval, line 240: I'm uncertain whether placing this statement here is appropriate. It involves an if/else structure, and within the else block, there is Podman-related code. @fgiorgetti

cmd/flow-collector/main.go
```diff
@@ -283,6 +355,33 @@ func main() {
 		w.WriteHeader(http.StatusNotFound)
 	}))
 
+	var userApi = api1.PathPrefix("/user").Subrouter()
+	userApi.StrictSlash(true)
+	userApi.HandleFunc("/", authenticated(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bartoval, line 360: Differently from the other functions, here I don't reach 'handler.go' as I don't need to use the Flow.Collector @ajssmith

bartoval (commented):
I was thinking that this APIs could also be implemented by the new developer as an onboarding task. But I have no idea when he might start it. @ajssmith @fgiorgetti

cmd/flow-collector/main.go
```diff
@@ -222,6 +284,16 @@ func main() {
 	}
 	c.FlowCollector.Collector.PrometheusUrl = prometheusUrl
 
+	// map the authentication mode with the function to get the user
+	userMap := make(map[string]func(*http.Request) UserResponse)
+	userMap[string(types.ConsoleAuthModeOpenshift)] = getOpenshiftUser
```
bartoval, line 289: too much Javascript style?   🤔

cmd/flow-collector/main.go
fgiorgetti: You can add the following block of code into the else part (podman related) and it will get the correct auth mode:

```
		authMode = types.ConsoleAuthModeUnsecured
		if flowUsers != "" {
			authMode = types.ConsoleAuthModeInternal
		}
```

fgiorgetti:
@bartoval I was wondering why not simply allowing "/logout" to unauthenticated users?

bartoval:
> @bartoval I was wondering why not simply allowing "/logout" to unauthenticated users?

Unauthenticated user cannot logout because they are "not logged in"
IMO we should remove the logout and username for the Auth Basic because we already have a logout when we close the browser.  It is stateless

ajssmith (approved):
lgtm

PR #1277, opened 2023-11-10
replace output message when the status info is not available yet

nluaces:


grs (approved):

grs (approved):

fgiorgetti (approved):
LGTM. It would be nice to have a similar behavior on skupper status command.

PR #1301, opened 2023-12-01
handle error from spinner

nluaces:


cmd/skupper/skupper_kube_site.go
```diff
@@ -115,7 +114,7 @@ func (s *SkupperKubeSite) Create(cmd *cobra.Command, args []string) error {
 	})
 
 	if err != nil {
-		return err
+		fmt.Println("Skupper status is not loaded yet, it will be available in a few seconds")
```
fgiorgetti, line 117: @nluaces This approach of displaying a message here, instead of returning an error, makes sense to me.
I am not sure if the message should say "it will be available in a few seconds".
Thoughts: @grs @c-kruse @ajssmith ?

grs: I agree; I would leave out that part.

nluaces: @fgiorgetti how about just "Skupper status is not loaded yet"?

fgiorgetti: Looks better to me. Thank you @nluaces.

fgiorgetti (approved):

PR #1305, opened 2023-12-04
superfluous response.WriteHeader warning for Gzip

bartoval:
Roll back to the previous implementation of gzip to avoid the warning message triggered by the Gorilla Mux library.

### DIFF
--- a/cmd/flow-collector/main.go
+++ b/cmd/flow-collector/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"compress/gzip"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -15,10 +14,10 @@ import (
 	"os/signal"
 	"path"
 	"strconv"
-	"strings"
 	"syscall"
 	"time"
 
+	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -47,11 +46,6 @@ type connectJson struct {
 	Host   string    `json:"host,omitempty"`
 	Port   string    `json:"port,omitempty"`
 	Tls    tlsConfig `json:"tls,omitempty"`
-}
-
-type gzipResponseWriter struct {
-	http.ResponseWriter
-	Writer *gzip.Writer
 }
 
 type UserResponse struct {
@@ -144,25 +138,6 @@ func authenticated(h http.HandlerFunc) http.HandlerFunc {
 	} else {
 		return h
 	}
-}
-
-func enableGzipMiddleware(next http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
-			w.Header().Set("Content-Encoding", "gzip")
-
-			gz := gzip.NewWriter(w)
-			defer gz.Close()
-
-			next.ServeHTTP(&gzipResponseWriter{ResponseWriter: w, Writer: gz}, r)
-		} else {
-			next.ServeHTTP(w, r)
-		}
-	})
-}
-
-func (grw *gzipResponseWriter) Write(b []byte) (int, error) {
-	return grw.Writer.Write(b)
 }
 
 func getOpenshiftUser(r *http.Request) UserResponse {
@@ -590,7 +565,7 @@ func main() {
 	log.Printf("COLLECTOR: server listening on %s", addr)
 	s := &http.Server{
 		Addr:    addr,
-		Handler: enableGzipMiddleware(mux),
+		Handler: handlers.CompressHandler(mux),
 	}
 
 	go func() {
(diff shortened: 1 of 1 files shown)


cmd/flow-collector/main.go
```diff
@@ -383,18 +408,15 @@ func main() {
...
 		if exists {
 			userResponse := handler(r)
 			if response, err := json.Marshal(userResponse); err == nil {
-				statusCode = http.StatusOK
 				fmt.Fprintf(w, "%s", response)
 			}
 		}
 
 		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(statusCode)
```
bartoval, line 397: I also received the same warning also for this statement

c-kruse:
@bartoval curious about what the warnings are you are seeing are!

bartoval:
> @bartoval curious about what the warnings are you are seeing are!

Flow collector logs
p

c-kruse (commented):
@bartoval  I had a hard time reproducing this, but I think you found the root of the issue on line 397 (calling http.ResponseWriter.WriteHeader after having already written the response on line 392.) I would be terribly surprised if the gorilla gzip handler was doing something wrong here - it appears to me that it just gets the blame because it wraps the ResponseWriter.

bartoval:
> @bartoval I had a hard time reproducing this, but I think you found the root of the issue on line 397 (calling http.ResponseWriter.WriteHeader after having already written the response on line 392.) I would be terribly surprised if the gorilla gzip handler was doing something wrong here - it appears to me that it just gets the blame because it wraps the ResponseWriter.

👍🏼  You are right. 
The only API affected by this warning is /user. It was like using an axe for a surgical operation.

At this point, however, the doubt is: what is the difference between this API and the others? All of them explicitly call 'responseWrite' in the controller.

For example, this one: 
`var siteApi = api1.PathPrefix("/sites").Subrouter()
	siteApi.StrictSlash(true)
	**siteApi.HandleFunc("/", authenticated(http.HandlerFunc(c.siteHandler))).Name("list")**`

that call this function:

`func (c *Controller) siteHandler(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Content-Type", "application/json")
	c.FlowCollector.Request <- flow.ApiRequest{RecordType: flow.Site, Request: r}
	response := <-c.FlowCollector.Response
	**w.WriteHeader(response.Status)**
	if response.Body != nil {
		fmt.Fprintf(w, "%s", *response.Body)
	}
}`

c-kruse:
>At this point, however, the doubt is: what is the difference between this API and the others? All of them explicitly call 'responseWrite'.

The other API handlers write headers first, then they write the body - it's the order that was the problem with /users. When we call fmt.Fprintf(w, ) we are really just indirectly calling http.ResponseWriter.Write. The go docs for that function try to describe it. Basically go expects you to have set any status and additional headers before you begin to write the body, otherwise it will make some guesses quickly to start sending the response headers and body.
```
~ » go doc net/http.ResponseWriter.Write
package http // import "net/http"

type ResponseWriter interface {

        // Write writes the data to the connection as part of an HTTP reply.
        //
        // If WriteHeader has not yet been called, Write calls
        // WriteHeader(http.StatusOK) before writing the data. If the Header
        // does not contain a Content-Type line, Write adds a Content-Type set
        // to the result of passing the initial 512 bytes of written data to
        // DetectContentType. Additionally...
```

So preserving the original intended behavior, I'd probably do something more like this:

if !exists {
  w.WriteHeader(204)
  return
}
response, err = unmarshal resposne
if err != nil { // maybe log and 500? we shouldn't really ever get there
  return
}
w.Header().Set("Content-Type", "application/json")
w.Write(response)

bartoval:
> > At this point, however, the doubt is: what is the difference between this API and the others? All of them explicitly call 'responseWrite'.
> 
> The other API handlers write headers first, then they write the body - it's the order that was the problem with /users. When we call fmt.Fprintf(w, ) we are really just indirectly calling http.ResponseWriter.Write. The go docs for that function try to describe it. Basically go expects you to have set any status and additional headers before you begin to write the body, otherwise it will make some guesses quickly to start sending the response headers and body.
> 
> ```
> ~ » go doc net/http.ResponseWriter.Write
> package http // import "net/http"
> 
> type ResponseWriter interface {
> 
>         // Write writes the data to the connection as part of an HTTP reply.
>         //
>         // If WriteHeader has not yet been called, Write calls
>         // WriteHeader(http.StatusOK) before writing the data. If the Header
>         // does not contain a Content-Type line, Write adds a Content-Type set
>         // to the result of passing the initial 512 bytes of written data to
>         // DetectContentType. Additionally...
>        Write([]byte) (int, error)
> }
> ```
> 
> So preserving the original intended behavior, I'd probably do something more like this:
> 
> if !exists { w.WriteHeader(204) return } response, err = unmarshal resposne if err != nil { // maybe log and 500? we shouldn't really ever get there return } w.Header().Set("Content-Type", "application/json") w.Write(response)

Perfect. Updated the PR.
Thanks for the clarification 👍🏼

c-kruse (approved):
Looks good to me! Thank you.

PR #1306, opened 2023-12-04
Fix the output in status commands when headless services are exposed in skupper

nluaces:
Context: 
When a headless service is exposed, Skupper creates statefulsets with a pod of the router image that includes a listener or a connector, separated from the skupper-router. 

Fixed issues: 
- The service status command didn't show the targets of the headless services because the command was just checking the skupper-router.
- The link status command, when getting all the direct connections from the router,  was querying the first ready pod with the label "skupper.io/component=router". This could return sometimes the pods from the statefulsets that have the same label; and because they did not have the expected connectors (links), the command show incorrectly that the links where not connected. For the same reason, the skupper version command, some times was getting the wrong router component and it was not showing the versions of the router and the config-sync.
- The skupper status command, was getting information about the linked sites from a statefulset-router, for that reason the output was showing sites connected indirectly (and incorrectly).
- For the same reason as before, the skupper network status command was not showing always the linked sites.

### DIFF
--- a/cmd/skupper/skupper_kube_network.go
+++ b/cmd/skupper/skupper_kube_network.go
@@ -93,7 +93,7 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 							routerId := strings.Split(routerStatus.Router.Name, "/")
 
 							// skip routers that belong to headless services
-							if strings.HasPrefix(routerId[1], siteStatus.Site.Name) {
+							if len(routerId) > 1 && strings.HasPrefix(routerId[1], siteStatus.Site.Name) {
 								routerItem := fmt.Sprintf("name: %s\n", routerId[1])
 								detailsRouter := map[string]string{"image name": routerStatus.Router.ImageName, "image version": routerStatus.Router.ImageVersion}
 
@@ -110,7 +110,6 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 											detailsLink["cost"] = strconv.FormatUint(link.LinkCost, 10)
 										}
 										links.NewChildWithDetail(linkItem, detailsLink)
-
 									}
 								}
 							}
--- a/cmd/skupper/skupper_kube_network.go
+++ b/cmd/skupper/skupper_kube_network.go
@@ -73,8 +73,17 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 
 				if len(siteStatus.RouterStatus) > 0 {
 
-					//to get the generic information about the links of a site, we can get the first router, because in case of multiple routers the information will be the same.
-					mapSiteLink := statusManager.GetSiteLinkMapPerRouter(&siteStatus.RouterStatus[0], &siteStatus.Site)
+					validRouterIndex := 0
+					for index, router := range siteStatus.RouterStatus {
+						// Ignore routers that belong to statefulsets for headless services and any other router
+						routerId := strings.Split(router.Router.Name, "/")
+						if strings.HasPrefix(routerId[1], router.Router.Namespace) {
+							validRouterIndex = index
+							break
+						}
+					}
+
+					mapSiteLink := statusManager.GetSiteLinkMapPerRouter(&siteStatus.RouterStatus[validRouterIndex], &siteStatus.Site)
 
 					if len(mapSiteLink) > 0 {
 						siteLinks := siteLevel.NewChild("Linked sites:")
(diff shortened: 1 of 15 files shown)


cmd/skupper/skupper_kube_network.go
```diff
@@ -73,8 +73,17 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 
 				if len(siteStatus.RouterStatus) > 0 {
 
-					//to get the generic information about the links of a site, we can get the first router, because in case of multiple routers the information will be the same.
-					mapSiteLink := statusManager.GetSiteLinkMapPerRouter(&siteStatus.RouterStatus[0], &siteStatus.Site)
+					validRouterIndex := 0
+					for index, router := range siteStatus.RouterStatus {
+						// Ignore routers that belong to statefulsets for headless services and any other router
+						routerId := strings.Split(router.Router.Name, "/")
+						if strings.HasPrefix(routerId[1], router.Router.Namespace) {
```
grs, line 80: I think this test only works when the site-name is not set?

pkg/network/network.go
```diff
@@ -17,39 +17,53 @@ func (s *SkupperStatus) GetServiceSitesMap() map[string][]SiteStatusInfo {
...
-					serviceSites = append(serviceSites, site)
-					mapServiceSites[listener.Name] = serviceSites
-				} else {
-					mapServiceSites[listener.Name] = []SiteStatusInfo{site}
+			for _, router := range site.RouterStatus {
+				for _, listener := range router.Listeners {
+					if len(mapServiceSites[listener.Address]) == 0 {
+						mapServiceSites[listener.Address] = []SiteStatusInfo{site}
+					} else if mapServiceSites[listener.Address] != nil && !sliceContainsSite(mapServiceSites[listener.Address], site) {
+						serviceSites := mapServiceSites[listener.Address]
```
ajssmith, line 25: iiuc suggest:
mapServiceSites[listener.Address] = append(mapServiceSites[listener.Address], site)

pkg/network/network.go
```diff
@@ -59,7 +73,11 @@ func (s *SkupperStatus) GetRouterSiteMap() map[string]SiteStatusInfo {
 			for _, routerStatus := range siteStatus.RouterStatus {
 				// the name of the router has a "0/" as a prefix that it is needed to remove
 				routerName := strings.Split(routerStatus.Router.Name, "/")
-				mapRouterSite[routerName[1]] = siteStatus
+
+				// Remove routers that belong to statefulsets for headless services
+				if strings.HasPrefix(routerName[1], siteStatus.Site.Namespace) {
```
fgiorgetti, line 78: same here...

cmd/skupper/skupper_kube_service.go
```diff
@@ -91,7 +87,7 @@ func (s *SkupperKubeService) Status(cmd *cobra.Command, args []string) error {
 						theSite := sites.NewChildWithDetail(item, map[string]string{"policy": site.Site.Policy})
 
 						if si.ConnectorCount > 0 {
-							t := mapSiteTargets[site.Site.Identity][si.Name]
+							t := mapSiteTarget[site.Site.Identity][si.Name]
```
fgiorgetti, line 90: When I run `skupper service status -v` on the namespace where my headless service has been exposed from (site1), it is not showing the local target.

But if I run it on a linked namespace (site2), then it shows that the service on `site1` provide the respective targets.

pkg/network/network.go
nluaces: It is indeed more readable now. Thanks! Updated in f2f23f1b0cde2c96ab96ffc027480a1c05f1899b

cmd/skupper/skupper_kube_network.go
```diff
@@ -87,26 +96,29 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 						routers := siteLevel.NewChild("Routers:")
 						for _, routerStatus := range siteStatus.RouterStatus {
 							routerId := strings.Split(routerStatus.Router.Name, "/")
-							routerItem := fmt.Sprintf("name: %s\n", routerId[1])
-							detailsRouter := map[string]string{"image name": routerStatus.Router.ImageName, "image version": routerStatus.Router.ImageVersion}
 
-							routerLevel := routers.NewChildWithDetail(routerItem, detailsRouter)
+							// skip routers that belong to headless services
+							if strings.HasPrefix(routerId[1], siteStatus.Site.Namespace) {
```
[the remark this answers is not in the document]
nluaces: Fixed in f2f23f1b0cde2c96ab96ffc027480a1c05f1899b

cmd/skupper/skupper_kube_service.go
nluaces: Nice catch! I could reproduce it and it is fixed in f2f23f1b0cde2c96ab96ffc027480a1c05f1899b Thanks!

pkg/network/network.go
nluaces: Fixed in f2f23f1b0cde2c96ab96ffc027480a1c05f1899b

grs (approved):

fgiorgetti (approved):

pkg/network/network.go
```diff
@@ -104,6 +117,20 @@ func (s *SkupperStatus) LinkBelongsToSameSite(linkName string, siteId string, ro
 
 }
 
+func (s *SkupperStatus) GetRouterIndex(site *SiteStatusInfo) (error, int) {
+
+	for index, router := range site.RouterStatus {
+		// Ignore routers that belong to statefulsets for headless services and any other router
+		routerId := strings.Split(router.Router.Name, "/")
+		if strings.HasPrefix(routerId[1], site.Site.Name) {
```
c-kruse, line 125: [nonblocker]: Accessing a slice index without bounds checking makes me nervous.

Looks like we've been doing things this way for a while now without getting bitten (above and in skupper_kube_network.go as well), and there could be a reason that we _really_ trust the router name data coming from SiteStatusInfo that I am missing.

c-kruse (approved):
:ship:

pkg/network/network.go
nluaces: Good point! totally agree, let me know if I have missed any of them. Thanks!

Fixed in f769553f34944b571b0d33582b2a222f811e0bcf

ajssmith (approved):
lgtm

PR #1316, opened 2023-12-11
Include a profile option in the flow-collector

c-kruse:
adds a `-profile` flag to the flow-collector that when enabled will expose the routes from the net/http/pprof package over loopback to allow us to better understand the collector's resource utilization.

To use profiling data while running in k8s:

* edit the service controller deployment to run the flow collector with the `-profile` flag set: 
`kubectl patch deployment skupper-service-controller -p '{"spec":{"template":{"spec":{"containers":[{"name":"flow-collector","command":["/app/flow-collector"],"args":["-profile"]}]}}}}'`
* port-forward port 9970 from the deployment to your host: `kubectl port-forward deployments/skupper-service-controller 9970:9970`
* Open `http://localhost:9970/debug/pprof/` in your browser

### DIFF
--- a/cmd/flow-collector/main.go
+++ b/cmd/flow-collector/main.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"path"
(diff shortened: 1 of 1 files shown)


c-kruse:
More of an exploratory tool for me as I start working on the collector. Some interesting findings so far profiling the collector sitting mostly idle with one other site connected, one service running and one console browser session up:

* spending ~15% of our CPU cycles under pkg/flow.getRealSize where we do a gob encoding of our data structures - trying to update the `collector_octets_total` metric in a rather round-about way.
* large majority of heap allocations coming from a double buffering we're doing in the http request handlers.

nluaces (approved):

fgiorgetti (approved):
LGTM

grs (approved):

PR #1324, opened 2023-12-15
fix psa warning

ajssmith:
For k8s 1.24+

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -3,12 +3,13 @@ package client
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


grs (approved):
Looks fine to me. Only question is whether we would ever want to disable the automatic addition of these fields? Or change the values, though the SeccompProfile seems like the only plausible change anyone could require.

fgiorgetti (approved):
Worked fine on my Minikube, OCP 4.15 and OCP 3.11

nluaces (approved):

PR #1326, opened 2023-12-15
Conditionally add routes custom-host resource

fgiorgetti:
* This change ensures that the routes/custom-host resource is only included when the ingress-host is specified
* On OpenShift Local, using the developer account a new site cannot be created after 1.5.1 (still works until 1.5.0), so this gives an alternative to only include that extra rule when needed

### DIFF
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -158,7 +158,7 @@ var ControllerPolicyRule = []rbacv1.PolicyRule{
 	{
 		Verbs:     []string{"get", "list", "watch"},
 		APIGroups: []string{"route.openshift.io"},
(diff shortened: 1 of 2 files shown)


api/types/types.go
```diff
@@ -172,6 +172,14 @@ var ControllerPolicyRule = []rbacv1.PolicyRule{
 	},
 }
 
+var ControllerRouteIngressPolicyRule = []rbacv1.PolicyRule{
```
grs, line 175: The name here doesn't convey that it is a specialised aspect of route that is enabled, namely the explicit setting of the host.

ajssmith (approved):

PR #1327, opened 2023-12-15
Fix races in the Flow Collector

c-kruse:
Lots of reorganization for hopefully zero functional change! :sweat: 

Probably easiest to review commit by commit. Glad to split it up if y'all prefer - they may just seem random without the end state in view.

* Stole the mock messaging.ConnectionFactory from service_sync and put it somewhere multiple packages could use. Getting any useful coverage in this package without some way to send and receive messages was going to be a train wreck. Curious if there's a way to quickly stand up a skupper router isolated per-test-case instead. Bonus points - pkg/service_sync went from like 20 (mostly test-utility related) data races to 3.

* Fixed some obvious race cases in flow/messaging.go shutdown. Also removed a timer in the sender that was doing nothing constructive. Also allowed me to write tests that can wait for the collector to finish.

* Added test coverage and adapted current collector tests to remove racy/flaky behavior.

* Fixed one of the collector data races around shutdown. (more important than it seems - config-sync can loose a leader election and shut down it's collector-lite.)

* ~~Reorganize most of the collector beaconUpdates loop so that it doesn't touch any of the collector's state. Introduces a few channels to communicate with the main loop. This I'm most on-the-fence about. It could instead be a couple line change if we nixed the concurrent beaconUpdates loop and kept everything in the recordUpdates loop. That said, we have no clue how sensitive the send buffers on the other side are - if anything we should be more aggressively reading flow records.~~ Change of heart. Have dropped the concurrent processing of beacons for now in order to solve the race conditions with a minimal changelist. Hope to come back around to reorganize the way our cache is accessed/updated anyhow, esp if we find that a single event handling loop isn't cutting it.


pkg/messaging/memory.go
ajssmith, line 1: suggest filename with functional intent (e.g. mock, broker, etc.)

ajssmith (approved):
Built and ran images, multi-cluster setup, ran tests, etc.

lgtm, nice work

grs (approved):

fgiorgetti (approved):

PR #1328, opened 2023-12-18
Skupper version manifest subcommand

### LINKED ISSUE #1303: Move functionality of `manifest` command into a `skupper` cli subcommand
The information on the hard-coded and on the computed images  used by the `skupper` cli would be very valuable for testing skupper and also for operational use.

This is a request to move the functionality that today exists in `manifest` into a subcommand of `skupper`, so the ability to generate the images manifest is readily available to anyone using skupper.  An alternative would be to have the functionality as an optional/alternate behaviour for the sub command `version`.

For testing, an example of use would be to validate that the images used after an update are as expected.  For operation, that could be used in 'disconnected' scenarios, for pre-pulling the downloads in an automated way.

Today, we can download the manifests from the release pages, but that is not as straightforward as getting the information straight out of the command that is used for all skupper operation.

nluaces:
Fixes #1303 

**Changes:**

- New skupper subcommand included in the version command:  `skupper version manifest`
- It generates a `manifest.json` file with all the images that skupper uses by default and the value of the environment variables for the images if there is any of them. Example:
```
{
   "images": [
      {
         "Name": "quay.io/skupper/skupper-router:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper-router"
      },
      {
         "Name": "quay.io/skupper/service-controller:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/controller-podman:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/config-sync:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/flow-collector:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/skupper/site-controller:main",
         "SHA": "",
         "Repository": "https://github.com/skupperproject/skupper"
      },
      {
         "Name": "quay.io/prometheus/prometheus:v2.42.0",
         "SHA": "",
         "Repository": ""
      },
      {
         "Name": "quay.io/openshift/origin-oauth-proxy:4.14.0",
         "SHA": "",
         "Repository": ""
      }
   ],
   "variables": {
      "QDROUTERD_IMAGE": "quay.io/skupper/skupper-router:latest"
   }
}
```
- There are no changes for the manifest command (external to skupper) that is being used in the release build.

### DIFF
--- a/cmd/manifest/manifest.go
+++ b/cmd/manifest/manifest.go
@@ -1,88 +1,12 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
+	"github.com/skupperproject/skupper/pkg/utils/configs"
 	"os"
 
-	"github.com/skupperproject/skupper/pkg/images"
 	"github.com/spf13/cobra"
 )
-
-type SkupperImage struct {
-	Name       string `yaml:"name"`
-	SHA        string `yaml:"sha"`
-	Repository string `yaml:"repository,omitempty"`
-}
-
-func generateManifestFile() error {
-	// Define a struct for the manifest file.
-	manifest := struct {
-		Images []SkupperImage `json:"images"`
-	}{
-		Images: []SkupperImage{
-			{
-				Name:       images.GetRouterImageName(),
-				SHA:        images.GetSha(images.GetRouterImageName()),
-				Repository: "https://github.com/skupperproject/skupper-router",
-			},
-			{
-				Name:       images.GetServiceControllerImageName(),
-				SHA:        images.GetSha(images.GetServiceControllerImageName()),
-				Repository: "https://github.com/skupperproject/skupper",
-			},
-			{
-				Name:       images.GetControllerPodmanImageName(),
-				SHA:        images.GetSha(images.GetControllerPodmanImageName()),
-				Repository: "https://github.com/skupperproject/skupper",
-			},
-			{
-				Name:       images.GetConfigSyncImageName(),
-				SHA:        images.GetSha(images.GetConfigSyncImageName()),
-				Repository: "https://github.com/skupperproject/skupper",
-			},
-			{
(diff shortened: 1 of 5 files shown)


grs (approved):

pkg/utils/configs/manifest.go
```diff
@@ -0,0 +1,208 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/skupperproject/skupper/pkg/images"
+	"os"
+	"strings"
+)
+
+type SkupperImage struct {
+	Name       string `yaml:"name"`
```
ajssmith, line 12: should struct tag be json?

cmd/manifest/manifest.go
```diff
@@ -1,89 +1,13 @@
...
+	"github.com/skupperproject/skupper/pkg/utils/configs"
 	"os"
 
-	"github.com/skupperproject/skupper/pkg/images"
 	"github.com/spf13/cobra"
 )
 
-type SkupperImage struct {
-	Name       string `yaml:"name"`
-	SHA        string `yaml:"sha"`
```
fgiorgetti, line 14: Should we include the omitempty modifier to the SHA as well?

fgiorgetti (approved):
I believe what @ajssmith said makes sense as manifest.json is not being
generated as expected. Other than that, it works fine.

pkg/utils/configs/manifest.go
nluaces: It should definitely, thanks! 

Fixed in 258203f9f1cb0d93e4572d8d8ec1d5a87dc3230d

PR #1329, opened 2023-12-18
Improves errors when pulling images on podman sites

fgiorgetti:
* Fixes libpod responses parsing (impacting proper error messages from being displayed)
* Adds a recommendation when skupper is unable to pull images (as well as display the error accordingly)
* Libpod responses for container exec and container logs are being parsed properly as well

### DIFF
--- a/client/podman/rest.go
+++ b/client/podman/rest.go
@@ -195,29 +195,29 @@ func (r *responseReaderID) ReadResponse(response runtime.ClientResponse, consume
 	}
 }
 
-type responseReaderByteStreamBody struct {
+type multiplexedBodyReader struct {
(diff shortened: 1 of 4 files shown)


client/podman/rest.go
```diff
@@ -194,10 +195,110 @@ func (r *responseReaderID) ReadResponse(response runtime.ClientResponse, consume
 	}
 }
 
-type responseReaderBody struct {
+type responseReaderByteStreamBody struct {
```
grs, line 198: very minor: MultiplexedBodyReader might be a more descriptive name as it isn't obvious how octet stream is different from byte stream

grs (approved):

PR #1338, opened 2023-12-22
Check if site version is compatible with CLI version

### LINKED ISSUE #1281: Incompatibility of 1.5.0 CLI with previous versions of the skupper operator
After using the skupper operator (1.4.3-rh-6) a user got into the following error running 1.5.0 skupper with `skupper status command`
```
$ skupper status
Error: configmaps "skupper-network-status" not found
```
The problem is that the 1.5.0 relies on the existence the configmap `skupper-network-status`, that is not contemplated in previous versions.

Provide more information about what is happening to the user.

nluaces:
Fixes #1281 

### What has changed: 

The following status commands check if the config-sync component (that is the one that contains the flow-collector lite process running) has the minimum version required (1.5.0) to work:
* skupper status
* skupper network status
* skupper link status
* skupper service status

When the site version is prior to 1.5.0, the message will be: 

~~Site version is 1.4.4, but CLI requires 1.5.0~~

EDIT: 
> Detected that the skupper version installed in the namespace is version 1.4.4. The CLI requires version 1.5.0. To update the installation, please follow the instructions found here: https://skupper.io/docs/index.html

### Testing:
 
- Installed last version of OLM: https://github.com/operator-framework/operator-lifecycle-manager/releases
- Installed skupper operator version 1.4.4 : https://github.com/skupperproject/skupper-operator/tree/1.4
- Executed all the status commands

### DIFF
--- a/cmd/skupper/skupper_kube_network.go
+++ b/cmd/skupper/skupper_kube_network.go
@@ -40,7 +40,7 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 
 	configSyncVersion := utils.GetVersionTag(s.kube.Cli.GetVersion(types.TransportContainerName, types.ConfigSyncContainerName))
 	if configSyncVersion != "" && !utils.IsValidFor(configSyncVersion, network.MINIMUM_VERSION) {
-		fmt.Printf("Site version is %s, but CLI requires %s", configSyncVersion, network.MINIMUM_VERSION)
+		fmt.Printf(network.MINIMUM_VERSION_MESSAGE, configSyncVersion, network.MINIMUM_VERSION)
 		fmt.Println()
 		return nil
 	}
--- a/cmd/skupper/skupper_kube_network.go
+++ b/cmd/skupper/skupper_kube_network.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/skupperproject/skupper/api/types"
 	"github.com/skupperproject/skupper/pkg/network"
+	"github.com/skupperproject/skupper/pkg/utils"
 	"github.com/skupperproject/skupper/pkg/utils/formatter"
 	"github.com/spf13/cobra"
 	"strconv"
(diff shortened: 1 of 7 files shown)


pkg/utils/version.go
```diff
@@ -97,3 +98,19 @@ func IsValidFor(actual string, minimum string) bool {
 	vb := ParseVersion(minimum)
 	return va.IsUndefined() || !va.LessRecentThan(vb)
 }
+
+func GetVersionFromImageTag(imageTag string) string {
```
grs, line 102: Very minor point, but strictly speaking I believe the argument passed in is an image descriptor comprised of a full image 'url' and a SHA. It is not a 'tag' in the usual sense of the word. A different name for the argument to this function might make it more obvious sometime in the future when someone looks at it in isolation.

grs (approved):

cmd/skupper/skupper_kube_network.go
```diff
@@ -36,6 +37,14 @@ func (s *SkupperKubeNetwork) Status(cmd *cobra.Command, args []string) error {
 		fmt.Println()
 		return nil
 	}
+
+	configSyncVersion := utils.GetVersionTag(s.kube.Cli.GetVersion(types.TransportContainerName, types.ConfigSyncContainerName))
+	if configSyncVersion != "" && !utils.IsValidFor(configSyncVersion, network.MINIMUM_VERSION) {
+		fmt.Printf("Site version is %s, but CLI requires %s", configSyncVersion, network.MINIMUM_VERSION)
```
ajssmith, line 43: Propose something similar to the following to suggest to user action to take:

"Detected that the skupper version installed in the namespace is version %s. The CLI requires version %s. To update the installation, please follow the instructions found here"

where here is link to skupper docs

ajssmith (approved):
lgmt, minor comment to maybe update message returned

cmd/skupper/skupper_kube_network.go
nluaces: Updated in f8197e767e2ae05ed76fd5debeae3bda0bf0bca1

PR #1339, opened 2023-12-23
Running the spinner if skupper is executed in the foreground

### LINKED ISSUE #1320: skupper init should detect when not running in foreground with tty
**Describe the bug**

With 1.5, `skupper init` now shows a helpful `Waiting for status...` message.  However, the message is shown even if the command was run on background.  The result is that the user will have to wait for the `init` command to finish before they can properly use that terminal again, as the message is updated several times each second.

**How To Reproduce**

```
$ skupper init &
Waiting for status...|
Waiting for status.../
Waiting for status.../
Waiting for status...\
Waiting for status...\
Waiting for status...-
Waiting for status...\
Wed 13 Dec 13:14:33 -03 2023
Waiting for status...
Skupper is now installed in namespace 'dh-new'.  Use 'skupper status' to get more information.
```
Notice the different `Waiting for status...` messages above are a product of hitting `ENTER`, and the date is a result of running the `date`  command while the `skupper init` ran in the background.  During that time, it was impossible to see what had been typed at the prompt.

**Expected behavior**

It's normal for commands to print messages while running the background.  However, that status message should be avoided in this case, as it negates the use of the background execution of `skupper init`

**Environment details**
 - Skupper CLI: 1.5.1
 - Platform: any

**Additional context**

nluaces:
Fixes #1320

### DIFF
--- a/pkg/utils/spinner.go
+++ b/pkg/utils/spinner.go
@@ -2,16 +2,20 @@ package utils
 
 import (
 	"github.com/briandowns/spinner"
+	"golang.org/x/sys/unix"
+	"os"
+	"syscall"
 	"time"
 )
 
 func NewSpinner(message string, maxRetries int, function func() error) error {
-
 	spin := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
 	spin.Prefix = message
 	spin.FinalMSG = message + "\n"
 
-	spin.Start()
+	if IsRunningInForegroundOrDefault() {
+		spin.Start()
+	}
 
 	err := RetryError(time.Second, maxRetries, function)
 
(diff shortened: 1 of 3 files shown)


pkg/utils/spinner.go
```diff
@@ -23,3 +27,18 @@ func NewSpinner(message string, maxRetries int, function func() error) error {
 
 	return nil
 }
+
+func IsRunningInForegroundOrDefault() bool {
+	// Get the foreground process group ID of the terminal
+	foregroundPGID, err := unix.IoctlGetInt(int(os.Stdout.Fd()), unix.TIOCGPGRP)
```
c-kruse, line 33: I've never really asked, but what platforms does the skupper CLI run on? The CLI strikes me as something that should maybe be able to run on windows. It at least compiles on main now. I assume this would not compile.

nluaces: @c-kruse After creating this pr, I've seen that the CI didn't pass the windows compilation, so no, this solution is not valid.

c-kruse: @nluaces  :+1: As a side note, I don't think it's totally unreasonable to fix this for only unix-y systems. I don't even know if windows has the same concept of foreground/background.

My suggestion would be to have a spinner_windows.go implementation with IsRunningInForegroundOrDefault that is hardcoded true, then a spinner_unix.go with appropriate build tags with this solution.

nluaces: thanks @c-kruse !

PR #1341, opened 2024-01-03
Fixes phantom process records observed in the console UI

c-kruse:
Only persist new process records when their EndTime is unset. When a message containing a record update setting process EndTime is delivered multiple times, subsequent deliveries should be ignored.


ajssmith (approved):
lgtm, thanks for the fix

bartoval (approved):
kill phantoms 🥇 
Thanks!

PR #1342, opened 2024-01-05
update skupper-local-client secret only if it is necessary 

nluaces:
And independently of the version of origin.


grs (approved):
Change is fine. Ideally I think we would update the secret here rather than deleting and recreating, but that could be a separate improvement.

fgiorgetti (approved):

PR #1346, opened 2024-01-08
Less slow skupper init

c-kruse:
My proposal to improve the experience starting a new skupper site. **Reduces startup time from ~25s to ~15s** *when the wind is blowing the right direction.

* added some logging to help me better understand where we are spending time
* exposed a way for the config-sync process to "Prime" its collector with phony BeaconRecords from where it anticipates the router and service controller to be listening in order to eliminate the time usually spent waiting for a beacon record.
* adjusts the timing service-controller uses to wait for the router to be ready, and the order in which the config-sync startup runs in order to start the controller/collector as soon as possible.
* adjusts the `get` command's http client to prevent it from hanging - observed in some failed tests.

I'd suggest reviewing with some skepticism. Especially regarding my attempts to learn / divine the router and site id from the config-sync container (SKUPPER_SITE_ID not set in environment.) I hit numerous issues in the integration test suites where the slightly adjusted timing caused failures and where my assumptions about the presence of the `skupper-site` CM didn't turn out to be valid.

### DIFF
--- a/cmd/config-sync/collector.go
+++ b/cmd/config-sync/collector.go
@@ -32,52 +32,58 @@ func updateLockOwner(lockname, namespace string, owner *metav1.OwnerReference, c
 	return err
 }
 
-func siteCollector(stopCh <-chan struct{}, cli *client.VanClient) *flow.FlowCollector {
+func siteCollector(stopCh <-chan struct{}, cli *client.VanClient) {
 	var fc *flow.FlowCollector
 	siteData := map[string]string{}
 	platform := config.GetPlatform()
-	if platform == "" || platform == types.PlatformKubernetes {
-		current, err := kube.GetDeployment(types.TransportDeploymentName, cli.Namespace, cli.KubeClient)
-		if err != nil {
-			log.Fatal("Failed to get transport deployment", err.Error())
-		}
-		owner := kube.GetDeploymentOwnerReference(current)
-
-		existing, err := kube.NewConfigMap(types.NetworkStatusConfigMapName, &siteData, nil, nil, &owner, cli.Namespace, cli.KubeClient)
-		if err != nil && existing == nil {
-			log.Fatal("Failed to create site status config map ", err.Error())
-		}
-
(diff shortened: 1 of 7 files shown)


nluaces (approved):
Looks good to me! but I would like to wait for @ajssmith review before merging this PR.

These are the tests using this branch's version of skupper in minikube:

```
~ » time skupper init   
Waiting for LoadBalancer IP or hostname...
Waiting for status...
Skupper is now installed in namespace 'public1'.  Use 'skupper status' to get more information.
skupper init  1.35s user 0.07s system 6% cpu *22.400 total*
------------------------------------------------------------
~ » time skupper init --routers 3                                                                   
Waiting for status...
Skupper is now installed in namespace 'public1'.  Use 'skupper status' to get more information.
skupper init --routers 3  1.34s user 0.10s system 5% cpu *24.717 total*

```

And these are the results using skupper 1.5.2 in minikube:
```
~ » time skupper init                                                                                
Waiting for LoadBalancer IP or hostname...
Waiting for status...
Skupper status is not loaded yet.
Skupper is now installed in namespace 'public1'.  Use 'skupper status' to get more information.
skupper init  1.14s user 0.15s system 2% cpu *55.275 total*

------------------------------
~ » time skupper init --routers 3                                                                    
Waiting for LoadBalancer IP or hostname...
Waiting for status...
Skupper is now installed in namespace 'public1'.  Use 'skupper status' to get more information.
skupper init --routers 3  1.14s user 0.12s system 3% cpu *36.287 total*

```

fgiorgetti (approved):
LGTM. It seems to reduce init time from ~ 35 to ~ 25 seconds (after several attempts).

grs (approved):
I'm ok with these changes. I think as you state in the comments, at some point we should rethink some of the paths, but if this improves things for now then no objection from me.

c-kruse:
@ajssmith - Noe thought you may have some insights in particular!

ajssmith (approved):
lgtm, thanks

PR #1351, opened 2024-01-09
update link for update instructions

pwright:


nluaces (approved):
Looks good to me. 

The origin of the issue was that after installing the skupper operator 1.4.x in the namespace, the user would have installed 1.5.x CLI version in local. 

Should this link also include how to update the skupper-operator version? because in this particular case, the user does not need to update the CLI, they were already using the latest.

What do you think?

pwright:
good point. I'm creating a skupper-website PR for that page and will link to operator docs from there.

pwright:
ood

PR #1356, opened 2024-01-16
Some proposed changes

grs:


doc/adr/0007-deprecate-servicesync.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Status
+
+Proposed
+
+Enabled By [6. Explicit resources for declaring service bindings](0006-explicit-resources-for-declaring-service-bindings.md)
+
+## Context
+
+ServiceSync is a mechanism whereby service ingress bindings are
```
nluaces, line 13: I think these are very good examples of ADRs. It is very useful to have available the links to previous ADRs that affect the following decisions.

PR #1358, opened 2024-01-16
Adr 7

grs:


doc/adr/0007-deprecate-servicesync.md
```diff
@@ -0,0 +1,48 @@
...
+allows more flexibility with regard to the ingress bindings
+ServiceSync would not be adequate as it is.
+
+## Decision
+
+ServiceSync will not be supported for automatic propoagation of
+ingress bindings between sites that have migrated to use explicit
+resources for each ingress binding.
+
+## Consequences
```
ajssmith, line 36: It might be worth noting that existing examples and workshops rely on service synch functionality

doc/adr/0007-deprecate-servicesync.md
```diff
@@ -0,0 +1,48 @@
...
+
+If the decision to have explicit resources for ingress bindings is
+adopted, it becomes even simpler to explicitly add the required
+ingresses to any site that requires them. Indeed as that proposal
+allows more flexibility with regard to the ingress bindings
+ServiceSync would not be adequate as it is.
+
+## Decision
+
+ServiceSync will not be supported for automatic propoagation of
```
ajssmith, line 32: clarify "will not be supported" (e.g. will be removed) as the consequence below indicates re-introducing as a new/separate "van service"

doc/adr/0007-deprecate-servicesync.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Status
+
+Proposed
+
+Enabled By [6. Explicit resources for declaring service bindings](0006-explicit-resources-for-declaring-service-bindings.md)
+
+## Context
+
+ServiceSync is a mechanism whereby service ingress bindings are
```
pwright, line 13: Service synchronisation (ServiceSync) is a feature that creates listeners  on every site 
when a service is exposed. This feature does not require user actions on those sites.

In other words, ServiceSync is a mechanism whereby service ingress bindings are

doc/adr/0007-deprecate-servicesync.md
```diff
@@ -0,0 +1,48 @@
...
+## Decision
+
+ServiceSync will not be supported for automatic propoagation of
+ingress bindings between sites that have migrated to use explicit
+resources for each ingress binding.
+
+## Consequences
+
+This will require extra declarations to be made by user at the sites
+at which they want to consume services. However this also gives much
```
pwright, line 39: link to ADR 8

doc/adr/0007-deprecate-servicesync.md
pwright, line 36: To me, the biggest consequence is the change to the mental model that we currently tell users about in docs.
However, this ADR only addresses the removal of the current model, and ADR 8 hints at the new model without defining it, but how about adding the following:

This decision removes the idea that users can expose a service on one site and then automatically consume that service on all other sites.

grs: Agreed. Done here and for adr-6 where the same applies.

doc/adr/0007-deprecate-servicesync.md
grs: Done.

doc/adr/0007-deprecate-servicesync.md
grs: I believe you mean adr 6? That is linked to above (using the adr tools linking feature).

doc/adr/0007-deprecate-servicesync.md
grs: This ADR only proposes the removal of a current feature. It doesn't in my view propose a different model per se. I agree that this is closely related to the decision proposed in ADR 6, and indeed they are linked together using the adr toolings linking support.

fgiorgetti:
The content looks clear and good to me.
One question. Should we create project milestones and associated "approved" ADRs (PRs/Issues) to be part of them?

nluaces (approved):

fgiorgetti (approved):

ajssmith (approved):

PR #1359, opened 2024-01-16
Adr 8

grs:


doc/adr/0006-explicit-resources-for-declaring-service-bindings.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Status
+
+Proposed
+
+Enables [7. Deprecate ServiceSync](0007-deprecate-servicesync.md)
+
+## Context
+
+Service bindings are the configurations that determine how a Skupper
```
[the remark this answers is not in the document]
grs: Sorry Paul, I just realised from your comment that this PR inadvertently included commits from other PRs. I;ve fixed that now, but your comment applies to #1358 I believe.

Am I right in understanding your point here being that it would be valuable to call out in the consequences that the decision will require extra declaration(s) as compared to 1.x? (If so, I agree, that would be a useful addition and I will make it).

doc/adr/0008-combine-site-controller-and-service-controller.md
```diff
@@ -0,0 +1,54 @@
...
+
+The controller will also support updating the site in line with any
+configuration changes.
+
+Features that require connection to the router will be moved into the
+config-sync container.
+
+## Consequences
+
+Removing an unnecessary component means simpler configuration and
```
fgiorgetti, line 45: I assume that as part of it, the `get` command as well as the `ConsoleServer` are also going away ?

nluaces: @fgiorgetti the absence of the get command affects the current policy manager, right?

grs: Good point! I have updated the consequences to call this out.

fgiorgetti: @nluaces yes, it does. But I believe the overall policy mechanism needs to be revisited.

fgiorgetti (approved):

PR #1369, opened 2024-01-24
Accept multivalue query filters in the Console API

c-kruse:
Treats repeated filter query parameters as a logical OR when filtering. For example GET processes?processRole=remote&processRole=external should return processes where processRole is either remote OR external.

### DIFF
--- a/pkg/flow/record.go
+++ b/pkg/flow/record.go
@@ -474,15 +474,15 @@ type Payload struct {
 }
 
 type QueryParams struct {
-	Offset             int               `json:"offset"`
-	Limit              int               `json:"limit"`
-	SortBy             string            `json:"sortBy"`
-	Filter             string            `json:"filter"`
-	FilterFields       map[string]string `json:"filterFields"`
-	TimeRangeStart     uint64            `json:"timeRangeStart"`
-	TimeRangeEnd       uint64            `json:"timeRangeEnd"`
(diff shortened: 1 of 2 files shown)


c-kruse:
@bartoval not urgent, but was playing around and think this is a relatively quick fix.

After a quick look, I suspect that on the console side we'd need to switch around how axios serializes query parameters, but I didn't see a reason why that wouldn't work.

bartoval:
> @bartoval not urgent, but was playing around and think this is a relatively quick fix.
> 
> After a quick look, I suspect that on the console side we'd need to switch around how axios serializes query parameters, but I didn't see a reason why that wouldn't work.
Thanks! 👍🏼 

I think the serialization should not be a problem. I control it https://github.com/bartoval/skupper-console/blob/main/src/API/REST.utils.ts

bartoval:
> > @bartoval not urgent, but was playing around and think this is a relatively quick fix.
> > After a quick look, I suspect that on the console side we'd need to switch around how axios serializes query parameters, but I didn't see a reason why that wouldn't work.
> 
> Thanks! 👍🏼 I think the serialization should not be a problem. I control it https://github.com/skupperproject/skupper-console/blob/main/src/API/apiMiddleware.ts#L29C5-L29C24

bartoval (approved):

PR #1372, opened 2024-01-26
Fix lack of service_name and other info in legacy /DATA path

grs:
The router now appears to be replying to management queries with the wrong type

for the correlation id. This change checks for the new type and uses that if it can.

This prevents errors when handling /DATA in the rest-api, and ensures that the available information is included in the response.


fgiorgetti (approved):

grs:
This is really a workaround for a change in the router between 1.4.3 and 1.5.0

grs:
This is fixed in the router by https://github.com/skupperproject/skupper-router/issues/1391

PR #1378, opened 2024-01-30
don't try to move claims for edge site on update

grs:
Without this fix, trying tp update a site where router-mode=edge results in an error `Error: services "skupper-router" not found`

### DIFF
--- a/client/router_update.go
+++ b/client/router_update.go
@@ -103,6 +103,7 @@ func (cli *VanClient) RouterUpdateVersionInNamespace(ctx context.Context, hup bo
 	substituteFlowCollector := false
 	addPrometheusServer := false
 	moveClaims := false
+	updateRole := false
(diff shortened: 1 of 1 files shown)


nluaces (approved):

fgiorgetti (commented):
There is a problem here.
The moveClaims is not just doing what it says.

It is also updating skupper-router role, adding extra permissions
needed by the collector lite. And without it, collector lite won't have
permissions to create the skupper-network-status configmap.

I believe we need to introduce an extra variable to control that separately from
the moveClaims.

grs:
Good catch, I'll push an update.

grs:
Fix pushed that ensures the role is updated for config-sync to be able to create the configmap and write the status.

fgiorgetti (approved):
LGTM

PR #1384, opened 2024-02-01
Add cpu and memory limits to podman sites

### LINKED ISSUE #1075: Define a default CPU limit to the skupper-router container on podman sites

fgiorgetti:
Fixes #1075

### DIFF
--- a/client/podman/container_mock.go
+++ b/client/podman/container_mock.go
@@ -284,11 +284,11 @@ func (r *RestClientMock) HandleContainerInspect(operation *runtime.ClientOperati
 			StartedAt:  strfmt.DateTime(c.StartedAt),
 		},
 	}
-	if c.Cpus > 0 || c.MemoryBytes > 0 {
+	if c.MaxCpus > 0 || c.MaxMemoryBytes > 0 {
 		res.Payload.HostConfig = &models.InspectContainerHostConfig{
-			CPUQuota:  int64(c.Cpus * 100000),
+			CPUQuota:  int64(c.MaxCpus * 100000),
 			CPUPeriod: 100000,
-			Memory:    c.MemoryBytes,
+			Memory:    c.MaxMemoryBytes,
 		}
 	}
 	return res, nil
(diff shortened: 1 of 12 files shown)


pkg/domain/podman/site.go
```diff
@@ -846,6 +863,8 @@ func (s *SiteHandler) prepareControllerDeployment(site *Site) *SkupperDeployment
 		endpoint = "/tmp/podman.sock"
 		volumeMounts[sockFile] = endpoint
 	}
+	memoryLimit, _ := strconv.ParseInt(site.ControllerOpts.MemoryLimit, 10, 64)
+	cpus, _ := strconv.Atoi(site.ControllerOpts.CpuLimit)
```
c-kruse, line 867: @fgiorgetti could you help me understand why you are confident to not check these errors?

I do see above you may be setting them yourself e.g. `site.RouterOpts.MemoryLimit = strconv.FormatInt(c.MemoryLimit, 10)`. If that's always the case I'm not sure I understand why we'd want to involve string marshaling/unmarshaling at all.

fgiorgetti: @c-kruse Actually I am missing an early validation of these values, which are validated somewhere else
on kubernetes sites. I will certainly add an early stage validation.

With regards to the other comment, that other validation is not needed, as it is just retrieving what is
currently applied to the container.

Thanks for finding it. I am going to update the PR once the validation is done.

client/podman/container_mock.go
```diff
@@ -350,6 +366,14 @@ func (r *RestClientMock) HandleContainerCreate(operation *runtime.ClientOperatio
 		RestartCount:  int(spec.RestartRetries),
 		CreatedAt:     time.Now(),
 	}
+	if spec.ResourceLimits != nil {
+		if spec.ResourceLimits.CPU != nil {
+			c.Cpus = int(spec.ResourceLimits.CPU.Quota / 100000)
+		}
+		if spec.ResourceLimits.Memory != nil {
+			c.MemoryBytes = spec.ResourceLimits.Memory.Limit
```
grs, line 374: Minor naming: To me, ResourceLimits.Memory.Limit is a MaxMemoryBytes or MemoryLimitBytes. This may just be my lack of familiarity with the podman APIs, or my inference from Kubernetes. MemoryBytes sounds like the current memory rather than a limit.

grs (approved):

client/podman/container_mock.go
fgiorgetti: Agreed. I will update it.

grs (approved):

nluaces (approved):

PR #1385, opened 2024-02-06
Nightly build

nluaces:
- Moves example tests from the normal build to the nightly build. 
- The example tests will be executed at 5:00 UTC / 0:00 EST

### DIFF
--- a/test/integration/examples/custom/helloworld/helloworld_podman_test.go
+++ b/test/integration/examples/custom/helloworld/helloworld_podman_test.go
@@ -1,4 +1,4 @@
-//go:build (podman && cli) || (podman && examples) || (podman && integration)
+//go:build (podman && cli) || (podman && integration)
 
 package helloworld
 
--- a/test/integration/examples/custom/helloworld/helloworld_test.go
+++ b/test/integration/examples/custom/helloworld/helloworld_test.go
@@ -1,5 +1,5 @@
-//go:build integration || cli || examples
-// +build integration cli examples
+//go:build cli || examples
+// +build cli examples
 
 package helloworld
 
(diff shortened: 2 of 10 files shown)


test/integration/examples/custom/helloworld/helloworld_podman_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build (podman && cli) || (podman && examples) || (podman && integration)
+//go:build (podman && cli) || (podman && integration)
```
fgiorgetti, line 1: I believe this one should use: `podman && examples` instead, right?

test/integration/examples/custom/helloworld/helloworld_test.go
```diff
@@ -1,5 +1,5 @@
-//go:build integration || cli || examples
-// +build integration cli examples
+//go:build cli || integration
+// +build cli integration
```
fgiorgetti, line 2: same here... `examples` instead of integration.

test/integration/examples/custom/helloworld/helloworld_podman_test.go
nluaces: I was planning to leave the HelloWorld examples (with and without podman) in the normal build; I think that the podman tests need to be executed just in case in every build. What do you think?

fgiorgetti: Makes sense.
What about moving them out of the test/integration/examples/custom into
test/integration/acceptance/custom ?

nluaces: good idea!

grs (approved):

PR #1388, opened 2024-02-12
status commands with podman collector lite

nluaces:
Status commands adapted to the podman collector lite: 
- skupper network status 
- skupper init
- skupper status
- skupper link status
- skupper service status

I did some refactoring to avoid code duplication. I plan to move all the `print` functions in a common package in further pull requests.

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/skupperproject/skupper/pkg/network"
 	"github.com/skupperproject/skupper/pkg/utils/configs"
 	"os"
 	"reflect"
@@ -85,7 +86,8 @@ type SkupperTokenClient interface {
 }
 
 type SkupperNetworkClient interface {
-	Status(cmd *cobra.Command, args []string) error
+	GetCurrentSite(ctx context.Context) (string, error)
+	Status(cmd *cobra.Command, args []string, ctx context.Context) (*network.NetworkStatusInfo, error)
 	StatusFlags(cmd *cobra.Command)
 	SkupperClientCommon
 }
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -86,7 +86,7 @@ type SkupperTokenClient interface {
 }
 
 type SkupperNetworkClient interface {
-	GetCurrentSite(ctx context.Context) (string, error)
+	GetCurrentSiteId(ctx context.Context) (string, error)
 	Status(cmd *cobra.Command, args []string, ctx context.Context) (*network.NetworkStatusInfo, error)
 	StatusFlags(cmd *cobra.Command)
 	SkupperClientCommon
--- a/cmd/skupper/skupper_kube_network.go
+++ b/cmd/skupper/skupper_kube_network.go
@@ -13,7 +13,7 @@ type SkupperKubeNetwork struct {
 	kube *SkupperKube
 }
 
-func (s *SkupperKubeNetwork) GetCurrentSite(ctx context.Context) (string, error) {
+func (s *SkupperKubeNetwork) GetCurrentSiteId(ctx context.Context) (string, error) {
 
 	siteConfig, err := s.kube.Cli.SiteConfigInspect(ctx, nil)
 	if err != nil || siteConfig == nil {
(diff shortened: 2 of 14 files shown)


cmd/skupper/skupper.go
```diff
@@ -85,7 +86,8 @@ type SkupperTokenClient interface {
 }
 
 type SkupperNetworkClient interface {
-	Status(cmd *cobra.Command, args []string) error
+	GetCurrentSite(ctx context.Context) (string, error)
```
grs, line 89: Minor naming issue, suggest GetCurrentSIteId.

grs (approved):

fgiorgetti (commented):
First of all, great work @nluaces !

I have noticed a few issues with service status, happening on both platforms.

1. If a service has multiple targets it is only displaying one
2. Targets with type "service" are not displayed properly
3. Podman targets are not rendered properly as well

Maybe item 3 could be handled separately?

nluaces:
@fgiorgetti thanks for your feedback!
Could you provide a screenshot of cases 2 and 3? 
What do you mean by handled separately in the case 3?

fgiorgetti:
> @fgiorgetti thanks for your feedback! Could you provide a screenshot of cases 2 and 3? What do you mean by handled separately in the case 3?

Here is what I have done for case 2.

I had two kubernetes services on my namespace: nginx1 and nginx2.
Then I created a skupper service, named: nginx-skupper and bound it to those two services, using:

```
skupper service bind nginx-skupper service nginx1.sk1.svc.cluster.local
skupper service bind nginx-skupper service nginx2.sk1.svc.cluster.local
```

Using 1.5.3, here is what I got:

$ skupper service status -v
Services exposed through Skupper:
╰─ nginx-skupper:8080 (tcp)
   ╰─ Sites:
      ╰─ 362c9f19-29f4-4555-a9a6-c6f575624c8e(sk1)
         │ policy: disabled
         ╰─ Targets:
            ├─ nginx-skupper:8080 
            ╰─ nginx-skupper:8080 


Now using 1.4.4:

$ skupper service status 
Services exposed through Skupper:
╰─ nginx-skupper (tcp port 8080)
   ╰─ Targets:
      ├─ nginx1.sk1.svc.cluster.local name=nginx1.sk1.svc.cluster.local 
      ╰─ nginx2.sk1.svc.cluster.local name=nginx2.sk1.svc.cluster.local 

The target service is no longer displayed with 1.5.x. So this is not a problem related to this PR,
so maybe we should handle it separately. 

Now, on case 3, the issue is similar.
If you create the following service and bind the following hosts:

```
$ skupper --platform podman service create nginx-skupper 8080 --host-port 8888
$ skupper --platform podman service bind nginx-skupper host 10.0.0.1
$ skupper --platform podman service bind nginx-skupper host 10.0.0.2
```

Using 1.5.3, this is the output of `service status`:

Services exposed through Skupper:

╰─ nginx-skupper:8080 (tcp)
   ├─ Host ports:
   │  ╰─ ip: * - ports: 8888 -> 8080
   ╰─ Targets: 
      ├─ host: 10.0.0.1 - ports: map[8080:8080]
      ╰─ host: 10.0.0.2 - ports: map[8080:8080]

Using the updated code, this is the output of `service status -v`:

Services exposed through Skupper:
╰─ nginx-skupper:8080 (tcp)
   ╰─ Sites:
      ╰─ 5161d7b6-33a3-4eee-a029-b3f481603dfb()
         │ policy: disabled
         ╰─ Targets:
            ├─ nginx-skupper:8080 name=nginx-skupper-10.0.0.1
            ╰─ nginx-skupper:8080 name=nginx-skupper-10.0.0.2

Local services, in this case, possibly need special handling.

I believe it should be OK to merge this PR, but then we'd need issues to track
these two issues. WDYT?

nluaces:
Thanks for your clarifications @fgiorgetti !

> The target service is no longer displayed with 1.5.x. So this is not a problem related to this PR,
so maybe we should handle it separately.

I see now, the current target output shows the address and the name of the process, provided by the collector; and in this case because it is a service there is no such data. I addressed this issue in #1393 

> Local services, in this case, possibly need special handling.

I think you are right,  I created this issue #1394 to show the info about local services as it was, and using the collector lite for the rest.

nluaces:
I was thinking on a second possibility regarding #1394: leave the skupper status command with the podman platform as it is in main for now, and in the next iteration use an hybrid output with the information the podman gives (host ports, etc) and the collector-lite, only in podman sites.

What do you think? @fgiorgetti @ajssmith @grs

grs:
Sounds sensible to me.

fgiorgetti:
Sounds good to me either @nluaces. Thank you.

fgiorgetti (approved):
LGTM. Thanks @nluaces for all your work and attention.

c-kruse (approved):

PR #1391, opened 2024-02-16
Assorted fixes on -use-cluster tests

hash-d:
- Use routes as ingress, when available
- Ignore unexpected additional clusterroles, when running as `-use-cluster`
- Shutdown previous informers between iterations
- Ignore additional cm `openshift-service-ca.crt`, along with `kube-root-ca.crt`
- Cherry pick Fernando's #1386 into 1.5

### DIFF
--- a/client/router_create_test.go
+++ b/client/router_create_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/skupperproject/skupper/pkg/utils"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/version"
 
@@ -291,6 +292,7 @@ func TestRouterCreateDefaults(t *testing.T) {
 
 		depsFound := []string{}
 		cmsFound := []string{}
+		cmsUpdated := []string{}
 		rolesFound := []string{}
 		clusterRolesFound := []string{}
 		clusterRolesResourcesFound := sets.NewString()
@@ -329,6 +331,10 @@ func TestRouterCreateDefaults(t *testing.T) {
 			AddFunc: func(obj interface{}) {
 				cm := obj.(*corev1.ConfigMap)
 				cmsFound = append(cmsFound, cm.Name)
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				cm := newObj.(*corev1.ConfigMap)
+				cmsUpdated = append(cmsUpdated, cm.Name)
 			},
 		})
 		roleInformer := informerFactory.Rbac().V1().Roles().Informer()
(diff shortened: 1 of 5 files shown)


client/router_create_test.go
```diff
@@ -338,14 +344,23 @@ func TestRouterCreateDefaults(t *testing.T) {
...
+		expectedClusterRoles := sets.NewString(c.clusterRolesExpected...)
 		clusterRoleInformer := clusterRoleInformerFactory.Rbac().V1().ClusterRoles().Informer()
 		clusterRoleInformer.AddEventHandler(&cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
 				clusterRole := obj.(*rbacv1.ClusterRole)
 				if strings.HasPrefix(clusterRole.Name, "skupper") {
-					clusterRolesFound = append(clusterRolesFound, clusterRole.Name)
-					for _, p := range clusterRole.Rules {
-						clusterRolesResourcesFound = clusterRolesResourcesFound.Insert(p.Resources...)
+					if isCluster && !expectedClusterRoles.Has(clusterRole.Name) {
```
fgiorgetti, line 354: I am wondering if this is the appropriate thing to do, or maybe change the comparison done
below with `cmp.Diff` to something that simply ensures all expectedClusterRoles have been found?

hash-d: I just tried it (see updated branch).  The problem is with the `clusterRolesResourcesFound` test; the rules are added into a set, with no reference back to the clusterRole that included them.  Without special treatment, it fails as follows:

```
 === RUN   TestRouterCreateDefaults
     router_create_test.go:461: TestRouterCreateDefaults test two cluster roles mismatch (-want +got):
           []string(Inverse(Sort, []string{
           	"skupper-service-controller-basic",
         + 	"skupper-service-controller-extended",
           }))
     router_create_test.go:462: Checking specifically for expected cluster roles
     router_create_test.go:471: TestRouterCreateDefaults test two cluster roles policy resources mismatch (-want +got):
           sets.String(
         - 	{"namespaces": {}, "nodes": {}, "skupperclusterpolicies": {}},
         + 	{"namespaces": {}, "nodes": {}, "pods": {}, "skupperclusterpolicies": {}},
```

ie, the clusterRole test works fine, ignoring the additional `skupper-service-controller-extended` role, but the `policy resources` (rules) test fails with a rule that comes from the ignored role.

I could refactor `clusterRolesResourcesFound`, so that it becomes a map from clusterRole name to set of rules...  But then I'd have to either specify ruleset per role on the test table, or check that each expected rule is present on at least one role and the other way around (every rule from a role is present on the  `clusterRolesResourcesFound`).

That would make the test a bit better, actually, but also more complex; is it worth it?

The original approach seemed cleaner to me, though.  The end result is the same, in any case: unexpected additional roles get logged and ignored, when running in a cluster.

hash-d: @fgiorgetti, as we talked, I reverted the changes back to the original approach.  It passed CI and it is ready for review again

fgiorgetti (approved):

PR #1398, opened 2024-02-26
use cache when building container images

c-kruse:
Reduces CI `build_and_save_test_images` time from 20+ minutes to ~7.

Instead of forcing docker to re-run everything in the builder step for each image produced (together `go mod download` and `make` seem to take around 3 minutes a pop in CI), allow docker layer caching.

~~Locally podman buildx build doesn't seem quite as clever about caching layers, still trying to sort that all out.~~ Local builds with podman are not totally insane for me now 3-4 minutes (though it's still very easy to work against yourself if you keep editing while the build is running - `COPY . .` means just about anything will break the cache. e.g. `time make docker-build` is ~3m. `time make docker-build | tee docker-build.log` is ~20m.)

### DIFF
--- a/Dockerfile.flow-collector
+++ b/Dockerfile.flow-collector
@@ -2,7 +2,6 @@ FROM --platform=$BUILDPLATFORM golang:1.20 AS builder
 
 ARG TARGETOS
 ARG TARGETARCH
-ARG BUILDARCH
 
 WORKDIR /go/src/app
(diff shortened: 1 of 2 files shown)


fgiorgetti (approved):
Looks good to me. I have tested it locally and it seems to work fine.
Thanks @c-kruse !

PR #1399, opened 2024-02-27
ADR for using CRDs

grs:


doc/adr/0009-use-crds-in-declarative-interface.md
```diff
@@ -0,0 +1,74 @@
...
+minimise the amount of work required.
+
+## Decision
+
+Instead of creating ConfigMaps and Secrets with annotations users will
+create custom resources to configure Skupper.
+
+## Consequences
+
+Standard RBAC can be used to limit, at a coarse granularity, what
```
fgiorgetti, line 34: On podman sites, if CRs are stored as regular files (which is my current inclination so far), those files must have been granted write permission to the current user id, as podman sites will need permissions to update their statuses.
Would it be fair to assume that an eventual filesystem directory can be owned by a certain unix group and then a similar limitation could be applied as well?

doc/adr/0009-use-crds-in-declarative-interface.md
```diff
@@ -0,0 +1,74 @@
...
+The use of CRDs also allows basic rules about what is allowed to be
+enforced in a way that is not easy with annotations and labels. As an
+example we can define the connector resource to require values for
+port and routingKey and to only accept one of host or selector, not
+both. This makes it possible to avoid certain misconfigurations.
+
+The controller can also use the status section of defined CRDs to
+provide information back to the user, e.g. regarding errors or
+indicating that configuration has been applied. This allows a better
+user experience using just the standard kubectl tool.
```
fgiorgetti, line 55: My initial tought would be to have a similar behavior on podman sites, by updating the statuses of CRs.
But in this case users won't simply have the benefit of kubectl, instead they'd still need the CLI for a better
user experience (other than using ls, grep and other shell tools).

doc/adr/0009-use-crds-in-declarative-interface.md
```diff
@@ -0,0 +1,74 @@
...
+
+For example, invoking `kubectl get sites -A` on a cluster will show
+all the namespaces in which skupper has been enabled. We can also make
+use of the 'additional printer columns' feature to show high level
+information about each site. Likewise the list of outgoing links
+configured, the connectors that expose workloads or the listeners that
+define skupper backed services offered can all be listed, either for a
+specific namespace or for the cluster as a whole.
+
+It will not be possible to use Skupper without first installing the
```
fgiorgetti, line 65: What about people trying Skupper using the CLI?
Do you anticipate the CLI also having the capability of installing the CRDs, possibly using the same client library mentioned below?

doc/adr/0009-use-crds-in-declarative-interface.md
grs: A podman site is single tenant, so I'm not sure how important any such restrictions would be? The host may allow separate users, and I guess you may then want some control over what users can access from the host machine, but I think that would be through the regular OS controls perhaps?

doc/adr/0009-use-crds-in-declarative-interface.md
grs: That sounds reasonable.

doc/adr/0009-use-crds-in-declarative-interface.md
grs: I don't think the CLI should install the CRDs by default. The CLI will create CRs though, and watch/report on their statuses.

nluaces (approved):

doc/adr/0009-use-crds-in-declarative-interface.md
fgiorgetti: Oh, I agree that it should not do it by default, and actually this question is possibly not appropriate for this architectural decision phase. But I was just curious about how it could be realised.

Anyway we could, eventually, have a command (i.e.: crd-install) or even a helm chart that could make it easier for users to try skupper just using the CLI.

I know that a prior kubectl apply execution won't be a big deal, but it is usually something that people will need to copy and paste, so if we could also provide a simpler way to achieve it, the user experience can be kept simple through the CLI only, as it is today.

grs: Yes, I agree, simplifying as much as we can is a good goal.

doc/adr/0009-use-crds-in-declarative-interface.md
fgiorgetti: Yes, my assumption is that people will see it as a single tenant as well and rely on OS related permissions to control that, not expecting skupper to do so.

fgiorgetti (approved):

PR #1400, opened 2024-02-27
Faster CI Tests

c-kruse:
Tweaks to circleci to make the `test` job run faster. Results TBD, but should be ~8 minutes quicker.

* Stop running `make package`. That target doesn't appear to be used anywhere else, and we don't do anything with the output. 5 minutes burnt.
* Chooses a more appropriate default for `go test -p`. Many of our tests have sleeps in them. Without sorting that out the second best thing we can do is to allow the go test runner to run the test suite from multiple packages in parallel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -468,6 +468,7 @@ jobs:
           coverpkg: "./api/...,./client/...,./cmd/...,./pkg/..."
           verbose: true
           coverprofile: "/tmp/artifacts/cover.out"
+          parallel: "3"
       - run:
           name: Generating HTML coverage report
           command: go tool cover -html /tmp/artifacts/cover.out -o /tmp/artifacts/cover.html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -458,8 +458,7 @@ jobs:
                 exit 1
               fi
       - go/mod-download-cached
-      - run: make
-      - run: make clean
+      - run: make generate-client
       - run:
           name: Creating artifacts directory
           command: mkdir /tmp/artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -459,7 +459,6 @@ jobs:
               fi
       - go/mod-download-cached
       - run: make
-      - run: make package
       - run: make clean
       - run:
           name: Creating artifacts directory


c-kruse:
Could use a double check on the meaningfulness of `make package`!

Also, I left a full `make` in there, but I'm not really sure if it does much for us. I'm tempted to change it to `make generate-client` to probably shave off another 60s

grs:
> Could use a double check on the meaningfulness of `make package`!

I believe this is a stale target from very early days where it was used for creating the cli executable. Removing it sounds sensible to me.

nluaces (approved):

.circleci/config.yml
```diff
@@ -458,16 +458,15 @@ jobs:
...
-      - run: make clean
+      - run: make generate-client
       - run:
           name: Creating artifacts directory
           command: mkdir /tmp/artifacts
       - go/test:
           coverpkg: "./api/...,./client/...,./cmd/...,./pkg/..."
           verbose: true
           coverprofile: "/tmp/artifacts/cover.out"
+          parallel: "3"
```
hash-d, line 469: I think this will actually decrease parallelism.  The job is running on `Linux Large`, which has 4 CPUs [1].  `go test`, by default, runs in parallel as many tests as there are CPUs on the machine running it, so I'd expect it to be running four tests in parallel, today.

[1] https://circleci.com/docs/configuration-reference/#resourceclass

c-kruse: The go/test orb is doing something behind the scenes I don't feel is very intuitive. If you look at the actual go command that it runs in main right now it is something like this: `go test  -p 1 ./...`, which effectively constrains it to one test at a time.

Three was chosen because the executor this is running on `go_cimg` doesn't specify a resource class - so as I understand it, it is defaulted to medium (2 vCPUs). The 1.5x ratio was kind of arbitrarily chosen based on experience and a hunch that a lot of our test time is comprised of sleeping anyhow.

hash-d: Gotcha.  My fault, I didn't see it was dealing with the `test` job, not with `main_tests_minikube_local_cluster`.

fgiorgetti (approved):

ajssmith (approved):

PR #1408, opened 2024-02-28
adr: Skupper network with v1.X and v2.X sites

nluaces:
If approved, it has to be merged **after** #1399


nluaces:
If the routers create not compatible tcpConnectors/tcpListeners  (lite vs legacy) with the different versions of Skupper, this adr is not going to be valid.

fgiorgetti (approved):

c-kruse (approved):

ajssmith (approved):

PR #1413, opened 2024-03-08
Stop leaking file descriptors in podman controller

c-kruse:
Looks like we are not closing the skupper-network-status file after writing it. Since this gets updated rather frequently I suspect there is some risk of exhausting FDs. This change closes that file.


fgiorgetti (approved):
Thanks for catching it.

grs (approved):

nluaces (approved):

PR #1416, opened 2024-03-15
Include http server TLS and timeout configuration

c-kruse:
Introduces a `tlscfg.Default()` configuration that is backwards compatible with as many older clients as possible while still excluding ciphers marked as insecure in the go standard library. Also includes a TLSv1.3 only configuration `tlscfg.Modern()` that I suggest we use for skupper-to-skupper APIs (e.g. claims service.)

Also introduces server read and write timeouts to help mitigate the risk of stalled clients tying up resources. Values chosen conservatively to avoid impacting any desired traffic.


fgiorgetti (approved):
Excellent!

PR #1425, opened 2024-03-29
Include Operational Metrics

c-kruse:
## What
Adds the following metrics:
```
# HELP active_links Number of active links by site and direction
# TYPE active_links gauge
active_links{ direction="incoming|outgoing", sourceSite="..." }

# HELP active_routers Number of routers
# TYPE active_routers gauge
active_routers

# HELP active_sites Number of sites attached to the network
# TYPE active_sites gauge
active_sites 2
```

## Caveats
I am still wavering on whether or not this is an acceptable solution. Without finding a way to really tackle some of the core issues that make tracking VAN state reliably in the collector, this was my best try to give closer to accurate operational metrics for active sites, routers and links.

active_sites and active_routers both lag (by 60s for vanflow source timeout) but I am otherwise confident in their accuracy.

The active Links metric is fallible in situations where multiple faults have occurred, but should in most cases be more responsive than active sites and routers the way I've implemented it (counting only full incoming/outgoing link pairs once per router pair.)

**Example: A fault occurs at 11:05**
image

In this example, a VAN with two sites is disconnected at 11:05. The collector side observes immediately that the incoming link from the remote side's connection was closed. At 11:06 the collector has not heard from the remote site in 60 seconds, so it expires its router.

### When link metrics lie

An example of where the link metric is fallible would be in a configuration where three sites are linked in a triangle. If you segment the site with the collector, then quickly remove the link between the other two sites, then restore connectivity to the collector site it will still observe 3 total links even though there should only be 2.

**After** doing this maneuver, the console still shows the link that was removed, and the link metrics reflect the same.
Screenshot from 2024-03-29 13-15-56
image

**After restarting the collector** the glitch is fixed:
image
image

### DIFF
--- a/pkg/flow/flow_mem_driver.go
+++ b/pkg/flow/flow_mem_driver.go
@@ -2792,8 +2792,12 @@ func (fc *FlowCollector) needForSiteProcess(flow *FlowRecord, siteId string, sta
 	return found
 }
 
-// graph relations between routers and sites using incoming/outgoing link
-// pairs.
+// graph relations between routers and sites using LinkRecords
+//
+// Collector state is falible, as such this graph is meant to be conservative
+// in the edges it includes between nodes. It deduplicates links (see
+// skupperproject/skupper-router issue #1456) and also for inter-router links
+// ensures both the listener and connector side links are present.
 func (fc *FlowCollector) graph() (routers, sites map[string]*node) {
 	siteByRouterID := make(map[string]string, len(fc.Routers))
 	routerIDByName := make(map[string]string, len(fc.Routers))
@@ -2804,22 +2808,27 @@ func (fc *FlowCollector) graph() (routers, sites map[string]*node) {
 		siteByRouterID[router.Identity] = router.Parent
 		routerIDByName[normalizeRouterName(*router.Name)] = router.Identity
 	}
-	setLink := func(set map[string]map[string]struct{}, local, peer string) {
+	setLink := func(set map[string]map[string]string, local, peer, role string) {
 		s, ok := set[local]
 		if !ok {
-			set[local] = make(map[string]struct{})
+			set[local] = make(map[string]string)
 			s = set[local]
 		}
-		s[peer] = struct{}{}
+		s[peer] = role
 	}
-	// find unique links by localRotuerID->peerRouterID
-	outgoing := make(map[string]map[string]struct{})
-	incoming := make(map[string]map[string]struct{})
+	// find unique links by localRotuerID and peerRouterID
+	outgoing := make(map[string]map[string]string)
+	incoming := make(map[string]map[string]string)
 	for _, link := range fc.Links {
 		if link.Direction == nil || link.Name == nil {
 			continue
 		}
 		localRouterID, peerRouterName := link.Parent, *link.Name
+
+		linkRole := "inter-router"
+		if link.Mode != nil && *link.Mode != "" {
+			linkRole = *link.Mode
+		}
 
 		if _, ok := fc.Routers[localRouterID]; !ok {
 			continue
@@ -2830,9 +2839,9 @@ func (fc *FlowCollector) graph() (routers, sites map[string]*node) {
 		}
 		switch *link.Direction {
 		case "incoming":
-			setLink(incoming, localRouterID, peerRouterID)
+			setLink(incoming, localRouterID, peerRouterID, linkRole)
 		case "outgoing":
-			setLink(outgoing, localRouterID, peerRouterID)
+			setLink(outgoing, localRouterID, peerRouterID, linkRole)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


grs:
I believe this *is* an acceptable solution. The key interface is the three counters. Even with an imperfect implementation, those counters bring real value. I think introducing a simple implementation that will help now and can be enhanced in the future without a fundamental change in the interface is a good tradeoff.

Question regarding the scenario described where the active link count is wrong: Is this where the link is actually removed rather than just the connection is lost and not re-established? If so would the latter case (which is the more important operationally) manifest the same error? Is there any signal of failed inter-router connections attempts? (If so could a counter of that provide a backup approach for detecting problems?)

c-kruse:
Thanks @grs!
> Question regarding the scenario described where the active link count is wrong: Is this where the link is actually removed rather than just the connection is lost and not re-established? If so would the latter case (which is the more important operationally) manifest the same error?

To produce this example I did remove the link, but the later case where the connection is lost and cannot be reestablished would manifest exactly the same.

> Is there any signal of failed inter-router connections attempts?

Good question! We do get log records that I take to mean that. Depending on exactly how I break the connection the log records look like this:
```
{
        "ID": "h8ht7:88",
        "StartTime": "2024-04-04T15:03:21.14258Z",
        "EndTime": "2024-04-04T15:03:21.14258Z",
        "LogSeverity": 3,
        "LogText": "LOG_SERVER: [C184] Connection from 10.42.0.1:12603 (to :55671) failed: amqp:resource-limit-exceeded local-idle-timeout expired",
        "SourceFile": "/build/src/server.c",
        "SourceLine": 1107
}

{
        "ID": "6xdf4:56",
        "StartTime": "2024-04-04T15:03:20.915248Z",
        "EndTime": "2024-04-04T15:03:20.915248Z",
        "LogSeverity": 3,
        "LogText": "LOG_SERVER: [C77] Connection to 192.168.11.22:55671 failed: amqp:connection:framing-error SSL Failure: error:0A000126:SSL routines::unexpected eof while reading",
        "SourceFile": "/build/src/server.c",
        "SourceLine": 1102
}
```
I can't really say I understand exactly what is triggering them, how they behave over long periods (log event limits, connection retry limit/backoff, etc), or if there is anything in LogText that we can consider part of a stable API to match on. Maybe we want a counter per severity level?
>   VFLOW_LOG_SEVERITY_INFO     = 1,  // normal operational events
    VFLOW_LOG_SEVERITY_WARNING  = 2,  // recoverable or transient failure with minor or no loss of service
    VFLOW_LOG_SEVERITY_ERROR    = 3,  // service affecting failure that requires intervention
    VFLOW_LOG_SEVERITY_CRITICAL = 4   // catastrophic loss of routing service

pkg/flow/flow_mem_driver.go
```diff
@@ -2771,3 +2791,113 @@ func (fc *FlowCollector) needForSiteProcess(flow *FlowRecord, siteId string, sta
...
+	}
+	siteNodes := make(map[string]*node, len(fc.Sites))
+	for id := range fc.Sites {
+		siteNodes[id] = &node{ID: id}
+	}
+
+	for router, links := range outgoing {
+		for peerRouter, role := range links {
+			switch role {
+			// Since the router does not presently include Link records for
```
nluaces, line 2860: thank you for writing this comment!

nluaces:
Thank you for this pull request @c-kruse!
I wonder if you also need to filter out routers (at least for these metrics) that are created when a service is exposed as headless, like the CLI does.

c-kruse:
Excellent call out @nluaces! I hadn't considered that. As-is, they definitely are get counted along with the "regular" edge routers. I definitely understand excluding them in the CLI, but here I could probably be swayed one way or the other in this view.

I suppose we'd need to use some similar name matching logic to determine this here - i.e. has prefix `0/{site.Name}` or something to that effect?

nluaces:
@c-kruse 
This is what I've found out so far:

1. headless routers are created with a name that contains the prefix `0/{address}-proxy`
2. regular routers (kubernetes sites) are created with a name that contains the prefix `0/{site.Name}` as you mentioned
3. regular routers (podman sites) are created with a name that matches exactly with `0/{site.Name}`
4. gateway routers are created with a name that contains the prefix `0/skupper-gateway`

The network status command displays all these routers except the headless service router(1).

c-kruse:
Thanks @nluaces! Boy. I don't envy having to sort that all out on the CLI. Maybe we should talk about including a router "deployment role" or something for v2 to solve that problem a little more directly?

I believe I've found a nice way to totally skirt this issue as it applies to these metrics. I've just excluded links between routers in the same site from the metrics. This should get us _pretty close_ to parity with what `skupper network status` displays. The only exception I've found so far is that this is still counting both links from an HA site (--routers 2) and the cli seems to de-duplicate those.

fgiorgetti:
The metrics seem to work quite well.

Today Skupper configures the router entity (in skupper-internal) with a metadata field. Maybe we should evaluate including something there to provide the actual function of the given router, although that would not propagate automatically, but it could eventually help distinguishing their purposes in metrics/console.

Anyway I like seeing the real picture that shows up the number of active links and routers.
As long as the number of sites is accurate, the number of routers/links won't be so easy for users to predict (consider gateways, podman services, headless svcs, sites initialized with > 1 routers for example), but I don't see it as a problem, maybe just something to be explained better in the docs.

One thing that might help (?) would be adding the role to the active_links metric, so users could eventually filter by type, if desired (edge/inter-router).

grs:
> As long as the number of sites is accurate, the number of routers/links won't be so easy for users to predict (consider gateways, podman services, headless svcs, sites initialized with > 1 routers for example), but I don't see it as a problem, maybe just something to be explained better in the docs.

I agree that this is not a big problem. The most important aspect is not the actual count, but any changes in the count.

grs (approved):

nluaces (approved):

bartoval (approved):

ajssmith (approved):
lgtm, thnx

PR #1430, opened 2024-04-05
Site controller is not handling connection token request properly

fgiorgetti:
When site controller is watching all namespaces, it is not processing connection-token-requests properly, as the correct namespaces is not passed along correctly.


nluaces (approved):

c-kruse (approved):

ajssmith (approved):
lgtm

PR #1432, opened 2024-04-08
exclude book info test from the CI build

nluaces:
```
curl -ik docker.io/maistra/examples-bookinfo-productpage-v1:0.12.0                                                                                                                 
HTTP/1.1 301 Moved Permanently
content-length: 0
location: https://docker.io/maistra/examples-bookinfo-productpage-v1:0.12.0
```
The images needed for this example are not available in the docker hub.

### DIFF
--- a/test/integration/examples/bookinfo_test.go
+++ b/test/integration/examples/bookinfo_test.go
@@ -1,5 +1,5 @@
-//go:build integration || examples
-// +build integration examples
+//go:build examples
+// +build examples
 
 package examples
 


fgiorgetti:
@nluaces @ajssmith It seems like they've moved the images.
Regardless we make the bookinfo test run as part of the examples only, we need
the fix below so that the example we have listed in skupper.io still works:
https://github.com/skupperproject/skupper-example-bookinfo/pull/4

@nluaces if the example above fixes it, maybe we should keep it in our CI.

nluaces:
@fgiorgetti I completely agree with your suggestion, I just wanted to exclude it temporarily because main and the rest of pull requests branches are failing because of this issue (perhaps I should have made this clarification).

I was thinking to make the main and the other branches green, fix the example and include the test again to the CI.

fgiorgetti:
> @fgiorgetti I completely agree with your suggestion, I just wanted to exclude it temporarily because main and the rest of pull requests branches are failing because of this issue (perhaps I should have made this clarification).
> 
> I was thinking to make the main and the other branches green, fix the example and include the test again to the CI.

Yes, and I agree with you as it makes sense. I have merged that example PR.
Lets see if main branch passes now, otherwise we merge this one. WDYT?

Issue #1427, opened 2024-04-04
null pointer dereference in policy controller


grs:
**Describe the bug**

On platforms without DeploymentConfig, if cluster policy CRD is installed (and therefore policy enabled), and a service is annotated, the service controller will crash with:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x131007f]

goroutine 304 [running]:
github.com/skupperproject/skupper/pkg/kube.GetDeploymentConfig({0xc000890b10, 0x9}, {0xc00004802a, 0x3}, {0x0?, 0x0?})
	/go/src/app/pkg/kube/deploymentconfigs.go:12 +0x3f
github.com/skupperproject/skupper/pkg/kube.GetServiceInterfaceTarget({0x17fc0e3?, 0x1?}, {0xc000890b10, 0x9}, 0x1, {0xc00004802a, 0x3}, {0x1a6d498?, 0xc00020cc60?}, {0x0, ...})
	/go/src/app/pkg/kube/misc.go:90 +0x69e
main.(*PolicyController).inferTargetType.func1({0xc000bd1b08?, 0x3, 0x0?})
	/go/src/app/cmd/service-controller/policy_controller.go:412 +0xea
main.(*PolicyController).inferTargetType(0x0?, {{0xc000890b10, 0x9}, {0xc00088b368, 0x17}, 0xc0009d8180, {0x0, 0x0}, {0x0, 0x0}}, ...)
	/go/src/app/cmd/service-controller/policy_controller.go:422 +0x150
main.(*PolicyController).validateExposeStateChanged(0xc000661c80)
	/go/src/app/cmd/service-controller/policy_controller.go:349 +0x338
main.(*PolicyController).validateStateChanged(0xc000661c80?)
	/go/src/app/cmd/service-controller/policy_controller.go:461 +0x3c
main.(*PolicyController).start.func1()
	/go/src/app/cmd/service-controller/policy_controller.go:118 +0x1e6
created by main.(*PolicyController).start
	/go/src/app/cmd/service-controller/policy_controller.go:89 +0x72
```

**How To Reproduce**

Initialise a namespace with skupper. Create a deployment with a service. Annotate the service with skupper.io/proxy=tcp. Install skupper ClusterPolicy CRD.

**Expected behavior**

No crash.

**Environment details**
 - Skupper CLI: 1.6.0
[...]

PR #1433, opened 2024-04-10
Prevent segfault when apps/v1 client not available

c-kruse:
Adds a nil check around the usage of the apps/v1 client to prevent a service-controller crash when it attempts to look for a DeploymentConfig matching a service target in a kubernetes environment without the DeploymentConfig resource installed.

Closes https://github.com/skupperproject/skupper/issues/1427

### DIFF
--- a/pkg/kube/misc.go
+++ b/pkg/kube/misc.go
@@ -87,6 +87,9 @@ func GetServiceInterfaceTarget(targetType string, targetName string, deducePort 
 		}
 		return &target, nil
 	} else if targetType == "deploymentconfig" {
+		if appscli == nil {
+			return nil, fmt.Errorf("can not read deploymentconfig without apps/v1 client")
+		}
 		depconfig, err := GetDeploymentConfig(targetName, namespace, appscli)
 		if err == nil {
 			target := types.ServiceInterfaceTarget{


grs:
This will prevent the segfault, which is certainly the most important aspect. However the policy controller will still not correctly infer the target type for an annotated service. We could certainly address that aspect in a separate fix however. Thoughts @fgiorgetti ?

fgiorgetti:
> This will prevent the segfault, which is certainly the most important aspect. However the policy controller will still not correctly infer the target type for an annotated service. We could certainly address that aspect in a separate fix however. Thoughts @fgiorgetti ?

That is correct. The infer logic is not assertive as it tries to guess the type. I was discussing about it with @c-kruse last
week and we are evaluating a precise approach that does it differently.

Basically the idea is to look at the target's selector, and determine its real owner by tracking its owner references.
That'd work for all allowed types, but we'd need a special condition for annotated services, as on annotated services,
the target information has just a "name" and the selector. In this case we'd need to look at the service's annotations to
differentiate whether the service was annotated or not.

I will raise a separate upstream tracker to handle that enhancement.

@grs @c-kruse what do you guys think?

grs:
Makes sense to me to track that issue separately. I would suggest we go for something simple to begin with if possible. E.g. if the current attempt at inference fails (i.e. inferred target type is ""), and the service has origin "annotation", then we either assume it is a service or perhaps check that an annotated service of the expected name exists before doing so.

grs (approved):

c-kruse:
Agreed. Thank you both!

PR #1442, opened 2024-04-16
Target type inference improvement on policy controller

### LINKED ISSUE #1439: Improve target type inference mechanism used by the policy controller
**Describe the bug**
If an existing service is exposed, using annotations only, the actual target are the
original selectors for that service and that is causing the mechanism to fail as it is
unable to determine the appropriate target type.

Along with that, the policy controller uses a mechanism to determine the target type
of a skupper service, by "guessing" what the type is, based on its exposed selectors,
or in case a target service (hostname) is used, then it assumes it is a service.

**How To Reproduce**

Enable policies, then through a Skupper Policy, allow only the exposure of a "deployment"
kind for your "my-service" resource name. Then expose an existing service, named my-service,
using annotations.

The policy controller will allow it to be exposed, as it is unable to infer the target type.

**Expected behavior**
Skupper should have an assertive way of determining the target type.

**Environment details**
- Skupper CLI: 1.6.0
- Platform: kubernetes, openshift

fgiorgetti:
Fixes #1439.

### DIFF
--- a/cmd/service-controller/policy_controller.go
+++ b/cmd/service-controller/policy_controller.go
@@ -407,6 +407,10 @@ func (c *PolicyController) inferTargetType(target types.ServiceInterfaceTarget, 
 	if target.Selector == "" {
 		return ""
 	}
+	service, _ := kube.GetService(target.Name, namespace, c.cli.KubeClient)
+	if service != nil && hasProxyAnnotation(*service) {
+		return "service"
+	}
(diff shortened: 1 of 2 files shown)


grs (approved):
LGTM

nluaces (approved):

c-kruse (approved):

PR #1445, opened 2024-04-17
Include Skupper v2 CLI specification

nluaces:
Original source: https://github.com/ssorj/skupper-cli-spec/tree/main

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ orbs:
 executors:
   go_cimg:
     docker:
-      - image: cimg/go:1.20
+      - image: cimg/go:1.21
     environment:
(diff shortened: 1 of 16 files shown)


fgiorgetti:
Should this PR be opened against the v2 branch instead of main ?

fgiorgetti:
> We are merging ADRs that are V2 related in main, should we treat this one differently?

I think that for ADRs, it should be ok. But my impression is that it might be confusing to
people looking at `site create` in the main branch.

ajssmith:
I would also lean to placing in v2 branch for now...thnx

nluaces:
I need to close this pr to no introduce any other commit from main.

PR #1447, opened 2024-04-18
Build images with docker format

### LINKED ISSUE #1434: Skupper 1.6 images will not run on OCP 3.11
**Describe the bug**

The pods for Skupper 1.6 fail to start on OCP 3.11, due to an image fetching error:

```
19m         19m          2       skupper-service-controller-7f646c5cc7-rfw74.17c5442b837d49d8   Pod           spec.containers{service-controller}   Warning   Failed              kubelet, node-2.x   Failed to pull image "quay.io/skupper/service-controller:main": rpc error: code = Unknown desc = missing signature key
```

That error on OCP 3.11 is typical of images that use the OCI manifest format, which is not supported by the older technology present on OCP 3.11.

Inspection of the images confirm it:

```
$ podman image inspect quay.io/skupper/config-sync | grep -i manifest
          "ManifestType": "application/vnd.oci.image.manifest.v1+json",
```

**How To Reproduce**

Just `skupper init` against an OCP 3.11 cluster.

Skupper pods are never started.

**Expected behavior**

Skupper site to be initialized.

**Environment details**
 - Skupper CLI: 1.6.0
 - Skupper Operator (if applicable): N/A
 - Platform: Openshift 3.11

**Additional context**

N/A

fgiorgetti:
* Use docker 26 in CI for building and publishing images
* Generated images should be compatible with older versions of docker

Fixes #1434.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-    docker: circleci/docker@1.0.1
+    docker: circleci/docker@2.6.0
     kube-orb: circleci/kubernetes@0.11.0
     go: circleci/go@1.7.1
 
@@ -15,6 +15,8 @@ executors:
     machine:
       image: ubuntu-2204:2022.10.2
     resource_class: large
+    environment:
+      CGO_ENABLED: 0
 
   local_cluster_policy_test_executor:
     machine:
@@ -479,7 +481,7 @@ jobs:
     steps:
       - docker/install-docker
       - checkout
-      - run: docker buildx create --use --name skupper-buildx
+      - run: docker buildx create --use --name skupper-buildx --bootstrap
       - run: make -e docker-build
       - run:
           name: persisting images to workspace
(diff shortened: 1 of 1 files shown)


grs (approved):

nluaces (approved):

hash-d (approved):
@fgiorgetti, I must confess I do not understand the need for the additional changes (executor, CGO_ENABLED and --bootstrap), but I do not see a problem with them either.  If you could educate me on that, I'd appreciate.  In any case, this lgtm.

fgiorgetti:
> @fgiorgetti, I must confess I do not understand the need for the additional changes (executor, CGO_ENABLED and --bootstrap), but I do not see a problem with them either. If you could educate me on that, I'd appreciate. In any case, this lgtm.

@hash-d Originally the process that build and push images was done using the docker executor.
Unfortunately I could not use a newer version of docker, that could build multi-platform images
with the docker format (all allowed versions of the docker executor only produces OCI).
This is the reason why I have modified this process to use the `machine` executor.

The CGO_ENABLED environment variable was already set on the docker executor, so I am just adding it 
to the machine executor so that we continue to produce statically linked binaries.

Last, the `--bootstrap` option seems to be needed now (with this newer version, or at least, in my local
environment it was), to create and start the buildx instance.

Issue #1312, opened 2023-12-06
skupper init --platform podman is not using podman context


bryonbaker:
When using the same console for configuring a podman site and kubernetes site, the skupper init is not recognising the --platform podman option and instead attempts to init in the kubernetes context.

 

$ skupper switch kubernetes

$ skupper init --platform kubernetes --site-name K8s --enable-console --enable-flow-collector --console-auth=internal --console-user=admin --console-password=password

Skupper is now installed in namespace 'comm'.  Use 'skupper status' to get more information.

$ skupper init --site-name MASTER --platform podman --ingress-host 10.10.10.48

Secret already exists:  skupper-local-server
Secret already exists:  skupper-local-client
Secret already exists:  skupper-service-client
Secret already exists:  skupper-console-users
Secret already exists:  skupper-site-server
Skupper is now installed in namespace 'comm'.  Use 'skupper status' to get more information.

However, this works:

$ skupper delete --platform kubernetes
Skupper is now removed from 'comm'.
$ skupper init --platform kubernetes --site-name K8s --enable-console --enable-flow-collector --console-auth=internal --console-user=admin --console-password=password
Skupper is now installed in namespace 'comm'.  Use 'skupper status' to get more information.

$ skupper switch podman
$ skupper init --site-name MASTER --platform podman --ingress-host 10.10.10.48
It is recommended to enable lingering for bryon, otherwise Skupper may not start on boot.
Skupper is now installed for user 'bryon'.  Use 'skupper status' to get more information.

Version information:
$ skupper version
client version                 1.5.0
transport version              quay.io/skupper/skupper-router:2.5.0 (sha256:55f014d0fcf4)
controller version             quay.io/skupper/service-controller:1.5.0 (sha256:11a88e11ee66)
config-sync version            quay.io/skupper/config-sync:1.5.0 (sha256:6a8f5846a805)
flow-collector version         quay.io/skupper/flow-collector:1.5.0 (sha256:f07ea035141f)

gabordozsa:
This might be closed now as per https://github.com/skupperproject/skupper/pull/1451#issuecomment-2072966500 (cc. @fgiorgetti )

PR #1451, opened 2024-04-23
Fix the early parsing of platform flag in init()

### LINKED ISSUE #1147: Platform flag not working as expected when exposing a resource in Skupper
## Expected Behavior
I was using podman to run a posgresql database, and did not set up the `SKUPPER_PLATFORM` environment variable.
After creating a skupper site with `--platform podman`, I used the same flag to expose the database into the VAN:
```
skupper expose host  database --address localhost --port 5432 --platform podman
```

## Current Behavior
It did not expose the database because it didn't recognize the target type host, but the command accepts the flag --platform.
After that, I exported SKUPPER_PLATFORM with the podman value, exposed the database with 

`skupper expose host  database --address localhost --port 5432`

and it worked fine.

## Failure Information 
```
skupper expose host  database --address localhost --port 5432 --platform podman                 
Error: target type must be one of: [deployment, statefulset, pods, service, deploymentconfig]
[...]
```

### Context
* skupper version: 1.4.0-rc4

gabordozsa:
The parsing of `--platform` happens very early in the `init()` in order to decide which type of client to instantiate. However, no other flags are defined yet at that point. This causes the parsing to fail if any other flag is encountered in the command line before the `--platform` one (and the code then falls back to the default Kubernetes platform). 

Fixes #1147

### DIFF
--- a/cmd/skupper/skupper.go
+++ b/cmd/skupper/skupper.go
@@ -1039,12 +1039,27 @@ func addCommands(skupperCli SkupperClient, rootCmd *cobra.Command, cmds ...*cobr
 	}
 }
 
+func parsePlatformFlagOnly(args []string) {
+	// This function is called when not every supported command flag is defined yet
+	for i, p := range args {
+		if p == "--platform" && i < len(args)-1 {
+			config.Platform = args[i+1]
+			break
+		} else if strings.HasPrefix(p, "--platform=") {
+			_, v, _ := strings.Cut(p, "=")
+			config.Platform = v
+			break
+		}
+	}
+}
+
 func init() {
 	rootCmd = &cobra.Command{Use: "skupper"}
 	routev1.AddToScheme(scheme.Scheme)
 
 	rootCmd.PersistentFlags().StringVarP(&config.Platform, "platform", "", "", "The platform type to use [kubernetes, podman]")
-	rootCmd.ParseFlags(os.Args)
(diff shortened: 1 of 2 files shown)


cmd/skupper/skupper.go
```diff
@@ -1039,12 +1039,34 @@ func addCommands(skupperCli SkupperClient, rootCmd *cobra.Command, cmds ...*cobr
...
+	initWrapper(os.Args)
+}
+
+func initWrapper(args []string) {
 	rootCmd = &cobra.Command{Use: "skupper"}
 	routev1.AddToScheme(scheme.Scheme)
 
-	rootCmd.PersistentFlags().StringVarP(&config.Platform, "platform", "", "", "The platform type to use [kubernetes, podman]")
-	rootCmd.ParseFlags(os.Args)
+	rootCmd.PersistentFlags().StringVarP(&config.Platform, "platform", "", "", "The platform type t [kubernetes, podman]")
```
lynnemorrison, line 1067: looks like "to use" was accidently cut

gabordozsa: Ops, good catch! Fixed.

fgiorgetti:
I believe it also fixes #1312.

grs (approved):

fgiorgetti (approved):

nluaces (approved):

c-kruse (approved):

ajssmith (approved):

PR #1458, opened 2024-04-26
Add expiry time to token.  

### LINKED ISSUE #1415: Add expiry date time to token yaml
I sometimes create default tokens and then get distracted before using them.
When I do use them, they are invalid

If the token yaml had more information, eg time created and token duration that would help me avoid wasting time 
think about:

- did the link fail because I've run out of time or because it was already used?
- is connectivity possible between these two sites

An expiry date time
No of uses 

**Additional context**
While this doesn't solve the problem of creating a link to the wrong site, because you've mixed up tokens it might be a help when the urls are long and all look the same at a glance.

lynnemorrison:
Add expiry time to token.  When token is being used to create a link verify the token hasn't expired.  If it has log a warning but continue with trying to bring up link.

fixes #1415

### DIFF
--- a/client/connector_create.go
+++ b/client/connector_create.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"strconv"
+	"time"
 
 	"github.com/skupperproject/skupper/pkg/domain"
 	domainkube "github.com/skupperproject/skupper/pkg/domain/kube"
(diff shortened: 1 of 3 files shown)


client/connector_create.go
```diff
@@ -113,6 +114,17 @@ func (cli *VanClient) ConnectorCreateSecretFromData(ctx context.Context, options
 				return nil, fmt.Errorf("Claims not supported. %s", err)
 			}
 		}
+		if expirationStr, ok := secret.ObjectMeta.Annotations[types.ClaimExpiration]; ok {
+			expiration, timeErr := time.Parse(time.RFC3339, expirationStr)
+			if timeErr != nil {
+				err = timeErr
+				return nil, fmt.Errorf("timer error %s", err)
```
grs, line 121: I'd suggest we just print a warning here also, e.g. 'cannot parse expiration'

lynnemorrison: fixed

grs: Its still returning an error rather than printing a warning.

lynnemorrison: sorry,  just focused on what was being printed, vs type of error.   I have now fixed to just log a warning but to continue on processing.

grs (approved):

nluaces (approved):

grs:
Thank you @lynnemorrison !

PR #1460, opened 2024-04-29
Sk1765 fix flow protocol race

c-kruse:
Presently when the collector processes a new flow record before that
flow's listener or connector record has been processed it is likely that
Protocol will remain unset indefinitely. This adds logic to set protocol
on flow record reconcile.

Signed-off-by: Christian Kruse <christian@c-kruse.com>

### DIFF
--- a/pkg/flow/flow_mem_driver.go
+++ b/pkg/flow/flow_mem_driver.go
@@ -2369,6 +2369,7 @@ func (fc *FlowCollector) reconcileFlowRecords() error {
 			if flow.SourceHost != nil {
 				if connector, ok := fc.Connectors[flow.Parent]; ok {
 					flow.Place = serverSide
+					flow.Protocol = connector.Protocol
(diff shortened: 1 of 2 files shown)


ajssmith (approved):

c-kruse:
The logic here is a little opaque, so here's the high level:

The collector handles vanflow  messages containing records serially, but not necessarily in the order they were sent - especially in the case of FLOW records which have their own amqp address. Instead of normalizing the Protocol field, and searching through the hierarchy each time we deal with a flow record we have denormalized, copying protocol onto the FLOW record when it is added.

When a new flow is discovered it gets added to the reconcile queue, and every few seconds the collector tries to match up flows in the reconcile queue with their corresponding listener/connector and process. Setting protocol from here should pretty much guarantee protocol gets set before it is associated with a flow pair. The flow "Place" field behaves similarly and seems to be working.

grs (approved):

PR #1464, opened 2024-05-02
skupper init for rootful podman should try /run/podman/podman.sock

### LINKED ISSUE #1401: skupper init for rootful podman should try /run/podman/podman.sock
**Describe the bug**
Attempting to init a podman site as root fails because it looks for podman.sock in the regular user scoped location `unix:///run/user/0/podman/podman.sock`. It appears that the default location for root is `/run/podman/podman.sock`.
**Environment details**
 - Skupper CLI: 1.5x

Karen-Schoener:
Rouine GetRuntimeDir returns path based on whether os.Getuid() is root.

Fixes #1401

### DIFF
--- a/pkg/config/local.go
+++ b/pkg/config/local.go
@@ -161,6 +161,9 @@ func GetConfigHome() string {
 }
 
 func GetRuntimeDir() string {
+	if os.Getuid() == 0 {
+		return "/run"
+	}
 	runtimeDir, ok := os.LookupEnv("XDG_RUNTIME_DIR")
 	if !ok {
 		runtimeDir = fmt.Sprintf("/run/user/%d", os.Getuid())


Karen-Schoener:
Test case: run as root: ./skupper --platform podman init
```
# ./skupper --platform podman init
Unable to create startup service - SystemD is not enabled at user level
The startup scripts: start-podman.sh and stop-podman.sh are available at /root/.local/share/skupper
Waiting for status...
Skupper status is not loaded yet.
Skupper is now installed for user 'root'.  Use 'skupper status' to get more information.
```

Test case: run as non-root: ./skupper --platform podman init
```
$ skupper --platform podman init
It is recommended to enable lingering for kschoener, otherwise Skupper may not start on boot.
Waiting for status...
Skupper is now installed for user 'kschoener'.  Use 'skupper status' to get more information.

[kschoener@fedora skupper]$ podman ps
CONTAINER ID  IMAGE                                    COMMAND               CREATED         STATUS         PORTS                                               NAMES
4bd9c0e3ae28  quay.io/skupper/skupper-router:2.5.2     /home/skrouterd/b...  16 seconds ago  Up 16 seconds  0.0.0.0:45671->45671/tcp, 0.0.0.0:55671->55671/tcp  skupper-router
3dfbb0b614b3  quay.io/skupper/controller-podman:1.6.0  /app/controller-p...  15 seconds ago  Up 15 seconds                                                      skupper-controller-podman
```

c-kruse (approved):
This looks right to me! I admittedly don't totally understand what the service unit RequiresMountsFor option is for, but it does seem to work as expected for both root and non-root podman!

fgiorgetti (approved):
LGTM!

fgiorgetti:
> This looks right to me! I admittedly don't totally understand what the service unit RequiresMountsFor option is for, but it does seem to work as expected for both root and non-root podman!

Just for reference. The template we use for the systemd service is based on the default one that is generated by podman
when you run: `podman generate systemd`.

PR #1472, opened 2024-05-10
Proposing bindHosts and subjectAlternativeNames to LinkAccess CRD

fgiorgetti:


api/types/crds/skupper_link_access_crd.yaml
```diff
@@ -28,8 +28,14 @@ spec:
                   type: string
                 tlsCredentials:
                   type: string
-                accessType:
-                  type: string
+                bindHosts:
+                  type: array
```
grs, line 32: Should this be an array or just a single value? Where we want to bind on multiple interfaces, would that not be via separate LinkAccess instances? One part of the role of this resource is to control the listeners in router config. I believe those can only specify one address to bind on? If so, then I think it is nicer to have a separate LinkACcess CR for each distinct address to bind to. (That also allows greater clarity on what names resolve to that interface).

pkg/apis/skupper/v1alpha1/types.go
```diff
@@ -42,7 +42,7 @@ type SkupperClusterPolicySpec struct {
 type Site struct {
 	v1.TypeMeta   `json:",inline"`
 	v1.ObjectMeta `json:"metadata,omitempty"`
-	Spec          SiteSpec `json:"spec,omitempty"`
+	Spec          SiteSpec   `json:"spec,omitempty"`
```
grs, line 45: These changes seem unrelated to the rest of the PR?

go.mod
```diff
@@ -58,6 +58,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.0 // indirect
 	github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96 // indirect
+	github.com/emicklei/go-restful v2.9.5+incompatible // indirect
```
grs, line 61: Is this a deliberate change?

fgiorgetti: After changing the CRD, I have executed: `FORCE=true scripts/update-codegen.sh`, and I
believe it could have caused those extra mod changes.

pkg/apis/skupper/v1alpha1/types.go
fgiorgetti: These are probably a result of `go fmt` that I ran locally.

api/types/crds/skupper_link_access_crd.yaml
fgiorgetti: I was actually in doubt about it as well. Thinking of eventual usages on kube sites, I believe it makes more
sense to have a single host. I will update it.

grs (approved):

PR #1489, opened 2024-05-23
Deploy namespaced resources in the skupper namespace

skitt:
The ClusterRoleBinding applies to skupper/skupper-controller, so the Deployment and ServiceAccount should be there too. Since the Certificate and SecureAccess resources are namespaced, they probably should be in the skupper namespace as well.


skitt:
Perhaps the namespaced resources should be moved to another file altogether :wink:

PR #1491, opened 2024-05-23
Controller example: deploy CRDs

skitt:
These are required for everything else to work.

### DIFF
--- a/cmd/controller/example/README.md
+++ b/cmd/controller/example/README.md
@@ -1,6 +1,15 @@
 # Installing the skupper controller
 
 ```
+kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/v2/api/types/crds/skupper_certificate_crd.yaml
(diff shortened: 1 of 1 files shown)


cmd/controller/example/README.md
```diff
@@ -1,6 +1,16 @@
 # Installing the skupper controller
 
 ```
+kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/v2/api/types/crds/skupper_certificate_crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/v2/api/types/crds/skupper_claim_crd.yaml
+kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/v2/api/types/crds/skupper_cluster_policy_crd.yaml
```
grs, line 6: Actually we don't want the policy CRD. It is from v1. The current CRDs need to be separated at some point, with the v2 ones pulled into their own grouping. For now, we can just not install the policy crd in this guide.

skitt: I’ve removed that CRD from the list.

grs:
Thanks @skitt !

PR #1493, opened 2024-05-28
do not overwrite unrecognised fields in router config entities

grs:


grs:
Fixes skupperproject/skupper-router#1132

fgiorgetti (approved):
It seems to work fine.

gabordozsa (approved):
Looks good! 

Just one comment: this approach requires changes in the `skupper-router` code as well. Otherwise, `skupper-router` cannot find the parameters defined  in `extraFields` (so it will  still fallback to the default values). This PR is a partial fix for https://github.com/skupperproject/skupper-router/issues/1132

grs:
> Just one comment: this approach requires changes in the `skupper-router` code as well. Otherwise, `skupper-router` cannot find the parameters defined in `extraFields` (so it will still fallback to the default values). 

The extraFields is purely an in memory holder for fields not explicitly named. The resulting config file which the router reads has the fields with their expected names in the entities as expected.

gabordozsa:
> > Just one comment: this approach requires changes in the `skupper-router` code as well. Otherwise, `skupper-router` cannot find the parameters defined in `extraFields` (so it will still fallback to the default values).
> 
> The extraFields is purely an in memory holder for fields not explicitly named. The resulting config file which the router reads has the fields with their expected names in the entities as expected.

Okay, I got confused. I'm testing this in my local setup and the router still picks up the default value for `idleTimeoutSeconds`. It might be some problem in my test setup.

gabordozsa:
> > > Just one comment: this approach requires changes in the `skupper-router` code as well. Otherwise, `skupper-router` cannot find the parameters defined in `extraFields` (so it will still fallback to the default values).
> > 
> > 
> > The extraFields is purely an in memory holder for fields not explicitly named. The resulting config file which the router reads has the fields with their expected names in the entities as expected.
> 
> Okay, I got confused. I'm testing this in my local setup and the router still picks up the default value for `idleTimeoutSeconds`. It might be some problem in my test setup.

It works for me now. I had a problem in my local setup. Sorry for the noise!

PR #1503, opened 2024-06-13
Improve handling of ssl profile

grs:
This should make the pkg/site/routeraccess .go (moved from pkg/kube/site/routeraccess.go) and pkg/site/link .go usable for other platforms where different paths may be required.


pkg/site/routeraccess.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil
+	}
+	var connectors []qdr.Connector
+	if role, ra := m.findInterRouterRole(); role != nil {
+		for _, group := range targetGroups {
+			name := group
+			connector := qdr.Connector{
+				Name:       name,
+				Host:       group,
+				Role:       qdr.RoleInterRouter,
```
fgiorgetti, line 42: Is this for the HA?

fgiorgetti (commented):
Overall it looks good to me.

pkg/kube/site/site.go
```diff
@@ -31,13 +30,11 @@ type Site struct {
...
 	name        string
 	namespace   string
-	siteId      string
-	config      *types.SiteConfigSpec
 	controller  *kube.Controller
 	bindings    *site.Bindings
 	links       map[string]*site.Link
 	errors      map[string]string
-	linkAccess  LinkAccessMap
+	linkAccess  site.RouterAccessMap
```
fgiorgetti, line 37: Maybe rename it to routerAccess ?

api/types/crds/skupper_link_crd.yaml
```diff
@@ -15,55 +15,39 @@ spec:
...
+                  items:
+                    type: object
+                    properties:
+                      name:
+                        type: string
+                      host:
+                        type: string
+                      port:
+                        type: string
+                      group:
```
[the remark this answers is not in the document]
grs: It is used for HA to group together the ports for a given router

pkg/site/routeraccess.go
grs: yes

pkg/kube/site/site.go
grs: The reason it is called linkAccess is that is only controlling the access for linking.  A separate field will control management access.

PR #1507, opened 2024-06-17
Compat API client implementation

fgiorgetti:
This will be used, as a replacement to libpod client implementation, in order to provide a common
way to deal with container engine sites in V2.


nluaces (approved):

Karen-Schoener (approved):

grs (approved):
Minor naming questions:

- is client the right place for this (as opposed to pkg)?
- is there a name other than 'compat' that would more clearly indicate what the package is for?

fgiorgetti:
> * is client the right place for this (as opposed to pkg)?

That is a good point. I believe we can move the `client/container` into `pkg/container`.

> * is there a name other than 'compat' that would more clearly indicate what the package is for?

With regards to the `client/compat` and the `client/generated` packages, I am not intending to modify
them at this point, because we have a target of moving them (as well as the `generate-client` Makefile's target)
into a separate repository and keep the skupper repo clean.

But at that point, I will try to structure things appropriately in the new repository.

Does that work or do you have any other thoughts?

grs:
> Does that work or do you have any other thoughts?

That works, certainly.

PR #1509, opened 2024-06-17
Add listener create cli command for V2

lynnemorrison:
Add new listener create CLI command and unit tests for V2

### DIFF
--- a/internal/cmd/skupper/listener/listener_create.go
+++ b/internal/cmd/skupper/listener/listener_create.go
@@ -1,47 +1,197 @@
 package listener
 
 import (
+	"context"
 	"fmt"
+	"os"
+	"strconv"
+
 	"github.com/skupperproject/skupper/client"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/utils"
+	"github.com/skupperproject/skupper/pkg/apis/skupper/v1alpha1"
+	skupperv1alpha1 "github.com/skupperproject/skupper/pkg/generated/client/clientset/versioned/typed/skupper/v1alpha1"
+	"github.com/skupperproject/skupper/pkg/utils/validator"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 var (
-	listenerCreateLong    = ""
-	listenerCreateExample = ""
+	listenerCreateLong    = "Clients at this site use the listener host and port to establish connections to the remote service."
+	listenerCreateExample = "skupper listener create database 5432"
+
+	listenerTypes = []string{"tcp"}
 )
 
+type ListenerFlags struct {
+	routingKey   string
+	host         string
+	tlsSecret    string
+	listenerType string
+}
+
 type CmdListenerCreate struct {
-	client   *client.VanClient
-	CobraCmd cobra.Command
+	client     skupperv1alpha1.SkupperV1alpha1Interface
+	CobraCmd   cobra.Command
+	flags      ListenerFlags
+	namespace  string
+	name       string
+	port       int
+	KubeClient kubernetes.Interface
 }
 
 func NewCmdListenerCreate() *CmdListenerCreate {
 
-	skupperCmd := CmdListenerCreate{}
+	skupperCmd := CmdListenerCreate{flags: ListenerFlags{}}
 
 	cmd := cobra.Command{
-		Use:     "create",
-		Short:   "",
+		Use:     "create <name>",
+		Short:   "create a listener",
 		Long:    listenerCreateLong,
 		Example: listenerCreateExample,
 		PreRun:  skupperCmd.NewClient,
 		Run: func(cmd *cobra.Command, args []string) {
 			utils.HandleErrorList(skupperCmd.ValidateInput(args))
-			skupperCmd.InputToOptions()
 			utils.HandleError(skupperCmd.Run())
+		},
+		PostRunE: func(cmd *cobra.Command, args []string) error {
+			return skupperCmd.WaitUntilReady()
 		},
 	}
 
 	skupperCmd.CobraCmd = cmd
+	skupperCmd.AddFlags()
 
 	return &skupperCmd
 }
 
-func (cmd *CmdListenerCreate) NewClient(cobraCommand *cobra.Command, args []string) {}
-func (cmd *CmdListenerCreate) AddFlags()                                            {}
-func (cmd *CmdListenerCreate) ValidateInput(args []string) []error                  { return nil }
-func (cmd *CmdListenerCreate) InputToOptions()                                      {}
-func (cmd *CmdListenerCreate) Run() error                                           { fmt.Println("Not implemented yet."); return nil }
-func (cmd *CmdListenerCreate) WaitUntilReady() error                                { return nil }
+func (cmd *CmdListenerCreate) NewClient(cobraCommand *cobra.Command, args []string) {
+	kubeconfig := os.Getenv("KUBECONFIG")
+	cli, err := client.NewClient("", "", kubeconfig)
+	utils.HandleError(err)
+
+	cmd.client = cli.GetSkupperClient().SkupperV1alpha1()
+	cmd.namespace = cli.Namespace
+	cmd.KubeClient = cli.KubeClient
+}
+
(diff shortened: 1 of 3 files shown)


internal/cmd/skupper/listener/listener_create.go
```diff
@@ -1,47 +1,188 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var (
-	listenerCreateLong    = ""
-	listenerCreateExample = ""
+	listenerCreateLong    = "Clients at this site use the listener host and port to establish connections to the remote service."
+	listenerCreateExample = "skupper listener create database 5432"
+
+	listenerTypes = []string{"tcp", "http"}
```
nluaces, line 22: I would leave only `tcp` for now

internal/cmd/skupper/listener/listener_create.go
```diff
@@ -1,47 +1,188 @@
...
+	utils.HandleError(err)
+	cmd.client = cli.GetSkupperClient().SkupperV1alpha1()
+	cmd.namespace = cli.Namespace
+}
+
+func (cmd *CmdListenerCreate) AddFlags() {
+	cmd.CobraCmd.Flags().StringVarP(&cmd.flags.routingKey, "routing-key", "r", "", "The identifier used to route traffic from listeners to connectors")
+	cmd.CobraCmd.Flags().StringVar(&cmd.flags.host, "host", "", "The hostname or IP address of the local listener")
+	cmd.CobraCmd.Flags().StringVar(&cmd.flags.tlsSecret, "tls-secret", "", "The name of a Kubernetes secret containing TLS credentials")
+	cmd.CobraCmd.Flags().StringVarP(&cmd.flags.listenerType, "type", "t", "tcp", "The listener type. Choices: [tcp|http].")
```
nluaces, line 77: `Choices: tcp`

internal/cmd/skupper/listener/listener_create.go
```diff
@@ -1,47 +1,188 @@
...
+}
+
+func (cmd *CmdListenerCreate) ValidateInput(args []string) []error {
+	var validationErrors []error
+	resourceStringValidator := validator.NewResourceStringValidator()
+	tlsStringValidator := validator.NewTlsStringValidator()
```
nluaces, line 83: Because the secret is the name of an existent kubernetes resource, in this case a secret, I would remove the tlsStringValidator and check if the secret exists in the namespace. To do this, the  CmdListenerCreate will need to have a kubeClient

pkg/utils/validator/simple_validator.go
```diff
@@ -37,6 +37,18 @@ func NewResourceStringValidator() *StringValidator {
...
+	re, err := regexp.Compile("^[A-Za-z0-9=:. ]+$")
+	if err != nil {
+		fmt.Printf("Error compiling regex: %v", err)
+		return nil
+	}
+	return &StringValidator{
+		Expression: re,
+	}
+}
+
```
nluaces, line 51: Because the --tls-secret flag is the name of an existent kubernetes secret, we won't need this; we need to check that the secret exists in the namespace.

internal/cmd/skupper/listener/listener_create_test.go
```diff
@@ -1 +1,450 @@
...
+	"github.com/spf13/pflag"
+	"gotest.tools/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	testing2 "k8s.io/client-go/testing"
+)
+
+func TestCmdListenerCreate_NewCmdListenerCreate(t *testing.T) {
+
+	t.Run("create command", func(t *testing.T) {
```
nluaces, line 18: "listener command"

internal/cmd/skupper/listener/listener_create.go
```diff
@@ -1,47 +1,188 @@
...
+			Type:           cmd.flags.listenerType,
+		},
+	}
+
+	_, err := cmd.client.Listeners(cmd.namespace).Create(context.TODO(), &listener, metav1.CreateOptions{})
+	return err
+}
+
+func (cmd *CmdListenerCreate) WaitUntilReady() error {
+	err := utils.NewSpinner("Waiting for Listener...", 5, func() error {
```
nluaces, line 168: Instead of "Waiting for Listener..." -> "Waiting for status..."

(I missed this documentation change in my site create command implementation, I will update it as well there)

internal/cmd/skupper/listener/listener_create.go
lynnemorrison: done

internal/cmd/skupper/listener/listener_create.go
lynnemorrison: done

internal/cmd/skupper/listener/listener_create.go
lynnemorrison: I've created the kubeClient,  but am having trouble figuring out how to create a fake kubeClient for the unit test.

pkg/utils/validator/simple_validator.go
lynnemorrison: removed.     Should we have a validator for the tls-secret name.   I'm not sure what characters are allowed in this name?

internal/cmd/skupper/listener/listener_create_test.go
lynnemorrison: done

internal/cmd/skupper/listener/listener_create.go
lynnemorrison: done

internal/cmd/skupper/listener/listener_create.go
lynnemorrison: done, thanks Neo

pkg/utils/validator/simple_validator_test.go
```diff
@@ -45,6 +46,34 @@ func TestStringValidator_Evaluate(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+
+			stringValidator := NewTlsStringValidator()
+			expectedResult := test.result
+			actualResult, _ := stringValidator.Evaluate(test.value)
+			assert.Assert(t, reflect.DeepEqual(actualResult, expectedResult))
+		})
+	}
+}
+
```
nluaces, line 76: If the TlsStringValidator is removed this test should be deleted as well

internal/cmd/skupper/listener/listener_create.go
```diff
@@ -1,47 +1,197 @@
...
+		cmd.name = args[0]
+		ok, err := resourceStringValidator.Evaluate(cmd.name)
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("listener name is not valid: %s", err))
+		}
+
+		cmd.port, err = strconv.Atoi(args[1])
+		if err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("listener port is not valid: %s", err))
+		}
```
nluaces, line 111: there is no need to do this conversion here, because the `numberValidator` is already doing it

lynnemorrison: I might be missing interpreting your comment above.  NewNumberValidator looks like it needs to be passed an integer.  I don't see how the code is doing the ascii conversion of value being passed to an integer.   I also need the value converted to save in cmd.port to be used in the Run function.  

I removed the above code and passed args[1] directly to the validator
	ok, err = numberValidator.Evaluate(args[1])
		if !ok {
			validationErrors = append(validationErrors, fmt.Errorf("listener port is not valid: %s", err))
		}

if I directly pass arg[1] to the function I get the following errors in unit test
=== RUN   TestCmdListenerCreate_ValidateInput/listener_type_is_not_valid

    listener_create_test.go:267: assertion failed: 

        --- actualErrorsMessages

        +++ test.expectedErrors

          []string{

        - 	"listener port is not valid: value is not an integer",

          	"listener type is not valid: value not-valid not allowed. It shou"...,

          }

nluaces: Oh, sorry for my confusion, you are right. Please ignore this comment.

nluaces (approved):
Waiting for this to be merged in this pull request: https://github.com/skupperproject/skupper/pull/1516

Just to point out that this pull request looks good to me.

lynnemorrison:
have merged all the listener commands into one commit.   This PR can be closed.

PR #1511, opened 2024-06-20
Non-kube phase 0

fgiorgetti:


cmd/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+		}
+	} else {
+		binary := "podman"
+		if platform == types.PlatformSystemd {
+			binary = "skrouterd"
+		} else if platform == types.PlatformDocker {
+			binary = "docker"
+		}
+		err := exec.Command("command", "-v", binary).Run()
```
c-kruse, line 72: This seems to work on my fedora workstation, but in general os/exec + shell builtins is a recipe for pain. If it must be done this way, I'd suggest exec.Command("/usr/bin/sh", "-c", ...). I think there may actually be a stdlib function for this exact check, though.

fgiorgetti: Yes, `exec.LookPath` is definitely a better choice. I will update it.

fgiorgetti: Done. Thank you @c-kruse.

cmd/bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,141 @@
+set -Ceu
+
+IMAGE="quay.io/skupper/bootstrap"
```
nluaces, line 3: Is the version of this image relevant to the manifest?

fgiorgetti: Certainly! I am leaning towards using the same approach that is used by the install.sh script to
determine the version to be installed.
I will raise a tracker to handle the bootstrap image version in the bootstrap procedure and will
work on it once we have our first official version out.

c-kruse (approved):
Really neat! Was able to take it for a small test drive on podman with a Site+Listener+RouterAccess and had it working with a kube site.

Karen-Schoener (approved):
With today's checkin for cgroup issues, bootstrap (with podman) ran fine on my fedora and rhel virtualbox VMs.  Thanks!

nluaces (approved):

pkg/certs/certs.go
```diff
@@ -214,8 +214,12 @@ func GenerateSecret(name string, subject string, hosts string, ca *corev1.Secret
 }
 
 func GenerateSecretWithExpiration(name string, subject string, hosts string, expiration time.Duration, ca *corev1.Secret) corev1.Secret {
-	caCert := getCAFromSecret(ca)
-	return generateSecret(name, subject, hosts, &caCert, expiration)
+	var caCert *CertificateAuthority
```
grs, line 217: Minor issue: feels to me like it would be nicer to change getCAFromSecret to return nil if the secret is nil or does not contain a valid CA.

grs (approved):

pkg/certs/certs.go
fgiorgetti: I agree @grs. It is better to transfer the responsibility to `getCAFromSecret` instead.
Done within the last commit.

PR #1517, opened 2024-06-20
Network console collector

c-kruse:
Reworks the flow collector such that it can be independently deployed as
the network-console-collector.
* Simplifies startup - relying on command arguments and a messaging.json
  config instead of platform-specific environment inference. Minimizes
  skupper package dependencies.
* Removes all authentication concepts from the collector api for now: I
  did not want to commit to the old configuration interface for them and
  their implementation does not pass for me.
* Adds example deployment configurations


cmd/network-console-collector/resources/openshift/prometheus.yaml
```diff
@@ -0,0 +1,207 @@
...
+    global:
+      scrape_interval:     15s
+      evaluation_interval: 15s
+    alerting:
+      alertmanagers:
+        - static_configs:
+            - targets:
+    rule_files:
+    scrape_configs:
+      - job_name: network-console-service
```
c-kruse, line 85: minor detail: in the original prometheus scrape configuration we just had a static target for the service name and used tls_config.insecure_skip_verify. This version uses service discovery instead because:
1. TLS: openshift service certs are signed for the fully qualified name (e.g. mysvc.namespace.svc.cluster.local instead of mysvc.) Getting the target from the kube api lets us get that fully qualified name without templating the yaml.
2. Include labels from the service resource similar to the default prometheus community helm chart setup - which may be nice if any users want to use federate or remote write metrics from this instance to somewhere central.

cmd/network-console-collector/resources/openshift/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+          name: network-console-certs
+      volumes:
+      - name: network-console-certs
+        secret:
+          defaultMode: 420
+          secretName: network-console-certs
+      - name: skupper-local-client
+        secret:
+          defaultMode: 420
+          secretName: skupper-local-client
```
c-kruse, line 88: Depends on the skupper-local-client secret being present - which is thus far only present in skupper 1. That means all of these examples only run alongside v1 for now.

The name of this static target will likely change in v2.

cmd/network-console-collector/config.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Config struct {
+	APIListenAddress     string
+	APIDisableAccessLogs bool
+	TLSCert              string
+	TLSKey               string
+
+	EnableConsole   bool
+	ConsoleLocation string
+	PrometheusAPI   string
```
c-kruse, line 17: Doubt: The collector api works as a reverse proxy for the prometheus v1 http api so that the console ui can run queries. If we aspire to support more than our very rudimentary unsecured network-console-prometheus instance this may become complicated. Think TLS config, different auth methods, etc.

cmd/network-console-collector/config.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import (
+	"encoding/json"
+	"os"
+	"time"
+)
+
+type Config struct {
```
c-kruse, line 9: note: Notice I've nixed the "console-auth" option for now. On openshift oauth still works mostly like it did before (assuming it is configured with the ouath-proxy.)

The only material loss here is basic auth. I'd entertain adding basic back in as an option if our hands are tied, but would want to rethink how it gets configured and implemented. Leaving it to the user to limit access by exposing the console in some restricted network space or using some ingress with custom authentication (like openshift oauth-proxy) is my preference.

cmd/network-console-collector/resources/openshift/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+        image: quay.io/openshift/origin-oauth-proxy:4.14.0
+        imagePullPolicy: Always
+        args:
+        - --https-address=:8443
+        - --provider=openshift
+        - --openshift-service-account=network-console-collector
+        - --upstream=http://127.0.0.1:8080
+        - --tls-cert=/etc/tls/proxy-certs/tls.crt
+        - --tls-key=/etc/tls/proxy-certs/tls.key
+        - --cookie-secret=SECRET
```
c-kruse, line 62: Note: this hardcoded secret is important to fix.

nluaces:
I think that it is necessary to do also the refactor in `images.go` and `image_utils.go` files (at least); in order to provide the information of what image version of the different skupper components is being used by default (or configured by the user through environment variables). 

usage: version, manifest

c-kruse:
> I think that it is necessary to do also the refactor in `images.go` and `image_utils.go` files (at least); in order to provide the information of what image version of the different skupper components is being used by default (or configured by the user through environment variables).
> 
> usage: version, manifest

Thanks! Good catch. Probably ties in a little to a conversation we'd best have sometime coming up: should skupper know about the network console at all? i.e. should this image stick around in the manifest, or should we just remove the flow-collector

nluaces:
> should skupper know about the network console at all? i.e. should this image stick around in the manifest, or should we just remove the flow-collector

Fair enough! In my opinion, maybe it is not that critical for the version command as it is no longer a sidecar of the controller, but maybe it would still be helpful for the manifest -release wise- for traceability.

cmd/network-console-collector/resources/README.md
```diff
@@ -0,0 +1,58 @@
...
+      http v1 api endpoint for the network-console-collector to query (e.g.
+      prometheus, thanos.)
+
+## OpenShift deployment:
+
+Should be a batteries included experience. Depends on the OpenShift service-ca
+Operator for provisioning certificates and on the OpenShift oauth proxy for
+authentication.
+
+1. Make sure skupper is running in the current context's namesapce. `skupper status`.
```
nluaces, line 35: `skupper site status`

nluaces (approved):

cmd/network-console-collector/resources/static/prometheus.yml
```diff
@@ -0,0 +1,11 @@
...
+global:
+  scrape_interval: 15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
+  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
+  # scrape_timeout is set to the global default (10s).
+
+scrape_configs:
+  - job_name: "prometheus"
+    metrics_path: "/metrics"
+    static_configs:
+      - targets: ["network-console:8080"]
```
fgiorgetti, line 11: Name resolution won't be choice on non-kube sites as well.

This could also be generated as part of the bootstrap process
determining the port dynamically based on availability, or we
could assume 127.0.0.1:8080 instead at first.

cmd/network-console-collector/resources/native/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+    matchLabels:
+      app.kubernetes.io/part-of: network-console
+      app.kubernetes.io/name: network-console-collector
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/part-of: network-console
+        app.kubernetes.io/name: network-console-collector
+        app.kubernetes.io/component: server
+    spec:
```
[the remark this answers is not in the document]
c-kruse: Good question: my reasoning in not including a service account was that this does not interact with the kube api. I am not sure what standard procedure is. On the openshift example I ended up using a service account (with no attached roles) in order to allow the oauth-proxy operate.

cmd/network-console-collector/resources/native/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+  issuerRef:
+    name: network-console-issuer
+    kind: Issuer
+    group: cert-manager.io
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/part-of: network-console
```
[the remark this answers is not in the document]
c-kruse: `skupper-network-console` feels right! Will update

c-kruse:
@nluaces @fgiorgetti - a few of your comments are on getting these working with v2. I had written the deployment examples under /resources to run alongside skupper v1 as sort of proof of concepts since v2 is just shy of ready.

For kube sites I believe we just need the skupper-local-client secret replacement to get provisioned, and for non-kube I think we could start looking at some of fernando's ideas once phase 0 is merged.

Thoughts on adding issues to come back and update the examples once those bits get brought in? Otherwise we could park this for a week or so until it could be ran alongside v2.

cmd/network-console-collector/config.go
bartoval: I agree that we can do better than Auth basic, but it may not be worth putting in too much effort. 
However, the reason we use a minimum level of protection is not for the console but to protect the APIs, and not just those of the collector, which honestly don't seem to contain sensitive data, but those of Prometheus. In the past, we had an issue with QE because we were publicly exposing the Prometheus route, which means exposing sensitive data if someone do a query. So, I think we will face the same problem if we don't protect the APIs from the collector in a situation without OCP auth proxy.

bartoval: Instead if we want to think about a solution maybe we keep in stanby keykloack but it would be another bomb to maintain and manage. Keycloak is currently used in Skupper X by Ted to grant API ACLs based on potential user permissions

cmd/network-console-collector/config.go
bartoval: I agree, we you can do better than Auth basic, but that it's probably not worth spending too much effort on at the current status. 
The problem, however, is that the reason why we use a minimum of protection is not for the console but to protect the APIs and not those of the collector which honestly do not seem to me to contain sensitive data but those of Prometheus. We have had a problem in the past with QEs because we were publicly exposing the prometheus route = expose sensitive data. So I think they will raise the same issue if we don't protect the bees for a situation without OCP auth proxy

bartoval: Instead if we want to think about a solution maybe we keep in stanby keykloack but it would be another bomb to maintain and manage. Also I don't think is the direction we want to follow. Howver, just to mentionate something: Keycloak is currently used in Skupper X by Ted to grant ACLs based on potential user permissions (At the end probably will be discarded this solution for something more light)

c-kruse:
@fgiorgetti @nluaces welcome back!

I've swung back around and updated the collector to work with v2 (this involved changing the specifics of how it connects to the router, but otherwise not much has changed with the code), and updated the examples as well.

For now I have punted on an example deployment for podman. While I was able to run this in podman, it was a little rough: like we had discussed, a podman-compose would probably need to be generated.

PR #1520, opened 2024-06-25
Moved libpod generated client code to an external repo

fgiorgetti:
This is just a draft PR that simply removed the need to generate libpod client code
and use it from a separate repository. 

At the moment, the PR is using a temporary repository but the idea is to transfer that
repository into the skupperproject org, once we are good with this change (and with
the repository name).

I am leaning towards, keeping it as: `skupperproject/skupper-libpod`.
Feel free to share your thoughts or concerns.

The repository will also be tagged with the appropriave OpenAPI spec version we are
using to generate clients and models. The dependency on that repo will also be based
on the respective tag, i.e: 4.0.3.


grs (approved):

c-kruse (approved):
+1 for skupperproject/skupper-libpod. IMO I'd take de-cluttering the package path a step further and generate packages at the root level (i.e. `github.com/skupperproject/skupper-libpod/client` instead of including the redundant pkg/libpod)

Looking forward to this one!

fgiorgetti:
> +1 for skupperproject/skupper-libpod. IMO I'd take de-cluttering the package path a step further and generate packages at the root level (i.e. `github.com/skupperproject/skupper-libpod/client` instead of including the redundant pkg/libpod)

I was also considering `/client` initially, but looking at some openapi client examples, I wasn't sure if `/client` and `/models`
in the repository root, would be a good idea. But considering that code is not something we are offering for others to
consume, I am fine moving generated code to the root folder.

Any objection?

nluaces (approved):

c-kruse (approved):

PR #1521, opened 2024-06-28
Configure oauth proxy cookie secret

c-kruse:
The oauth-proxy uses the cookie secret as a key to sign and validate
session cookies. This change generates a random 32 byte key and stores
it in the skupper-console-session secret to be used by the proxy for
this purpose.

Signed-off-by: Christian Kruse <christian@c-kruse.com>

### DIFF
--- a/client/router_create.go
+++ b/client/router_create.go
@@ -52,7 +52,7 @@ func OauthProxyContainer(serviceAccount string, servicePort string) *corev1.Cont
 			"--upstream=http://localhost:" + servicePort,
 			"--tls-cert=/etc/tls/proxy-certs/tls.crt",
 			"--tls-key=/etc/tls/proxy-certs/tls.key",
-			"--cookie-secret=SECRET",
+			"--cookie-secret-file=/etc/skupper-console-session/session_secret",
 		},
 		Ports: []corev1.ContainerPort{
 			{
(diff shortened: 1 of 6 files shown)


c-kruse (commented):
notes for merge:

this was split into two commits because the site spec labels feature was added in 1.5. If we want to back port this to 1.4.x it should go cleanly if we remember to not squash these commits.

grs (approved):

client/router_create.go
```diff
@@ -977,6 +978,16 @@ func (cli *VanClient) GetRouterSpecFromOpts(options types.SiteConfigSpec, siteId
...
 		})
 	}
+
+	if options.EnableFlowCollector && options.AuthMode == string(types.ConsoleAuthModeOpenshift) {
+		sessionCreds, err := kube.GenerateConsoleSessionCredentials(nil)
+		if err != nil {
+			fmt.Println("Error generating console session credentials:", err)
+		}
+		sessionCreds.Labels = options.Labels
+		credentials = append(credentials, sessionCreds)
```
fgiorgetti, line 988: if an error was returned, should it be added here?

fgiorgetti (approved):
Minor question, but looks good to me.

client/router_create.go
c-kruse: That is a good question! Was not sure what to make of the error handling (or the lack there of) in this bit.

Not that I expect us to ever hit this (maybe some error could happen running out of entropy in crypto/rand?), but this did give me some pause. I suspect it'd probably half-provision a site?

Updated so that it would print an error telling the user that it is skipping that secret, but the command should finish - the controller pod would end up hanging waiting on that secret to be provisioned.

c-kruse: tbh it's probably best if we were to just panic. Here's what things look like with a fake EOF error injected

f2d650b:
> [christian@rh skupper] (configure-oauth-proxy-cookie-secret *)$ ./skupper init --enable-flow-collector --enable-console --console-auth openshift
error generating credentials: error generating console session key: EOF
Could not create secret:  Secret "" is invalid: metadata.name: Required value: name or generateName is required
Error: Secret "" is invalid: metadata.name: Required value: name or generateName is required


955edc0:
> [christian@rh skupper] (configure-oauth-proxy-cookie-secret *)$ ./skupper init --enable-flow-collector --enable-console --console-auth openshift
Skipping provisioning for secret skupper-console-session due to error generating credentials: error generating console session key: EOF
Waiting for status.../^C
[christian@rh skupper] (configure-oauth-proxy-cookie-secret *)$ k get pods
NAME                                         READY   STATUS              RESTARTS   AGE
skupper-prometheus-5df4949b9c-nkfnz          1/1     Running             0          16s
skupper-router-5596958665-5jrq7              2/2     Running             0          19s
skupper-service-controller-95c9df59c-c2m7p   0/3     ContainerCreating   0          17s

c-kruse: Added a small refactor to return that error and handle it from the caller.

ajssmith (approved):

PR #1527, opened 2024-07-09
Move client to internal

ajssmith:
In preparation for removing legacy client pkg. Also introduced fake client for unit tests. Updated the site unit tests but not the complement of cli tests.


nluaces (approved):
lgtm

grs (approved):

PR #1537, opened 2024-07-10
Configures the router with tcpConnector processIds

c-kruse:
Adds the new processId configuration option for tcp connectors and sets it as the pod UID when a selector is used to target that pod.

### DIFF
--- a/pkg/site/connector.go
+++ b/pkg/site/connector.go
@@ -9,11 +9,15 @@ import (
 
 func UpdateBridgeConfigForConnector(siteId string, connector *skupperv1alpha1.Connector, config *qdr.BridgeConfig) {
 	if connector.Spec.Host != "" {
-		UpdateBridgeConfigForConnectorWithHost(siteId, connector, connector.Spec.Host, config)
+		UpdateBridgeConfigForConnectorWithHostProccess(siteId, connector, connector.Spec.Host, "", config)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/site/connector.go
```diff
@@ -9,11 +9,15 @@ import (
...
-		UpdateBridgeConfigForConnectorWithHost(siteId, connector, connector.Spec.Host, config)
+		UpdateBridgeConfigForConnectorWithHostProccess(siteId, connector, connector.Spec.Host, "", config)
 	}
 }
 
 func UpdateBridgeConfigForConnectorWithHost(siteId string, connector *skupperv1alpha1.Connector, host string, config *qdr.BridgeConfig) {
+	UpdateBridgeConfigForConnectorWithHostProccess(siteId, connector, host, "", config)
+}
+
+func UpdateBridgeConfigForConnectorWithHostProccess(siteId string, connector *skupperv1alpha1.Connector, host string, processID string, config *qdr.BridgeConfig) {
```
c-kruse, line 20: Not especially proud of how I went about this - having simply piled on the poor site package's public interface.

grs (approved):
Looks fine to me, apart from the extra 'c' in the new function name which is trivial.

fgiorgetti (approved):

pkg/site/connector.go
Karen-Schoener: Nit: typo in function name: Proccess -> Process

PR #1543, opened 2024-07-15
Vanflow foundations

c-kruse:
The code in pkg/flow has become tricky to adapat as it is coupled to the network-console api and large parts of it are gated based on a collector "mode". This is a large departure that moves to a new set of vanflow primitives that I have mostly copied from an independent implementation I had done (https://github.com/c-kruse/vanflow) save for a few adaptations made from lessons learned. My goals here were to make it sustainable to develop multiple vanflow components independently and to allow each piece to be better tested.

@grs
I've got a few comments commit-by-commit.

4e1b91c0275e69a6eaad126ca6224dc2205dfd5b: This is the new set of primitives and is the most polished bit of this draft PR. Maybe noteworthy/contentious, adds two new dependencies that I thought were sensible when this started as an independent implementation - the azure vcabbage fork, and the exponential backoff module I was most familiar with. Also includes a new class of unit/integration tests that were super useful for me: tests ran against a real skupper router (most of which fall back on a mock router.) Not sure if those have a future or not as they are likely to be flakey/resource  dependent.

e263854212548fb09b37d479977484e49b7e8147: Replaces the controller in pkg/flow with a new one in pkg/kube/flow. This was the first and most straight-forward vanflow component to rewrite (most all of the work is done by a `eventsource.Manager`.) I am sure I've made a bit of a mess of the controller and its interface just from not really understanding the existing controller constructs (it was hard to handle deletes since the work queue + cache key off namespace/name and I'm storing records by object uid: by the time I understood that I had done my own controller loop.)

e9519f67af51d8440cc5f1c78955d2978bf8d152: A new "collector-lite" implementation that I have called pkg/kube/flow.StatusSync. The interface is still rough and my plans for it not totally realized (e.g. no need to use vanflow to get records from the controller running in the same process). Was feeling less confident on this one so it is running alongside the old implementation presently: would be interesting we could find an easy way to continuously diff their output. That aside - I hope it is easy enough to reason. My goal was that any of us could pretty confidently manage to adapt context from vanflow records into the network status info.

**EDIT**: Now that I think of it: perhaps both components I put in pkg/kube/flow should get rewritten as a single controller.


grs (approved):
This looks like a good start to me. I switched the new status sync to the skupper-network-status configmap and it works as before. The link records only seem to have the name in them at present. Not sure if that is expected or if e.g. I have the wrong router version (quay.io/skupper/skupper-router:main).

c-kruse:
> The link records only seem to have the name in them at present.

@grs Expected! Was going to sort that out as a followup. I suspect the right thing to do is going to end up being adding the new Link counterpart, RouterAccess, to network status.

PR #1556, opened 2024-07-29
Remove confusing error message from configSyc log

### LINKED ISSUE #1545: Confusing "sync failed" message in config-sync log
There are messages like:
```
2024/07/16 10:43:41 sync failed: Bridge config is not synchronised yet
```
in the config-sync log whenever bridge config changed (i.e. listeners/connectors added or deleted). The reason for the log message is that the `syncConfig()` function always returns value `false` whenever the bridge config changed: https://github.com/skupperproject/skupper/blob/c26bce4079fffbd19a98656713428fc57e037452/cmd/config-sync/config_sync.go#L203.

Is this `false` return value correct here? As all the calls for the sync succeeded at this point it is confusing to see a "sync failed" error message. On the other hand,  if this is a bug and the return value should be `true` here then this `synced` return flag seems to be superfluous as it simply denotes if an error occurred or not.

gabordozsa:
Fixes #1545

### DIFF
--- a/cmd/config-sync/config_sync.go
+++ b/cmd/config-sync/config_sync.go
@@ -217,8 +217,8 @@ func (c *ConfigSync) syncConfig(desired *qdr.BridgeConfig) error {
 	if err != nil {
 		return fmt.Errorf("Error while syncing bridge config : %s", err)
 	}
-	if !synced {
-		return fmt.Errorf("Bridge config is not synchronised yet")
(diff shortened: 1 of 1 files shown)


grs (approved):

grs:
The CI failure is not related to this change.

fgiorgetti (approved):
LGTM

PR #1558, opened 2024-07-30
skupper configures closeConnectionsOnDelete flag

Karen-Schoener:


pkg/qdr/qdr.go
```diff
@@ -501,13 +505,14 @@ type Address struct {
...
-	SslProfile     string `json:"sslProfile,omitempty"`
-	VerifyHostname *bool  `json:"verifyHostname,omitempty"`
+	Name                     string `json:"name,omitempty"`
+	Host                     string `json:"host,omitempty"`
+	Port                     string `json:"port,omitempty"`
+	Address                  string `json:"address,omitempty"`
+	SiteId                   string `json:"siteId,omitempty"`
+	SslProfile               string `json:"sslProfile,omitempty"`
+	VerifyHostname           *bool  `json:"verifyHostname,omitempty"`
+	CloseConnectionsOnDelete *bool  `json:"closeConnectionsOnDelete,omitempty"`
```
gabordozsa, line 515: Could the type be `bool` instead of `*bool` ?  That would simplify the rest of the code.

grs:
Is there a use-case for configuring this per tcpConnector? Why would you want one behaviour for one connector and the other behaviour for another?

gabordozsa:
> Is there a use-case for configuring this per tcpConnector? Why would you want one behaviour for one connector and the other behaviour for another?

The new flag is defined per tcpConnector and tcpListener in the router. This was suggested by the router team as opposed to have a global router config flag.  The justification was that router code changes would be more or less the same.    I'm not aware of any use case though for turning the flag on for some tcpListeners and tcpConnectors but not for some others. In fact, from the MESH project perspective, we need a way to set this flag globally for skupper (e.g. via a new` skupper init ` flag as @Karen-Schoener suggested before).

pkg/qdr/qdr.go
Karen-Schoener: Ah, I see that I misread the corresponding skupper-router PR.  

When I saw this line in the skupper-router PR, I (incorrectly) assumed that the new optional flag was a bool pointer.
My mistake.
` config->terminate_conns   = qd_entity_opt_bool(entity, "closeConnectionsOnDelete", false);`
 
I will update to change CloseConnectionsOnDelete to be a bool.

Karen-Schoener:
Closing in favor of https://github.com/skupperproject/skupper/pull/1564

PR #1565, opened 2024-08-01
Ensure original target ports annotation is not overwritten once set

grs:


nluaces (approved):

fgiorgetti (approved):
Original annotation target ports are being preserved now.
Thank you!

PR #1566, opened 2024-08-01
keep original port assignments within a service

grs:


fgiorgetti (approved):
LGTM

cmd/service-controller/controller.go
```diff
@@ -1039,7 +1039,16 @@ func (c *Controller) updateServiceBindings(required types.ServiceInterface, port
...
+				portMap := portAllocations[required.Address]
+				for _, publicPort := range required.Ports {
+					if targetPort, ok := portMap[publicPort]; ok {
+						ports = append(ports, targetPort)
+						delete(portMap, publicPort)
+					}
+				}
+				for _, remaining := range portMap {
+					ports = append(ports, remaining)
+				}
```
nluaces, line 1051: If the remaining ports (that are not included in required.Ports) are being added anyway to `ports`, why checking the existence of the required ports in `portMap`? Is it because the order is important?

PR #1567, opened 2024-08-02
Check if the site is active using provided functions

### LINKED ISSUE #1559: update site commands to new status subresources
More information about the new values for the status here: https://github.com/skupperproject/skupper/pull/1539

nluaces:
fixes #1559

### DIFF
--- a/internal/cmd/skupper/site/kube/site_create.go
+++ b/internal/cmd/skupper/site/kube/site_create.go
@@ -210,7 +210,7 @@ func (cmd *CmdSiteCreate) WaitUntil() error {
 			return err
 		}
 
-		if resource != nil && resource.Status.StatusMessage == "OK" {
+		if resource != nil && resource.IsActive() {
 			return nil
 		}
 
--- a/internal/cmd/skupper/site/kube/site_create.go
+++ b/internal/cmd/skupper/site/kube/site_create.go
@@ -210,7 +210,7 @@ func (cmd *CmdSiteCreate) WaitUntil() error {
 			return err
 		}
 
-		if resource != nil && resource.IsActive() {
+		if resource != nil && resource.IsConfigured() {
 			return nil
 		}
 
@@ -218,10 +218,10 @@ func (cmd *CmdSiteCreate) WaitUntil() error {
 	})
 
 	if err != nil {
-		return fmt.Errorf("Site %q not ready yet, check the logs for more information\n", cmd.siteName)
+		return fmt.Errorf("Site %q not configured yet, check the logs for more information\n", cmd.siteName)
 	}
 
-	fmt.Printf("Site %q is ready\n", cmd.siteName)
+	fmt.Printf("Site %q is configured. Check the status to see when it is ready\n", cmd.siteName)
 
 	return nil
 }
--- a/internal/cmd/skupper/site/kube/site_update.go
+++ b/internal/cmd/skupper/site/kube/site_update.go
@@ -236,7 +236,7 @@ func (cmd *CmdSiteUpdate) WaitUntil() error {
 			return err
 		}
 
-		if resource != nil && resource.IsActive() {
+		if resource != nil && resource.IsConfigured() {
 			return nil
 		}
 
--- a/internal/cmd/skupper/site/kube/site_update.go
+++ b/internal/cmd/skupper/site/kube/site_update.go
@@ -236,7 +236,7 @@ func (cmd *CmdSiteUpdate) WaitUntil() error {
 			return err
 		}
 
-		if resource != nil && resource.Status.StatusMessage == "OK" {
+		if resource != nil && resource.IsActive() {
 			return nil
 		}
 
--- a/internal/cmd/skupper/connector/kube/connector_create_test.go
+++ b/internal/cmd/skupper/connector/kube/connector_create_test.go
@@ -834,7 +834,7 @@ func TestCmdConnectorCreate_Run(t *testing.T) {
 	}
 }
 
-func TestCmdConnectorCreate_WaitUntilReady(t *testing.T) {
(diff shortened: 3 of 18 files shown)


nluaces:
@grs the `site status` command is just showing the `StatusMessage` in the same way as `kubectl get site`. Is this enough? or would it be better to show the last condition (eg. Configured)?

internal/cmd/skupper/site/kube/site_create.go
```diff
@@ -210,7 +210,7 @@ func (cmd *CmdSiteCreate) WaitUntilReady() error {
 			return err
 		}
 
-		if resource != nil && resource.Status.StatusMessage == "OK" {
+		if resource != nil && resource.IsActive() {
```
grs, line 213: @nluaces @ssorj at present, there are (up to) three conditions associated with site status. The site is considered successfully 'configured'  once the various resources have been created. If the site accepts incoming links (i.e. linkAccess is not 'none', which is the default) the site is considered successfully 'resolved' once the hosts/ports through which it can be linked to are discovered. The site is considered to be 'running' when there is at least one running router pod.

The site is currently considered 'ready' when it is configured, resolved if linkAccess is not none, and running.

The IsActive() function was mostly intended to be used in transitioning from old status to new. It just tests the configured condition. So using that method here is certainly in line with what was there before and not 'wrong'. However, as the function here is WaitUntilReady() I felt it worth starting a conversation on the meaning of 'ready' so we can eventually align.

Is the current site definition of 'ready' what is wanted here? In the case that it waits for resolution, or the images to be pulled, that can take some time. However, once ready then the site is much more likely to be in the usable state the the user would likely expect.

grs (approved):

grs:
> @grs the `site status` command is just showing the `StatusMessage` in the same way as `kubectl get site`. Is this enough? or would it be better to show the last condition (eg. Configured)?

I think showing the status message is the right thing. If we want more detail in that message we can change the controller to do so.

internal/cmd/skupper/site/kube/site_create.go
nluaces: @grs @ssorj some time ago, we discussed that after creating a site, it should be ready for the user to operate freely with it, that was the purpose to make the user wait. In principle, I think that the WaitUntilReady() should wait for the site to pass the isReady condition. I have tested this on my local setup: 

```
~/demo2 » time skupper site create noe --enable-link-access -n north                                  
Waiting for status...
Site "noe" is ready
skupper site create noe --enable-link-access -n north  0.10s user 0.08s system 0% cpu 25.067 total
-----------------------------------------------------------------------------------------------------------------------
~/demo2 » time skupper site create eon --enable-link-access -n south                                  
Waiting for status...
Site "eon" is ready
skupper site create eon --enable-link-access -n south  0.13s user 0.09s system 0% cpu 33.071 total
```
Some times it takes around 25~30 seconds, and other times when my minikube is not working properly, it could take more than 1 minute (which is something that the user may face and subsequently think that skupper is responsible for that). 

I also do think we need to check the existence of a site that is ready with the rest of the commands that rely on that assumption. 

Option A:
- Wait until the site is just configured (isActive)
- Rename `WaitUntilReady` to `WaitUntil`
- Output: "Site "my-site" is configured. Check the status to see when it is ready."
- Check that the site is ready in the rest of commands

Option B:
- Wait until the site is ready after its creation (timeout: 120s)
- Check that the site is ready in the rest of commands

Which option do you prefer? Do you have any other ideas?

lynnemorrison: I like option B.    I also think that we could change the CLI status commands, right now we just show status OK,
Maybe we could show status as configured/resolved/ready.   Right now I always use the kubectl command to 
get status.

lynnemorrison: I like option A.    Have the command wait until configured.   Then when user issues listener/connector commands we check that the site is Ready before allowing command.

I think we need to do the same with the listener/connector commands, I'm currently waiting for the status to show configured and return success.

lynnemorrison: In pkg/apis/skupper/v1alpha1/types.go Should we change IsActive to IsConfigured for the Site?

internal/cmd/skupper/site/kube/site_update.go
```diff
@@ -236,7 +236,7 @@ func (cmd *CmdSiteUpdate) WaitUntilReady() error {
 			return err
 		}
 
-		if resource != nil && resource.Status.StatusMessage == "OK" {
+		if resource != nil && resource.IsActive() {
```
lynnemorrison, line 239: In pkg/apis/skupper/v1alpha1/types.go Should we change IsActive to IsConfigured for the Site?

lynnemorrison (approved):
changes look good!!

internal/cmd/skupper/site/kube/site_create.go
grs: I think I prefer A as well

internal/cmd/skupper/site/kube/site_update.go
grs: I agree, that would be clearer. (Bad naming on my part!)

internal/cmd/skupper/site/kube/site_create.go
nluaces: fixed: a958b42f8caa260a955778454ed2b4404992512c

internal/cmd/skupper/site/kube/site_update.go
nluaces: fixed: a958b42f8caa260a955778454ed2b4404992512c

lynnemorrison (approved):
looks good!!

PR #1568, opened 2024-08-05
improved link status

grs:
Adds the remote site id and name to the status. Link considered ready only when shown operational in flow records.


pkg/apis/skupper/v1alpha1/types.go
```diff
@@ -478,7 +502,8 @@ func (l *Link) setReady(err error) bool {
 }
 
 func (l *Link) isReady() bool {
```
nluaces, line 504: could this function be public?

nluaces (approved):

cmd/controller/controller.go
```diff
@@ -349,8 +352,10 @@ func extractSiteRecords(status network.NetworkStatusInfo) []skupperv1alpha1.Site
 
 				if site, ok := routerAPs[link.Peer]; ok {
 					record.Links = append(record.Links, skupperv1alpha1.LinkRecord{
-						Name:         link.Name,
-						RemoteSiteId: site,
+						Name:           link.Name,
+						RemoteSiteId:   site,
+						RemoteSiteName: siteNames[site],
+						Operational:    strings.EqualFold(link.Status, "up"),
```
c-kruse, line 358: Looks like we probably want to look at line 349 as well!

grs: Yes, thanks, I missed that! Fixed now.

pkg/apis/skupper/v1alpha1/types.go
grs: Certainly, it can be made public when needed. However thinking  little more about it, from the cli we would probably want to check whether the condition existed at all, i.e. whether it was true or false, as if it is false with an error we would want to report that rather than just waiting indefinitely.

c-kruse (approved):

PR #1578, opened 2024-08-07
Sets site-server processes as bound/exposed.

c-kruse:
The flow collector derives site-server processes when skupper does not know about the process handling service traffic. Update this logic so that the site-servers have a processBinding of "bound" - indicating that the process is configured to handle skupper service traffic.


c-kruse:
@bartoval I could use pointers on reproducing the exact issue we think we can resolve by addressing this. I can only think to test this out with a regular client + external service.

bartoval (approved):
+1

c-kruse:
Merging. This is correct behavior, however it turns out to not be the root of the issue we were hoping to resolve.

PR #1581, opened 2024-08-08
 fixes to route handling

### LINKED ISSUE #1577: [development v2] Namespace Scoped Controller Routes Errors
Using this deployment for a namespace scoped skupper v2 controller (with adjustments to WATCH_NAMESPACE): https://gist.github.com/c-kruse/0cf3ed5db3176e8eb330cfaa510f992c

I first ran into errors here.
 https://github.com/skupperproject/skupper/blob/43784cfe7ed6c39185a4e97eda89bd92f34feaf8/pkg/kube/securedaccess/access.go#L304-L305
The controller attempts to use the namespace component of the key as the route name. After "fixing" that the error stopped but access grants stopped working presumably because the route is gone. That's where I left off: I'm sure you've got a better idea of the intentions here.

grs:
Also adds a yaml example for a namespace scoped controller (however at present it requires changing the value of the SKUPPER_CLAIMS_GET_BASE_URL_FROM env var so that the prefix matches the namespace deployed in. I will make changes to avoid needding to do that in a separate PR.

Fixes #1577

### DIFF
--- a/cmd/controller/deploy_namespace_scope.yaml
+++ b/cmd/controller/deploy_namespace_scope.yaml
@@ -1,0 +1,223 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: skupper-controller
+  labels:
+    application: skupper-controller
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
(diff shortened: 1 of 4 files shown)


c-kruse (approved):
Thank you for this @grs. That worked for me!

PR #1584, opened 2024-08-09
Site bundles

fgiorgetti:
This PR introduces the ability to produce site bundles that can be installed to run
with podman, docker or systemd (a local installation of the skupper-router).

It can produce two different site bundle formats. One is a self-extracting shell
script and the other is a tarball that contains an installation script.

To specify the desired format, you need to export the `SKUPPER_PLATFORM` environment
variable and set it to `bundle` (self-extracting) or `tarball`.

After specifying the desired format, you just need to execute the bootstrap (binary)
or the bootstrap.sh (script) inside a directory that contains your site definition,
which is merely a set of Skupper V2 custom resources (CR).

After the bundle generation is completed, it will inform you where you can find the
generated artifact.

Along with this new feature, this PR also performs a couple of other related changes:

* Enhanced bootstrap and remove scripts to comply with Posix Shell
* Renamed non-kube and non_kube packages to nonkube 
* New unit tests
* Fixed failing nonkube unit tests
* Renamed internal environment variable from OUTPUT_PATH to SKUPPER_OUTPUT_PATH


internal/nonkube/bundle_extract.go
```diff
@@ -0,0 +1,69 @@
...
+	return path.Join(s.OutputPath, fmt.Sprintf("skupper-install-%s.sh", s.SiteName))
+}
+
+func (s *SelfExtractingBundle) Generate(tarBall *Tarball) error {
+	var data = new(bytes.Buffer)
+	var err error
+
+	write := func(buf interface{}) error {
+		var size, written int
+		switch buf.(type) {
```
c-kruse, line 27: assigning the result of type assertions can prevent bugs doing type assertions in the case statements below.

ex: https://go.dev/play/p/TSTaFNXFQHR

pkg/nonkube/bundle/site_state_renderer.go
```diff
@@ -0,0 +1,209 @@
...
+	if err = CreateSystemdServices(s.siteState); err != nil {
+		return err
+	}
+	if err = CreateStartupScripts(s.siteState); err != nil {
+		return err
+	}
+	if err = s.createBundle(); err != nil {
+		return err
+	}
+	if err = s.removeSiteFiles(); err != nil {
```
c-kruse, line 86: empty branch

internal/nonkube/bundle_extract.go
fgiorgetti: Fixed. Thanks for the feedback.

pkg/nonkube/bundle/site_state_renderer.go
fgiorgetti: D'oh! :) I am still wondering why?
Fixed.

fgiorgetti:
Added a README file under `cmd/bootstrap` explaining how to bootstrap
non-kube sites and site bundles.
I hope it is helpful.

internal/nonkube/command.go
```diff
@@ -0,0 +1,29 @@
+package nonkube
```
ajssmith, line 1: Should this be place in a utils package as it is not specific to nonkube?

internal/nonkube/bundle_common.go
```diff
@@ -0,0 +1,21 @@
+package nonkube
```
ajssmith, line 1: Suggest bundle as its own package and putting files into the package directory as other nonkube packages might show up over time

ajssmith (approved):

internal/nonkube/command.go
fgiorgetti: The tarball is also generic, maybe they should both move.

I was a bit skeptical about a generic `internal/utils`, but I think it
makes sense to place them in a more generic location.

I will move them to `internal/utils` for now, unless someone has a
better suggestion.

nluaces (approved):
Looks good to me! and thank you for providing the README file.

Karen-Schoener (approved):

grs (approved):

PR #1586, opened 2024-08-12
Add collector api specification

c-kruse:
Includes the specification as well as a change to the collector to host a swagger editor + spec. Does not include implementation using spec-generated interfaces.

### DIFF
--- a/cmd/network-console-collector/README.md
+++ b/cmd/network-console-collector/README.md
@@ -1,0 +1,34 @@
+# Network Console Collector
+
+The network console collector is the main component of the network console
+application, serving as a bridge between skupper network telemetry and the
+user. When installed alongside a skupper site it will collect operational data
+from ALL sites in the network and expose them via the API and metrics that back
+the [Skupper Console](https://github.com/skupperproject/skupper-console) web
+application.
+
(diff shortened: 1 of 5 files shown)


c-kruse:
@bartoval  this includes some of the adjustments we had discussed but not all.

notably
* I have yet to extract `parent` from baseRecord.
* document still in yaml - if we want to store it in json i'd like uniform tooling to convert to/from yaml to use with the swagger editor.

bartoval:
> @bartoval this includes some of the adjustments we had discussed but not all.
> 
> notably
> 
> * I have yet to finish extracting `parent` from baseRecord. Needs readded to the applicable resources
> * document still in yaml - if we want to store it in json i'd like uniform tooling to convert to/from yaml to use with the swagger editor.

I think we can continue using YAML. Converting the file to JSON or lint potential errors should be easy
We tried this one and it works withour removing refs: 
https://www.geeksforgeeks.org/yaml-to-json-converter/
https://www.geeksforgeeks.org/json-to-yaml-converter/

I suppose Every AI tool similar to CHAT GPT can check syntax or convert in JSON easly

PR #1587, opened 2024-08-12
defer spinner stop

### LINKED ISSUE #1585: Spinner breaks my shell cursor.
**Describe the bug**
When a skupper command with a spinner doesn't complete successfully (usually I get impatient and ^C it) the cursor in my shell disappears and I have to run `reset` to get it back (though I'm sure there's probably a more clever way to go about it.)

I suspect that it is our usage of the spinner library sending control codes but not cleaning up after ourselves.

**How To Reproduce**
Steps to reproduce the behavior: Run a long skupper command with a spinner like `skupper init` and kill the process.

otherwise run this little gizmo.
```
package main
import (
        "context"
        "fmt"
        "os"
        "os/signal"
        "github.com/skupperproject/skupper/pkg/utils"
)
func main() {
        ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
        defer cancel()
        err := fmt.Errorf("nope")
        go func() {
                if err := utils.NewSpinner("kill this process and notice your bash cursor being missing", 20, func() error { return err }); err != nil {
                        fmt.Printf("spinner finished with error: %s\n", err)
                }
        }()
        <-ctx.Done()
        fmt.Println("exiting")
}
```

**Expected behavior**
I'd love for it to clean up after itself consistently.

**Environment details**
 - Skupper CLI: 1.7.3

**Additional context**

nluaces:
Fixes #1585

EDIT:
- Created a context that cancels on interrupt
- Set up signal handling to cancel context on SIGINT or SIGTERM

### DIFF
--- a/pkg/utils/spinner.go
+++ b/pkg/utils/spinner.go
@@ -8,14 +8,13 @@ import (
 func NewSpinner(message string, maxRetries int, function func() error) error {
 
 	spin := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+	defer spin.Stop()
 	spin.Prefix = message
 	spin.FinalMSG = message + "\n"
 
 	spin.Start()
 
 	err := RetryError(time.Second, maxRetries, function)
-
-	spin.Stop()
 
 	if err != nil {
 		return err
--- a/pkg/utils/spinner.go
+++ b/pkg/utils/spinner.go
@@ -1,11 +1,24 @@
 package utils
 
 import (
+	"context"
 	"github.com/briandowns/spinner"
+	"os"
+	"os/signal"
 	"time"
 )
 
 func NewSpinner(message string, maxRetries int, function func() error) error {
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, os.Interrupt, os.Kill)
+	go func() {
+		<-signals
+		cancel()
+	}()
 
 	spin := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
 	defer spin.Stop()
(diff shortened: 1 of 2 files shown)


c-kruse (commented):
Afraid the issue is probably not so tidy to clear up.

In go the default signal handler causes the program to exit. Deferring the spin.Stop() function does not change any behavior here because deferred functions are not ran on exit. What would need to happen is for the cli to tap into the [signal](https://pkg.go.dev/os/signal) package to change what happens with interrupts, and in this case pipe in a signal to this bit of logic to let it know it is time to tidy up - most often done with a context object.

pkg/utils/spinner.go
```diff
@@ -1,21 +1,35 @@
...
+	"context"
 	"github.com/briandowns/spinner"
+	"os"
+	"os/signal"
 	"time"
 )
 
 func NewSpinner(message string, maxRetries int, function func() error) error {
 
+	ctx, cancel := context.WithCancel(context.Background())
```
c-kruse, line 13: I am fairly certain there is a shortcut to this.

```
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill)
defer stop()
```

c-kruse: That said, I do think there are some dragons in registering signal handlers here. May be well enough for v1, since it does solve the problem. Just want to bring up the dangers in case you judge it not worth it.

Those dragons: it is probably best practice to leave signal handling to `main` (or somewhere obvious and close to main.) Registering and un-registerting signal handlers is a global thing. Were there multiple things happening (a nonsense example would be two spinners on stdout and stderr) one of them would "win". Or if we installed signal handlers in main, I believe this would end up stomping those.

skitt:
Instead of trying to handle cleanup in all cases (which is a worthwhile goal ultimately), the cursor issue can be addressed by not hiding it: #1616.

PR #1592, opened 2024-08-14
Fix image-pull test when not using :main tag

hash-d:
The test was creating an invalid image name by replacing `:main` by `:invalid`.  However, if the image was not using `:main` in the first place, nothing was replaced, and the generated name still pointed to a valid image.

### DIFF
--- a/client/podman/container_test.go
+++ b/client/podman/container_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"regexp"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


client/podman/container_test.go
```diff
@@ -116,7 +117,8 @@ func TestContainer(t *testing.T) {
 	// Pulling image
 	t.Run("image-pull", func(t *testing.T) {
 		assert.Assert(t, cli.ImagePull(ctx, image))
-		invalidImage := strings.Replace(images.GetSiteControllerImageName(), ":main", ":invalid", 1)
+		tagOrDig := regexp.MustCompile("(@.*$|:[-a-zA-Z_.]*$|$)")
```
fgiorgetti, line 120: One minor suggestion here. Also include `0-9` and maybe just move end character
after the group, like: `"(@.*|:[-a-zA-Z0-9_.]*|)$"`.
This might be helpful while testing from specific release branches, where the actual version
tag is defined in images.go instead of main.

hash-d: @fgiorgetti, thanks for catching  `0-9`, I have added it.

The `$` is inside the group to catch the scenario where the image name has no tag (`:`) or digest (`@`).  That's why I have that `|$)`: if no tag nor digest, it still adds the `:` .

fgiorgetti (approved):

PR #1596, opened 2024-08-15
Reduce excessive sleep time in unit tests

c-kruse:
All of our unit tests could be rather quick to run, but many of them sleep for a long time and hold the test suite up. Without changing any of the tests other than tuning some durations in the utils retry tests, and running tests that repeatedly sleep in parallel I got the test runtime down under 20s for this branch as compared to 60s+ on v2. (excluding a test that was broken for me locally `go test -count=1 -skip TestContainer ./...`)

I think in general we could use some thought on how the _WaitUntil tests had aught to work. Many of them take 3+ seconds each even when ran in parallel.

### DIFF
--- a/internal/cmd/skupper/connector/kube/connector_create_test.go
+++ b/internal/cmd/skupper/connector/kube/connector_create_test.go
@@ -910,6 +910,7 @@ func TestCmdConnectorCreate_WaitUntil(t *testing.T) {
 	}
 
 	for _, test := range testTable {
+		test := test
 		cmd, err := newCmdConnectorCreateWithMocks("test", test.k8sObjects, test.skupperObjects, test.skupperErrorMessage)
 		assert.Assert(t, err)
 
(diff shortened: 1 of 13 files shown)


c-kruse:
Hm, the affect in CI was less dramatic than I had hoped: failed in 1m 25s instead of 1m 57s.

fgiorgetti (approved):

c-kruse:
@lynnemorrison thank you! Changed those as well.

lynnemorrison (approved):

internal/cmd/skupper/connector/kube/connector_create_test.go
```diff
@@ -910,6 +910,7 @@ func TestCmdConnectorCreate_WaitUntil(t *testing.T) {
 	}
 
 	for _, test := range testTable {
+		test := test
```
nluaces, line 913: why is this needed?

c-kruse: It is quite tricky. These redeclare the test variable in the correct scope - a quirk of go that is changing in the language with 1.22 https://go.dev/blog/loopvar-preview

It is one of the reasons I prefer to not use t.Parallel(), but felt it was worth it in these cases.

nluaces (approved):

PR #1601, opened 2024-08-19
Assorted fixes on -use-cluster tests (#1391)

hash-d:
- Use routes as ingress, when available
- Ignore unexpected additional clusterroles, when running as -use-cluster
- Shutdown previous informers between iterations
- Ignore additional cm openshift-service-ca.crt, along with kube-root-ca.crt

### DIFF
--- a/pkg/domain/podman/link_test.go
+++ b/pkg/domain/podman/link_test.go
@@ -6,9 +6,13 @@ package podman
 import (
 	"context"
 	"fmt"
+	"log"
+	"net"
+	"net/url"
 	"testing"
 	"time"
 
+	"github.com/skupperproject/skupper/pkg/utils"
 	"gotest.tools/assert"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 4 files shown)


hash-d:
Note that the checks failed because the branch main is pointing to skupper-router `:main image`, which is already tracking for Skupper 2.

I have run both system and integration tests with this branch against an OCP 4.16 cluster, and they worked fine.

Running locally against Minikube, only `TestServiceInterfaceCreate` failed, with an error unrelated to the change that's been made (`Timed out trying to communicate with the API: context deadline exceeded`, whereas the change was to add `openshift-service-ca.crt` to the list of expected/auto-created config maps)

hash-d:
@nluaces, please consider this one before branching for 1.8

nluaces:
> @nluaces, please consider this one before branching for 1.8

@hash-d 1.8 will be created from main, commits are cherry picked when branches are already created.

pkg/domain/podman/link_test.go
```diff
@@ -42,6 +46,27 @@ func TestLinkHandlerPodman(t *testing.T) {
...
+			// On some clouds, it may take a while for the service DNS name to be externally
+			// resolvable.  So, we extract that URL and wait for the name resolution to work
+			// before creating the link.  If anything fails, we just log and keep going, as
+			// that's not the focus of the test; the whole thing  may fail down the road,
+			// but with additional information for debugging.
+			skupperUrl := token.Annotations["skupper.io/url"]
+			if skupperUrl != "" {
+				parsed, err := url.Parse(skupperUrl)
+				if err != nil {
+					log.Printf("The skupper.io/url annotation did not parse as an URL (%q): %v", skupperUrl, err)
```
fgiorgetti, line 59: Why not failing the test here if the annotation does not contain a valid url?

hash-d: @fgiorgetti, my intention was to make the changes to address the problem at hand only (link creation being requested before the DNS name was resolvable, making the test flaky on some platforms).

If I addressed an invalid URL on the test code, I might have given an early warning of a problem on the token generation (good), but I'd have removed the chance for Skupper code to react to that.

At that point, I thought it best to let any errors from bad URLs be processed business as usual.  The additional logging would help on the investigation if that ever happened.

The chance of that happening, though, should be fairly small, and the testing for that better placed elsewhere (ie, on the token generation tests).

fgiorgetti: I agree with your name resolution retry logic. But here the URL defined in the token must be "valid" (which
does not mean resolvable). If skupper is generating a bad URL that cannot be parsed, then it is an error.

hash-d: @fgiorgetti, one middle-ground solution might be to log and mark the test as a failure, but let it continue to see how the link code reacts (`t.Errorf()` instead of `t.Fatal()`).  Perhaps that was what you were proposing?

This way, the token error is not ignored, but the link code is still exercised.

granzoto (approved):
LGTM

pkg/domain/podman/link_test.go
fgiorgetti: My point was just that if Skupper is generating an url that cannot be parsed,
it is a critical token generation error and it should not be ignored.
If the URL is well formed, then it is ok to proceed, even if the destination cannot
be resolved or cannot be reached, but if it is malformed, then it must fail.

PR #1606, opened 2024-08-20
Update the collector to stop errant DNS Lookups

c-kruse:
The collector attempts to resolve a connector's destination to a process record by IP. When the connector is configured with a hostname instead of an IP it should NOT attempt to resolve the hostname to an IP, as there is no guarantee that the collector is running with an identical network environment to the router with the connector in question.

This may impact niche cases. ~~The most likely may be where a collector is running local to a podman site with podman containers exposed by hostname.~~ EDIT: I still think this code path could be in use, but the collector will fallback on other (more sensible) means in the case of podman containers exposed by name.


fgiorgetti (approved):

grs (approved):

c-kruse:
Should resolve https://issues.redhat.com/browse/SKUPPER-2085

Issue #1498, opened 2024-06-06
HandleError: specialize returned error in V2


nluaces:
**Is your feature request related to a problem? Please describe.**
The HandleError is always returning 0 as the exit code.

**Describe the solution you'd like**
Specialize the errors returned by the actual commands, so that the
HandleError can properly determine whether to return 0 or not.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1608, opened 2024-08-21
[v2] The CLI exits with success on an unknown command

nluaces:
**Description of the issue**

```
~/code/skupper-example-hello-world$ skupper init
Error: unknown command "init" for "skupper"
Run 'skupper --help' for usage.
unknown command "init" for "skupper"

~/code/skupper-example-hello-world$ echo $?
0 
```
It should instead exit with an error code.

**Description of the solution**
When handling an error by default the exit command will be `1`.
When handling command validation errors the exit command will be `2`.

Fixes: #1498

### DIFF
--- a/internal/cmd/skupper/utils/handle_error.go
+++ b/internal/cmd/skupper/utils/handle_error.go
@@ -8,7 +8,7 @@ import (
 func HandleError(err error) {
 	if err != nil {
 		fmt.Println(err)
-		syscall.Exit(0)
+		syscall.Exit(1)
 	}
 }
 
@@ -18,7 +18,7 @@ func HandleErrorList(errList []error) {
 			fmt.Println(err)
 		}
 
-		syscall.Exit(0)
+		syscall.Exit(2)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/cmd/skupper/utils/handle_error.go
```diff
@@ -18,7 +18,7 @@ func HandleErrorList(errList []error) {
 			fmt.Println(err)
 		}
 
-		syscall.Exit(0)
+		syscall.Exit(2)
```
grs, line 21: What motivated the use of 2 as the exit code here? Not saying it is wrong, just curious.

grs (approved):

internal/cmd/skupper/utils/handle_error.go
nluaces: It was based on something said in this article: https://www.redhat.com/sysadmin/exit-codes-demystified

> Exit status 2 appears when there's a permissions problem or a missing keyword in a command or script.

I thought it was kind of related, because the `HandleErrorList` displays all the validation errors found in the command execution.

fgiorgetti: Possibly adding a list of constants to help identifying the purposes of the error codes, might help.

nluaces: Aggreed :+1:

fgiorgetti (approved):

PR #1616, opened 2024-08-26
Avoid hiding the cursor when displaying a spinner

skitt:
This ensures the cursor is still visible if the skupper command is interrupted.

Fixes: #1585


c-kruse (approved):
An excellent find! Thank you.

nluaces (approved):
Thank you!

PR #1627, opened 2024-09-02
Basic Skupper Helm chart

nluaces:
(UPDATED)

### Description 

This Helm chart is designed to simplify the deployment of Skupper in a Kubernetes environment. 
It ensures that the Skupper Custom Resource Definitions (CRDs) and the Skupper controller 
are correctly installed and configured.

### Scope: namespace
To deploy Skupper using this Helm chart, simply run the following command, specifying your 
namespace:

```
helm install skupper-namespace-setup . --set scope=namespace --namespace <your-namespace>
```

If the namespace is not specified it will be deployed in the current namespace.

### Scope: cluster
To deploy Skupper using this Helm chart, simply run the following command:

```
helm install skupper-namespace-setup . --set scope=cluster 
```

### How to uninstall the helm chart
```
helm uninstall skupper-namespace-setup
```


deployments/helm/templates/deploy_namespace_scope.yaml
```diff
@@ -0,0 +1,223 @@
+---
+apiVersion: v1
```
grs, line 2: Should the helm chart install namespace or cluster scoped?

deployments/helm/templates/deploy_namespace_scope.yaml
```diff
@@ -0,0 +1,223 @@
...
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+      - name: controller
+        image: {{.Values.controller.image}}
+        imagePullPolicy: Always
+        env:
+        - name: SKUPPER_ENABLE_GRANTS
+          value: "true"
+        - name: SKUPPER_CLAIMS_GET_BASE_URL_FROM
```
grs, line 180: The configuration has changed a little recently and is now simpler and the Certificate and SecuredAccess resources are no longer required.

deployments/helm/crds/skupper_access_grant_crd.yaml
```diff
@@ -0,0 +1,102 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
grs, line 2: Ideally the CRDs would be defined in one place or copied by the build. Otherwise it is easy to make changes to one copy and not the other.

deployments/helm/templates/deploy_namespace_scope.yaml
fgiorgetti: Would it be more interesting if the helm chart allows user to choose the target scope?
It could assume cluster scope by default but offer a flag (`--set scope=namespace` for example)
to change the default scope?

adibrastegarnia:
It would be nice if we can merge this PR that we can use the helm chart.

c-kruse:
@nluaces I've took your idea here and ran with it a little while trying to learn enough to tackle a network console chart.

Have found it helpful for myself to quickly deploy a controller with specific images/controller config, etc without having to constantly edit deployments by hand and such. I don't think I could go back to working that way at this point. https://github.com/c-kruse/skupper-charts. Feel free to borrow ideas and copy whatever. As Gordon mentioned, the "rough" spot is getting consistent crd definitions and probably controller role rules as well.

I've been using it for a while now so feel like at least some of the kinks have been worked out. It is published as an oci image you can try out. ex: `helm upgrade --install skupper-controller oci://quay.io/ckruse/skupper-charts/skupper --namespace skupper --create-namespace`

deployments/helm/templates/deploy_namespace_scope.yaml
nluaces: @fgiorgetti I'll check this.

deployments/helm/templates/deploy_namespace_scope.yaml
nluaces: I will add a script to update the CRDs in the helm chart.

nluaces:
> I've been using it for a while now so feel like at least some of the kinks have been worked out. It is published as an oci image you can try out. ex: `helm upgrade --install skupper-controller oci://quay.io/ckruse/skupper-charts/skupper --namespace skupper --create-namespace`

@c-kruse I'll take a look! Thanks!

c-kruse (approved):

ajssmith (approved):

PR #1641, opened 2024-09-09
Nonkube sites enhancement to use namespaces

fgiorgetti:
* Using namespaces to isolate sites
* Container names are now prefixed with namespace name
* Systemd services are now prefixed with namespace name
* Bootstrap README file updated (cmd/bootstrap/README.md)
* Bootstrap can now re-initialize a given namespace
* Bootstrap now accepts bundle strategy flag (or SKUPPER_BUNDLE env variable)
* Bundle and Tarball strategies are no longer valid SKUPPER_PLATFORM values
* Bundles can be installed to different and multiple namespaces
* Bootstrap binary and script now use similar flags
* Router names now have a dynamic suffix
* Static status is now added to processed CRs
* Installation summary is clearer now


cmd/bootstrap/bootstrap.go
```diff
@@ -6,53 +6,117 @@ import (
...
+flag is provided.
+
+A namespace is just a directory in the file system where all site specific
+files are stored, like certificates, configurations, the original sources
+(original custom resources used to bootstrap the nonkube site) and
+the runtime files generated during initialization.
+
+In case the path (-p) flag is omitted, Skupper will try to process
+custom resources stored at the sources directory of the default namespace,
+or from the namespace provided through the namespace (-n) flag.
```
c-kruse, line 47: Really like the usage documentation!

The one thing that stood out to me as missing here as well as in the README is where namespaces get installed (`XDG_DATA_HOME`.)

cmd/bootstrap/bootstrap.go
```diff
@@ -77,72 +141,142 @@ func main() {
...
 		bundleSuffix = " (as a distributable bundle)"
+	} else {
+		bundleSuffix = fmt.Sprintf(" on namespace %q", siteState.GetNamespace())
+		// create bootstrap.out file
+		if api.IsRunningInContainer() {
+			outFile, err := os.Stat("/bootstrap.out")
+			if err == nil && !outFile.IsDir() {
+				err = os.WriteFile("/bootstrap.out", []byte(siteState.GetNamespace()), 0644)
+				if err != nil {
+					fmt.Println("Failed to write to bootstrap.out:", err)
```
c-kruse, line 183: is this just informational? or should we do something (like exit) here?

cmd/bootstrap/README.md
```diff
@@ -212,11 +282,29 @@ spec:
...
 ```shell
-cp ${HOME}/.local/share/skupper/sites/west/runtime/token/link-go-west.yaml ./east/links/link-go-west.yaml
+cp ${HOME}/.local/share/skupper/namespaces/west/runtime/link/link-go-west.yaml ./east/link-go-west.yaml
+```
+
+Make sure to update the namespace on the generated `link` that will be used
+in the `east` site. To do that, run:
+
+```shell
+sed -i 's/namespace: west/namespace: east/g' ./east/link-go-west.yaml
```
c-kruse, line 307: When one does `skupper token issue` for a kube site we end up with a AccessToken resource without a namespace set - so that it ends up ready to apply to another site. Would it make sense to do the same for the generated accesstokens for non-kube sites?

cmd/bootstrap/bootstrap.go
fgiorgetti: I agree. This is an important information to both readme and usage, I will add it.

cmd/bootstrap/bootstrap.go
fgiorgetti: Actually, the message is not clear and it needs to be improved.

Basically if the file cannot be written for any reason, the bootstrap.sh script
won't be able to create the systemd service on the host machine, so this needs
to be explained, instead of only showing the error message.

cmd/bootstrap/README.md
fgiorgetti: To make it easier as well, I have also modified the example defined at the README file to have all CRs
defined without a namespace. This way it will be easier to copy the static token from a site to another.

fgiorgetti:
@c-kruse I have addressed the initial feedback.
Let me know if your thoughts.

cmd/bootstrap/README.md
```diff
@@ -41,43 +45,74 @@ unix socket will be used based on the current user and platform selected.
...
-## Usage
-
 ### Bootstrapping and removing non-kubernetes sites
 
 Place all your Custom Resources (CRs) on a local directory.
 You must always provide a `Site` (CR), plus some other resource
 that makes your running site meaningful, like Listeners and/or Connectors.
 
-In case you want your local site to listen for incoming links from other sites,
+In case you want your local site to accept incoming links from other sites,
```
nluaces, line 105: So the RouterAccess CR is only necessary if we want to enable link access in a nonkube site?

nluaces: Ok. New issue for the CLI: #1657

cmd/bootstrap/README.md
```diff
@@ -87,65 +122,98 @@ Now that you have all your CRs placed on a local directory, just run:
...
-To remove your site, you can run a local script, that takes the site name
+To remove your site, you can run a local script, that takes the namespace
 as an argument. The platform is identified by the script so you don't need
 to export it.
 
+If the namespace is omitted, the "default" namespace is used. 
+
 ```shell
-./cmd/bootstrap/remove.sh <site-name>
+./cmd/bootstrap/remove.sh [namespace]
```
nluaces, line 189:
```
$ ./bootstrap -p ~/test-podman -n east                                                                                                                  
Skupper nonkube bootstrap (version: nodeport-899-ge39ee03d)
Site "east" has been created on namespace "east"
Platform: podman
Definition is available at: /home/nluaces/.local/share/skupper/namespaces/east/sources
 

$ ./remove.sh east                                                                                                                       
east-skupper-router
Removed "/home/nluaces/.config/systemd/user/default.target.wants/skupper-east.service".
```
What does it mean `default.target.wants`? the directory `east` was deleted though.

fgiorgetti: This is actually being displayed by systemctl when the systemd service is removed.
I have updated the `remove.sh` script to provide a better response message.

ajssmith (approved):
Built bootstrap locally, ran redis example in myriad ways:
- w/wo namespace
- bundle types
- removal
- re-init
LGTM

grs (approved):

nluaces (approved):

c-kruse (approved):
Very nice @fgiorgetti! Thank you!

PR #1644, opened 2024-09-09
Only copy identity related fields for owner references

### LINKED ISSUE #1642: [v2] Grant autoconfigure error
**Describe the bug**
Starting a new namespace scoped controller with the latest namespace-scoped deployment (`kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/938f13c/cmd/controller/deploy_namespace_scope.yaml`) on openshift I see the following error logged on controller startup, and no certificates get created:

```
Auto configuration of grant server failed: Error creating resources for grant server: certificates.skupper.io "skupper-grant-server-ca" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>
```

I am not sure exactly what is meant to happen here - it looks like we are setting the certificate owner to the pod's owner (a replicaset.) and the kube api is displeased with this?

EDIT: It looks like the deployment/rs controller sets `blockOwnerDeletion: true` on the pod's owner references and we are "copying" that directive.

grs:
fixes #1642

### DIFF
--- a/pkg/kube/grants/autoconfigure.go
+++ b/pkg/kube/grants/autoconfigure.go
@@ -26,7 +26,14 @@ func (s *AutoConfigure) getConfigurationFromPod(clients internalclient.Clients, 
 	if err != nil {
 		return err
 	}
-	s.ownerRefs = pod.ObjectMeta.OwnerReferences
+	for _, or := range pod.ObjectMeta.OwnerReferences {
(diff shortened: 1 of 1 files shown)


c-kruse (approved):
Nice! Thank you @grs

Issue #1645, opened 2024-09-09
config-sync stops syncing network status after losing skupper-site-leader lock


c-kruse:
**Describe the bug**
I suspect that when a router holds the skupper-site-leader lock and _some_ disruption causes it to lose that lock, that router will never attempt to reclaim the lock. This would mean the network status would stop getting updated until that router is restarted (and in v2 the vanflow controller event source would stop producing events.)

* Router starts up and acquires lease
```
2024/09/09 17:01:38 CONFIG_SYNC: Version: v2-release-4-g03e177b
2024/09/09 17:01:38 CONFIG_SYNC: Waiting for Skupper router to be ready
2024/09/09 17:01:43 CONFIG_SYNC: Starting collector...
2024/09/09 17:01:43 CONFIG_SYNC: Starting controller loop...
I0909 17:01:43.424827       1 leaderelection.go:243] attempting to acquire leader lease ck-a/skupper-site-leader...
2024/09/09 17:01:43 CONFIG_SYNC: Waiting for informers to sync...
I0909 17:01:43.478119       1 leaderelection.go:253] successfully acquired lease ck-a/skupper-site-leader
2024/09/09 17:01:43 COLLECTOR: Leader skupper-router-d8fc996dc-8msxt starting site collection after 53.452727ms
```
* config sync carries on, network status info is updated by kube/flow, the vanflow controller runs and sends out events, etc. All is okay.

```
2024/09/09 17:01:43 INFO record message sent component=vanflow.eventsource.manager instance=33a2e9ea-5033-4f22-939b-4c3983d31346 record_count=10
2024/09/09 17:01:44 INFO record message sent component=vanflow.eventsource.manager instance=33a2e9ea-5033-4f22-939b-4c3983d31346 record_count=1
2024/09/09 17:01:44 INFO servicing flush component=vanflow.eventsource.manager instance=33a2e9ea-5033-4f22-939b-4c3983d31346 source=33a2e9ea-5033-4f22-939b-4c3983d31346
2024/09/09 17:01:44 INFO record message sent component=vanflow.eventsource.manager instance=33a2e9ea-5033-4f22-939b-4c3983d31346 record_count=1
2024/09/09 17:01:44 INFO record message sent component=vanflow.eventsource.manager instance=33a2e9ea-5033-4f22-939b-4c3983d31346 record_count=11
2024/09/09 17:01:45 INFO updating network status info component=kube.flow.statusSync configmap=skupper-network-status
[...]

PR #1646, opened 2024-09-09
Make config-sync exit on lost leader election

c-kruse:
When config-sync acquires the skupper-site-leader lock it is expected that it will continue to keep that lease active by periodically checking in. When some disruption causes that check-in to fail the k8s client-go implementation of leaderelection.RunOrDie will return and no further attempts are made to re-acquire the lock.

This change makes it so that when config-sync is running and it fails to hold the leader lease, it will restart itself so that a clean instance can attempt to reclaim the lock.

### DIFF
--- a/cmd/config-sync/collector.go
+++ b/cmd/config-sync/collector.go
@@ -101,60 +101,44 @@ func startFlowController(ctx context.Context, cli *internalclient.KubeClient) er
 	return nil
 }
 
-func runLeaderElection(lock *resourcelock.ConfigMapLock, ctx context.Context, id string, cli *internalclient.KubeClient) {
-	var (
-		leaderCtx context.Context
-		cancel    context.CancelFunc = func() {}
-	)
+func runLeaderElection(lock *resourcelock.ConfigMapLock, id string, cli *internalclient.KubeClient) {
+	ctx := context.Background()
 	begin := time.Now()
 	podname, _ := os.Hostname()
-	limiter := time.NewTicker(time.Second * 15)
-	defer limiter.Stop()
-	for {
-		leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
-			Lock:            lock,
-			ReleaseOnCancel: true,
-			LeaseDuration:   15 * time.Second,
-			RenewDeadline:   10 * time.Second,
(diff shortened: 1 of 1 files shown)


c-kruse:
Closes https://github.com/skupperproject/skupper/issues/1645

Would be glad for suggestions on how I might be able to test this live and/or stubbed out. Maybe fiddling with networking policies at some level to disrupt pod -> kube api traffic?

cmd/config-sync/collector.go
```diff
@@ -108,34 +108,41 @@ func runLeaderElection(lock *resourcelock.ConfigMapLock, ctx context.Context, id
...
-				if current_id == id {
-					// Remain as the leader
-					return
-				}
-				log.Printf("COLLECTOR: New leader for site collection is %s\n", current_id)
-			},
-		},
-	})
+		})
+		if ctx.Err() != nil {
```
grs, line 141: Do we need some kind of sleep or backoff?  If there are two pods competing for the lock, once one gets one, will the other just busy loop?

c-kruse: That's probably a good call! In normal operation it shouldn't be strictly necessary, according to [docs](https://pkg.go.dev/k8s.io/client-go@v0.20.0/tools/leaderelection#RunOrDie) and a skim of the leaderelection code it looks like the only way we will ever hit this loop is when we are able to acquire the lock then fail to renew - so pods competing for the the lock should all be neatly polling at least until they can acquire the lock. I will throw in some limiting out of caution, though. It seems that by far the most popular usage of leaderelection is to call os.Exit from the `OnStoppedLeading` callback (which is not what I think we want), so we are in slightly uncharted waters here.

grs (approved):

cmd/config-sync/collector.go
c-kruse: I suppose I agree! In my observations when we lose this lock it is because it cannot read or update the lock (configmap) within that 10 second lease RenewDeadline, which seems exceptional (as we have been running this logic for years and just encountered this.)

Updated the PR to just log and exit. Want to do a little live testing before I merge.

ajssmith (approved):

PR #1660, opened 2024-09-17
prepare config for circle ci to generate 2.X skupper versions

nluaces:
Changes:
- Allow to generate new 1.X releases as usual (executing integration/policy tests and generating a manifest) with a new filter that selects tags that start by 1.
- Duplication of the release steps for v2: this is necessary because the requirements for publishing images and the name of the images are different from 1.X. Its filter selects tags that start by 2.

The way I see it, this duplication should be maintained until 1.X versions are no longer released.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -360,6 +360,20 @@ yaml-templates:
       tags:
         only: /[0-9].*/
 
+  release_v1_filters: &run_for_v1_tags
+    filters:
+      branches:
+        ignore: /.*/
+      tags:
+        only: /^1.*/
+
+  release_v2_filters: &run_for_v2_tags
+    filters:
+      branches:
+        ignore: /.*/
+      tags:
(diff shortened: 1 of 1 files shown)


nluaces:
Thanks @ajssmith and @fgiorgetti for answering my questions:
- This circle ci configuration will be applied only in this branch, so we don't need to maintain 1.X config here.
- 1.X maintenance will have its own branch (main for now). There is no need for step duplication.

Latest change:
- Enable again release-steps (publish-github-release-images, generate-manifest and publish-github-release-artifacts) in order to be triggered by tags that start with a number, with v2 requirements (build and test steps)

nluaces:
@fgiorgetti @ajssmith is it ok if I merge this?

ajssmith (approved):

fgiorgetti (approved):
LGTM

PR #1662, opened 2024-09-17
V2 make name optional in token issue command

### LINKED ISSUE #1640: [v2] make name optional in kube token command
currently name is required in token command.
skupper token issue <name> ~/token.yaml

The desire is to have the command  remove the required name and add an optional parameter to set name.
skupper token issue ~/token.yaml

The default name will be:
assigning a random UUID as the default name for example:
like west-70dc0eb4 (<sitename>-<randomvalue>).

lynnemorrison:
Make name optional in token issue command.

fixes #1640

### DIFF
--- a/cmd/skupper/README.md
+++ b/cmd/skupper/README.md
@@ -67,7 +67,7 @@ skupper listener create backend 8080 --routing-key backend -n west
 Create a AccessGrant in west site and generate a file with the AccessToken:
 
 ```
-skupper token issue my-token ~/my-token.yaml -n west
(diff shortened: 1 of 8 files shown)


nluaces (approved):
Looks good to me!

Could you update this line 
https://github.com/skupperproject/skupper/blob/21ca026fa6118eb6eacf3f80a0bc5051016cac01/cmd/skupper/README.md?plain=1#L70

in the README file, by removing the token name giving that is no longer necessary?

lynnemorrison:
fixed readme

PR #1663, opened 2024-09-17
V2 Default values for listener and connector commands

lynnemorrison:
Default routing-key in listener and connector commands
Default selector in connector command

if routing-key flag is not set then set the routing-key to equal the name of the listener or connector.  If selector, host or workload are not set in the connector command default to selector to equal "app=connector-name".

### DIFF
--- a/cmd/skupper/README.md
+++ b/cmd/skupper/README.md
@@ -53,13 +53,13 @@ skupper site create site2  -n east
 # Expose backend in east site (create a connector)
 
 ```
-skupper connector create backend 8080 --routing-key backend --selector app=backend -n east
+skupper connector create backend 8080 -n east
 ```
 
 # Consume backend in west site (create a listener)
 
 ```
-skupper listener create backend 8080 --routing-key backend -n west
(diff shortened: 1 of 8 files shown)


nluaces:
>If selector, host or workload are not set in the connector command default to selector to equal "app=".

Do you mean "app=connector-name", right?

nluaces (approved):
Looks good to me!

Could you update these lines of the README?:
https://github.com/skupperproject/skupper/blob/21ca026fa6118eb6eacf3f80a0bc5051016cac01/cmd/skupper/README.md?plain=1#L56

Remove: `--routing-key backend --selector app=backend`

https://github.com/skupperproject/skupper/blob/21ca026fa6118eb6eacf3f80a0bc5051016cac01/cmd/skupper/README.md?plain=1#L62

Remove: `--routing-key backend`
Add: `--host backend`

Thank you!

nluaces:
Oh, sure, you are right, it is better `skupper listener create backend 8080 -n west`

PR #1669, opened 2024-09-20
Use explicit resource description rather than trying to discover this dynamically

### LINKED ISSUE #1668: [v2] Controller runs kube api discovery on each site reconcile
**Describe the bug**
It looks like `pkg/kube/site/resources.Apply` (used in site reconcile) spends a lot of time doing superfluous stuff - querying the discovery API likely.

I believe it probably comes down to this line [here](https://github.com/skupperproject/skupper/blob/148dd3dfce89649973c7d8f679190a957f89f6ce/pkg/kube/site/resources/apply.go#L134) where we are creating a fresh cache for each invocation of this deferred discovery gizmo.

It is my understanding that this is where all of these "Throttling" log lines are coming from: getting rate limited on our discovery api usage - querying apis we have no business with `Throttling request took 5.399145702s, request: GET:https://172.30.0.1:443/apis/topolvm.io/v1?timeout=32s`

grs:
Fixes #1668

### DIFF
--- a/pkg/kube/site/resources/apply.go
+++ b/pkg/kube/site/resources/apply.go
@@ -11,10 +11,9 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


c-kruse (approved):
Thank you @grs! This is great! Already feels snappier.

PR #1674, opened 2024-09-20
V2 update CLI commands to use timer validator

### LINKED ISSUE #1670: [v2] add common timer validation code to CLI commands
update connector, listener, token commands to use the new timerValidator code

lynnemorrison:
Fix connector, listener and token commands to use the new timer validator function.

fixes #1670


internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -96,18 +96,22 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) []error {
...
 	//TBD is there a limit to number of redemptions
-	if cmd.Flags.RedemptionsAllowed < 1 {
+	if cmd.Flags != nil && cmd.Flags.RedemptionsAllowed < 1 {
 		validationErrors = append(validationErrors, fmt.Errorf("number of redemptions is not valid"))
 	}
 
-	//TBD what is valid times
-	if cmd.Flags.ExpirationWindow <= 0*time.Minute {
-		validationErrors = append(validationErrors, fmt.Errorf("expiration time is not valid"))
+	if cmd.Flags != nil && cmd.Flags.ExpirationWindow.String() != "" {
```
nluaces, line 103: If I specified a expiration window of less than 10 seconds this check is not going to pass (which is correct), but I wonder if it is correct to show that the minimum expiration window is 10 seconds, that is an amount of time too small to operate with. We might need an ExpirationWindowValidator (minimum: 5 minutes? 15 minutes?) in the same fashion as the TimeoutValidator, but it can be done in a future pull request.

nluaces (approved):

internal/cmd/skupper/token/kube/token_issue.go
lynnemorrison: agree that makes sense

nluaces: #1729

Issue #1676, opened 2024-09-23
[v2] [nonkube] system service runs router without SKUPPER_SITE_ID set.


c-kruse:
**Describe the bug**
Most of the time skrouterd is ran with SKUPPER_SITE_ID as an environment variable - some of the vanflow stuff depends on it. For systemd it doesn't look like this gets set, so things get weird - mainly in the network console.

PR #1677, opened 2024-09-23
Update nonkube to include router Site config

c-kruse:
Exposes router configuration for including a site vanflow record to be emitted from the router in lieu of a dedicated controller to share this information.

### DIFF
--- a/pkg/qdr/qdr.go
+++ b/pkg/qdr/qdr.go
@@ -22,7 +22,7 @@ type RouterConfig struct {
 	Connectors  map[string]Connector
 	Addresses   map[string]Address
 	LogConfig   map[string]LogConfig
-	SiteConfig  SiteConfig
+	SiteConfig  *SiteConfig
 	Bridges     BridgeConfig
 }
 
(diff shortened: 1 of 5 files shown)


c-kruse:
There's some work to do yet end-to-end before this is perfect:

Shows a kube site with 3 nonkube sites attached (podman, docker and systemd)
image

* version field is tbd - also kind of awkward since we usually expect a control plane version there.
* collector api work to add systemd platform
* the "island" site ls1 there has to do with https://github.com/skupperproject/skupper/issues/1676 - where we do not correctly correlate the router with the site record without a consistent SKUPPER_SITE_ID.
* misc frontend enhancements for adding logos and such

fgiorgetti (approved):
LGTM

pkg/qdr/qdr.go
```diff
@@ -22,6 +22,7 @@ type RouterConfig struct {
 	Connectors  map[string]Connector
 	Addresses   map[string]Address
 	LogConfig   map[string]LogConfig
+	SiteConfig  SiteConfig
```
grs, line 25: Would having a pointer here work and avoid needing the Present flag?

grs (approved):

PR #1678, opened 2024-09-23
Collector api bugfixes

c-kruse:
Quick followup to the updated network-console-collector fixing a few bugs.

* Fix several bugs in "pair" manager - including double sitepair records and fixing processgrouppairs.
* Update our sortBy logic to work with the types generated by openapi-spec instead of only primitives.
* /routers should have parent that points to a site.

### DIFF
--- a/cmd/network-console-collector/internal/collector/connections.go
+++ b/cmd/network-console-collector/internal/collector/connections.go
@@ -580,38 +580,6 @@ func (c *connectionManager) run(ctx context.Context) {
 						continue
 					}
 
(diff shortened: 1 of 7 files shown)


bartoval:
It seems correct to me 👍🏼 .  Perhaps I would add the peername to the links, given that in this PR you add the sourcePGName and destPGName

c-kruse:
Will do a followup for our changes to routerlinks! That's a little larger in scope.

bartoval (approved):

PR #1683, opened 2024-09-24
Simplify current uid retrieval

### LINKED ISSUE #1682: [v2] nonkube sites not detecting user id on ubuntu using bootstrap script
**Describe the bug**
When the bootstrap container is executed on an ubuntu23.04, the UID
is not being properly identified. The bootstrap.sh script itself is detecting
properly, but the container is hitting an internal error:
"Failed to get current user: user: unknown userid 1002"

**How To Reproduce**
Run the bootstrap.sh script on an ubuntu 23.04 VM.
No error will occur, but the router won't have permissions to read
the TLS certificates causing it to be in stuck in a failure state.

**Expected behavior**
Router is using `keep-id` as the namespace mode.

**Environment details**
- Skupper v2 preview 1
- Platform: nonkube

fgiorgetti:
Fixes #1682

### DIFF
--- a/internal/nonkube/client/compat/container.go
+++ b/internal/nonkube/client/compat/container.go
@@ -8,6 +8,7 @@ import (
 	"os/user"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/go-openapi/runtime"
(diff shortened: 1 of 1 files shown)


c-kruse (approved):
FWIW, I was unable to reproduce the issue - tried docker and podman with ubuntu 23.04: maybe there was some other environmental factor?

Just getting the uid instead seems like a fine way to go regardless!

grs (approved):

PR #1684, opened 2024-09-24
Add fields to /routerlinks collector API

c-kruse:
Expose additional context around a link. Adds router name, destination/peer router ID and name, renames peer to routerAccessId for clarity.


bartoval (approved):
🔥

PR #1686, opened 2024-09-25
Adding SKUPPER_SITE_ID to systemd sites

### LINKED ISSUE #1676: [v2] [nonkube] system service runs router without SKUPPER_SITE_ID set.

fgiorgetti:
Fixes #1676.


c-kruse:
Thank you @fgiorgetti

After rebasing this commit on the latest v2 branch I was able to see this working end to end!

c-kruse (approved):

PR #1694, opened 2024-10-01
Document proposed site operations metrics

c-kruse:
A proposal for a first round implementation of some of the operational metrics.

I chose a pretty aggressively coarse view into these signals, cutting out most dimensions aside from `site_id` intentionally. My reasoning is that once a problem is identified with a particular site in a VAN, operators should be able to proceed from that point forward using the cli and/or kube api.

### DIFF
--- a/cmd/network-console-collector/README.md
+++ b/cmd/network-console-collector/README.md
@@ -38,15 +38,34 @@ files in the internal/api package.
 
 ## Metrics
 
-The network console collector exposes a set of prometheus metrics alongside the
+The network console collector exposes a set of Prometheus metrics alongside the
 API at `/metrics`.
 
-### Application Network Traffic
+### Operational Site Metrics
+
+This set of metrics exposes coarse details pertaining to the operation of a
+site.
+
+* `skupper_site_info`: Metadata about the active sites in the network. Labels are `site_id`, `name` and `version`.
+* `skupper_routers_total`: Number of active routers in each site. Labels are `site_id` and `mode`.
+* `skupper_site_links_total`: Number of links from each site. Counts all router
+  links so may not map directly to the number of links.skupper.io resources in
+  the case of HA deployments. Labels are `site_id`, `status` and `role`.
+* `skupper_site_link_errors_total`: Count of link connection errors from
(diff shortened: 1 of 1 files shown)


cmd/network-console-collector/README.md
```diff
@@ -38,16 +38,35 @@ files in the internal/api package.
...
 
-### Application Network Traffic
+### Operational Site Metrics
+
+This set of metrics exposes coarse details pertaining to the operation of a
+site.
+
+* `skupper_site_info`: Metadata about the active sites in the network. Labels are `site_id`, `name` and `version`.
+* `skupper_routers_total`: Number of active routers in each site. Labels are `site_id` and `mode`.
+* `skupper_site_links_total`: Number of links from each site. Counts all router
```
grs, line 51: Just to clarify, this is the total number of currently active links, correct?

cmd/network-console-collector/README.md
```diff
@@ -38,16 +38,35 @@ files in the internal/api package.
...
+  links so may not map directly to the number of links.skupper.io resources in
+  the case of HA deployments. Labels are `site_id`, `status` and `role`.
+* `skupper_site_link_errors_total`: Count of link connection errors from
+  routers in a site. Labels are `site_id` and `role`.
```
grs, line 55: How are the errors detected and for how long will they be reported?

c-kruse: Here was my plan for that:

Routers keep a counter for each link of the transitions to oper status down and expose that through vanflow. The most consistent way I think we can convey that is thorough a counter of our own. So when the collector has a link and observes that counter being incremented it will increment skupper_site_link_errors_total. This way folks can monitor the signal as a rate `rate(skupper_site_link_errors_total[5m])` instead of of some scalar gauge value that is going to behave erratically when the collector is restarted or big topology changes occur.

If people _want_ a scalar to watch for "down" they should be able to get that with `skupper_site_links_total`.

Does that approach make sense?

grs: yes

grs (approved):

PR #1699, opened 2024-10-01
There can be only one link between routers.  Cleanup old link records.

### LINKED ISSUE #1653: ConfigMap skupper-network-status shows incorrect linkCost after quickly deleting/recreating link with new cost
**Describe the bug**
ConfigMap skupper-network-status shows incorrect linkCost after quickly deleting/recreating link with new cost.

Test steps are basically:
  - create link with default cost of 1.
  - observe that linkCost in configmap skupper-network-status shows as 1 (as expected)
  - delete link
  - do not wait ~60 seconds for link record to age out of the skupper-network-status.
  - create link with cost of 10
  - observe that linkCost in configmap skupper-network-status shows as 1

**How To Reproduce**

Create sites: east, west
```
    skupper -n west init
    skupper -n east init
```

Link sites (with default cost):
```
    skupper -n west token create skupper_token_west.out
    skupper -n east link create skupper_token_west.out
```

In east, view configmap skupper-network-status.
Observe that link cost shows as: 1

Delete link.
```
    skupper -n east link delete link1
```

Create link (with cost=10).
Do not wait for link to age out of the skupper-network-status before creating link (with cost=10).
```
    skupper -n west token create skupper_token_west.out
    skupper -n east link create skupper_token_west.out --cost=10
```

In east, view configmap skupper-network-status.
Observe that link cost shows as: 1

**Expected behavior**
Expected behavior: linkCost in configmap skupper-network-status matches the configured link cost.

**Environment details**
 - Skupper CLI: 1.8.0
 - Skupper Operator (if applicable): NA
 - Platform: minikube

lynnemorrison:
When adding a new outgoing link, make sure there are no leftover link records.
If user adds a link, deletes the link and then adds a new link in quick succession
the flow collector doesn't cleanup leftover link records.   This change checks
for any leftover links when a new outgoing one is created and removes them.

fixes #1653


c-kruse (commented):
~~I am a little nervous about the premise here: that two routers can only have a single link between them. I understand the cli steers you away from adding multiple links between sites, but I am fairly confident that it is a valid (if not very useful) router configuration.~~

~~Maybe someone with more familiarity on that side could weigh in for me here? @ted-ross, @grs or @ajssmith.~~

~~I could be convinced if there are not ill effects to this, but reading through the issue my feeling is that the real root of the problem is more central to our design of vanflow: a disconnected event listener cannot receive events about link records being deleted, and cannot know that it has missed those events when the connection is reestablished.~~

**EDIT:** I've been caught up, and am feeling more comfortable with this. Will defer to Andy and/or Gordon for review since it sounds like they've been involved!

pkg/flow/flow_mem_driver.go
```diff
@@ -805,6 +805,15 @@ func (fc *FlowCollector) updateRecord(record interface{}) error {
 		if link, ok := record.(LinkRecord); ok {
 			if current, ok := fc.Links[link.Identity]; !ok {
 				if link.StartTime > 0 && link.EndTime == 0 {
+					// adding a new link, can only have one link between routers
```
ajssmith, line 808: I would suggest removing the "peer" link record when the local link record is deleted. For example, a link for the parent may not get created if there is an overall van topology change. Also, the link name may be the same as the deleted link.

c-kruse: +1 to this suggestion!

I am not certain there will be anything useful to reuse here, but I've somewhat recently added similar logic [here](https://github.com/skupperproject/skupper/blob/72e6df22bc8bc764fd4c4574d56bfe48677fb345/pkg/flow/flow_mem_driver.go#L2798-L2904) for counting links in some of our operational metrics.

pkg/flow/flow_mem_driver.go
```diff
@@ -809,6 +809,20 @@ func (fc *FlowCollector) updateRecord(record interface{}) error {
 				}
 			} else {
 				if link.EndTime > 0 {
+					// incoming link being deleted so remove matching outgoing link
+					if current.Direction != nil && *current.Direction == Incoming {
+						// name of peer router is part of link name
```
c-kruse, line 814: Thank you @lynnemorrison  - I like this approach.

Mind testing this out with a podman site as well first, though? I'm not confident that the way we are correlating a router name to a link parent ID is something the router holds as a stable part of its API. That is, there may be scenarios where there isn't exactly 5 dashes in the router name (podman, gateways, kube sites with a dash in the namespace), and it could be that the router's ROUTER record may not always use `<prefix>:0` as its id.

lynnemorrison: @c-kruse is there a better way to get the router id from a link name?  I looked through the code to see if there were any macos/functions but couldn't find any.

c-kruse: @lynnemorrison I'm not positive there's an excellent example going in this direction. That `graph` function and the linkResponseHandler get there in slightly different ways. My understanding is that you're pretty close, but maybe missing a condition or two. Links can get pretty exciting with corner cases.

I'd recommend looking at using normalizeRouterName - with that you should be able to consistently get the peer router name of a link. That's what I've been promised is the most concrete way to complete the link-to-peer-router relation.

So starting with a subject incoming link, I think you'd want to find that subject link's router and note the router name. Then iterate peer links: Links where the direction is outgoing and where the link's normalized name matches our subject link's router name. Then finally check that that peer link's router's name matches the subject link's normalized name.

pkg/flow/flow_mem_driver.go
```diff
@@ -809,6 +809,25 @@ func (fc *FlowCollector) updateRecord(record interface{}) error {
 				}
 			} else {
 				if link.EndTime > 0 {
+					if current.Direction != nil && *current.Direction == Incoming {
```
ajssmith, line 812: iiuc, the removal of the peer link should be regardless of direction. The selection of the peer link should then be for the "opposite" direction.

ajssmith (approved):
Locally tried k8s sites both directions and podman to k8s site

thnx!

c-kruse (approved):
Appreciate you bearing with me @lynnemorrison! Looks good to me.

Issue #1701, opened 2024-10-02
[v2] After creating/deleting sites east/west, see many logs reporting: [ControllerError] Error while handling Certificate


Karen-Schoener:
**Describe the bug**
After creating/deleting sites east/west, see many logs reporting: 
`[ControllerError] Error while handling Certificate west/skupper-site-ca: secrets \"skupper-site-ca\" not found
`

Issue is consistently reproducible.

**How To Reproduce**
```
    kubectl apply  -f ./cmd/controller/deploy_cluster_scope.yaml

    kubectl apply  -n west -f cmd/controller/example/site1.yaml
    Wait 10 seconds.

    kubectl apply  -n east -f cmd/controller/example/site2.yaml
    Wait 10 seconds.

    kubectl delete -n west -f cmd/controller/example/site1.yaml
    At this point, skupper-controller issues many logs: [ControllerError] Error while handling Certificate... 
    
```

Snippet of skupper-controller logs:
```
time=2024-10-02T21:13:17.630Z level=INFO msg="Checking Certificate west/skupper-local-server"
time=2024-10-02T21:13:17.824Z level=INFO msg="[ControllerError] Error while handling Certificate west/skupper-local-server: secrets \"skupper-local-server\" not found"
time=2024-10-02T21:13:17.829Z level=INFO msg="Checking Certificate west/skupper-local-server"
time=2024-10-02T21:13:18.027Z level=INFO msg="[ControllerError] Error while handling Certificate west/skupper-local-server: secrets \"skupper-local-server\" not found"
time=2024-10-02T21:13:18.032Z level=INFO msg="Checking Certificate west/skupper-local-server"
time=2024-10-02T21:13:18.226Z level=INFO msg="[ControllerError] Error while handling Certificate west/skupper-local-server: secrets \"skupper-local-server\" not found"
time=2024-10-02T21:13:18.232Z level=INFO msg="Checking Certificate west/skupper-local-server"
time=2024-10-02T21:13:18.427Z level=INFO msg="[ControllerError] Error while handling Certificate west/skupper-local-server: secrets \"skupper-local-server\" not found"
time=2024-10-02T21:13:18.433Z level=INFO msg="Checking Certificate west/skupper-local-server"
time=2024-10-02T21:13:18.624Z level=INFO msg="[ControllerError] Error while handling Certificate west/skupper-local-server: secrets \"skupper-local-server\" not found"
```

**Expected behavior**
Quiet logs after deleting sites east/west.

[...]

PR #1703, opened 2024-10-07
Fix handling of secrets in certificate manager

### LINKED ISSUE #1673: [v2] controller does not recover certificate state after restart
**Describe the bug**
After restarting the controller certificates are broken. It looks like this may just be a matter of a missing label we expect to be on a secret - associated with this TODO comment: https://github.com/skupperproject/skupper/blob/7d9c4c4fe9ac93e2b8c92502e5b1311ae9a6f67d/pkg/kube/certificates/mgr.go#L232

Certificates after restarting the controller.
```
skupper     skupper-grant-server      skupper-grant-server-ca   true                        CA "skupper/skupper-grant-server-ca" not found
skupper     skupper-grant-server-ca                                               true      secrets "skupper-grant-server-ca" already exists
```

grs:
Fixes #1673 and should I think help with #1701

### DIFF
--- a/pkg/kube/certificates/mgr.go
+++ b/pkg/kube/certificates/mgr.go
@@ -2,6 +2,7 @@ package certificates
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"reflect"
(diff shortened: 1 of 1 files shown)


c-kruse (approved):
Thank you @grs!

PR #1704, opened 2024-10-07
Cherry pick #1462, #1467 into 1.5

hash-d:


hash-d:
The check was successful when forcing Minikube to 1.33.1: https://app.circleci.com/pipelines/github/skupperproject/skupper/3987/workflows/ac87a85e-f2da-4bc8-a5d9-321119d37739

I have reverted that change, though.

PR #1713, opened 2024-10-11
V2 nonkube connector CLI

lynnemorrison:
Add Connector CLI commands for nonkube implementation
includes create, update status and delete.

### DIFF
--- a/internal/cmd/skupper/connector/nonkube/connector_create.go
+++ b/internal/cmd/skupper/connector/nonkube/connector_create.go
@@ -2,12 +2,13 @@ package nonkube
 
 import (
 	"fmt"
+	"net"
 	"strconv"
 
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common/utils"
 	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
-	"github.com/skupperproject/skupper/pkg/apis/skupper/v1alpha1"
+	"github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 	"github.com/skupperproject/skupper/pkg/utils/validator"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -54,6 +55,7 @@ func (cmd *CmdConnectorCreate) ValidateInput(args []string) []error {
 	numberValidator := validator.NewNumberValidator()
 	connectorTypeValidator := validator.NewOptionValidator(common.ConnectorTypes)
 	outputTypeValidator := validator.NewOptionValidator(common.OutputTypes)
+	hostStringValidator := validator.NewHostStringValidator()
 
 	// Validate arguments name and port
 	if len(args) < 2 {
@@ -101,7 +103,13 @@ func (cmd *CmdConnectorCreate) ValidateInput(args []string) []error {
 		}
 	}
 	if cmd.Flags.Host != "" {
-		// TBD what is valid host
+		ip := net.ParseIP(cmd.Flags.Host)
+		ok, _ := hostStringValidator.Evaluate(cmd.Flags.Host)
+		if !ok && ip == nil {
+			validationErrors = append(validationErrors, fmt.Errorf("host is not valid: a valid IP address or hostname is expected"))
+		}
+	} else {
+		validationErrors = append(validationErrors, fmt.Errorf("host name must be configured: an IP address or hostname is expected"))
 	}
 	if cmd.Flags.TlsSecret != "" {
 		// TBD what is valid TlsSecret
@@ -129,16 +137,16 @@ func (cmd *CmdConnectorCreate) InputToOptions() {
 }
 
 func (cmd *CmdConnectorCreate) Run() error {
-	connectorResource := v1alpha1.Connector{
+	connectorResource := v2alpha1.Connector{
 		TypeMeta: metav1.TypeMeta{
-			APIVersion: "skupper.io/v1alpha1",
+			APIVersion: "skupper.io/v2alpha1",
 			Kind:       "Connector",
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cmd.connectorName,
 			Namespace: cmd.namespace,
 		},
-		Spec: v1alpha1.ConnectorSpec{
+		Spec: v2alpha1.ConnectorSpec{
 			Host:           cmd.host,
 			Port:           cmd.port,
 			RoutingKey:     cmd.routingKey,
--- a/internal/cmd/skupper/connector/nonkube/connector_create.go
+++ b/internal/cmd/skupper/connector/nonkube/connector_create.go
@@ -2,15 +2,29 @@ package nonkube
 
 import (
 	"fmt"
+	"strconv"
+
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common/utils"
+	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
+	"github.com/skupperproject/skupper/pkg/apis/skupper/v1alpha1"
+	"github.com/skupperproject/skupper/pkg/utils/validator"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type CmdConnectorCreate struct {
-	CobraCmd  *cobra.Command
-	Flags     *common.CommandConnectorCreateFlags
-	Namespace string
-	siteName  string
+	connectorHandler *fs.ConnectorHandler
+	CobraCmd         *cobra.Command
+	Flags            *common.CommandConnectorCreateFlags
+	namespace        string
+	connectorName    string
+	port             int
+	output           string
+	host             string
+	routingKey       string
+	connectorType    string
+	tlsSecret        string
 }
 
 func NewCmdConnectorCreate() *CmdConnectorCreate {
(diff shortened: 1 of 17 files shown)


lynnemorrison:
@nluaces For unit tests I created a dummy test file and then deleted it after test.   Want to verify this is acceptable.
Also need to evaluate names for yaml encoding functions in custome_resource_handler.go

internal/nonkube/client/fs/custom_resource_handler.go
```diff
@@ -27,6 +34,15 @@ func (b *BaseCustomResourceHandler) EncodeToYaml(resource interface{}) (string,
 	return string(result), err
 }
 
+// TBD better name
+func (b *BaseCustomResourceHandler) EncodeYaml(content []byte, resource interface{}) error {
```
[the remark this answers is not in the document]
lynnemorrison: done

internal/nonkube/client/fs/custom_resource_handler.go
```diff
@@ -2,20 +2,27 @@ package fs
...
+
 	"sigs.k8s.io/yaml"
 )
 
 type CustomResourceHandler[T any] interface {
 	Add(T) error
 	Update(T) error
-	Get(name string) T
+	Get(name string) (T, error)
+	GetRuntime(name string) (T, []fs.DirEntry, error)
```
nluaces, line 16: The idea of the CustomResourceHandler interface was to have a CRUD methods, similar to the kubernetes client, like: `Add`, `Update`, `Get` and `Delete`. If we need a method that gets more than one element we should create `List`. I think that we should have a `Get` method that retrieves a resource by its name, and a `List` methods that retrieves all the resources of the same type.
Also, this `List` method should return a list of Connectors, instead of DirEntry, the implementation of the method should take care of the transformation from file to connectors.

nluaces: The `List` method should replace `GetRuntime`

nluaces:
> timeout flag is specified in refdog for nonkube but not sure how that is used.
The timeout flag should be not used for now, giving that there is no controller for non kube environments. I suggest to hide it for the non_kube implementations.

Regarding the custom resource handler interface, remember to write implementations (empty if it is necessary) to all the resources that implement the interface (i.e. if you implement the List method for connectors, you also need to create the List method in sites)

internal/nonkube/client/fs/custom_resource_handler.go
lynnemorrison: done

internal/cmd/skupper/connector/nonkube/connector_update.go
```diff
@@ -2,28 +2,168 @@ package nonkube
...
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("connector name is not valid: %s", err))
+		} else {
+			cmd.connectorName = args[0]
+		}
+	}
+
+	// Validate that there is already a connector with this name in the namespace
+	if cmd.connectorName != "" {
+		connector, err := cmd.connectorHandler.Get(cmd.connectorName)
```
fgiorgetti, line 77: I believe that the resource handlers should return resources from the `input/sources` directory,
when the `runtime/state` directory does not exist for a given namespace.

The reason is because, returning from `runtime/state` only works if a namespace has already been
bootstrapped. If the user is using the CLI to prepare the site state, but haven't yet initialized it,
this will cause an error, while the resource is already placed as an input source.

Here is an example:

```
$ ./skupper connector create myservice 8080
File written to /home/user/.local/share/skupper/namespaces/default/input/sources/connectors/myservice.yaml
$ ./skupper connector status
No connectors found:
failed to read directory: open /home/user/.local/share/skupper/namespaces/default/runtime/state/connectors: no such file or directory
```

The absence of the `runtime` directory means that the given namespace was not yet bootstrapped.

Maybe if a namespace hasn't been initialized we should read from `input/sources` and warn about the
site not yet initialized? 

In case the namespace is initialized, there will be two source directories: `sources` and `input/sources`.
The first represents the current sources loaded during bootstrap, while the latter represents the desired
state, provided by the user (usually using the CLI).

If they differ, it means that user has made changes to the `input/sources` and there are pending changes to
be processed by a forced bootstrap execution.

internal/nonkube/client/fs/path_provider.go
```diff
@@ -17,3 +17,15 @@ func (p *PathProvider) GetNamespace() string {
 	}
 	return fmt.Sprintf(".local/share/skupper/namespaces/%s/input/sources", p.Namespace)
 }
+
+func (p *PathProvider) getRuntimeDefaultNamespace() string {
+	return ".local/share/skupper/namespaces/default/runtime/state"
```
fgiorgetti, line 22: One general comment to the `PathProvider` would be to use the `api.GetHostNamespaceHome()` function
from `pkg/nonkube/api/environment.go`. This function already considers the current user id as well as it
evaluates the XDG env vars to return the correct path.

internal/cmd/skupper/connector/nonkube/connector_update.go
lynnemorrison: done

internal/cmd/skupper/connector/nonkube/connector_status_test.go
```diff
@@ -1 +1,290 @@
 package nonkube
+
+import (
+	"os"
+	"testing"
+
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common/utils"
+	fs2 "github.com/skupperproject/skupper/internal/nonkube/client/fs"
```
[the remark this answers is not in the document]
lynnemorrison: fixed

internal/cmd/skupper/connector/nonkube/connector_create_test.go
```diff
@@ -1 +1,281 @@
 package nonkube
+
+import (
+	"testing"
+
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common/utils"
+	fs2 "github.com/skupperproject/skupper/internal/nonkube/client/fs"
```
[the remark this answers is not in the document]
lynnemorrison: fixed

internal/cmd/skupper/connector/nonkube/connector_create.go
```diff
@@ -2,28 +2,162 @@ package nonkube
...
+		}
+	}
+	if cmd.Flags.Output != "" {
+		ok, err := outputTypeValidator.Evaluate(cmd.Flags.Output)
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("output type is not valid: %s", err))
+		}
+	}
+	if cmd.Flags.Host != "" {
+		// TBD what is valid host
```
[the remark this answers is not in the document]
lynnemorrison: done

internal/cmd/skupper/connector/nonkube/connector_update.go
nluaces: I agree with the warning of  "Site not initialized yet" and to search in the input directory as well, but can we move the warning to the implementation of the Get command and just return a value or an error? I think that logic is part of how Get search for resources.

lynnemorrison: done

internal/nonkube/client/fs/connector_handler.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (s *ConnectorHandler) Add(resource v1alpha1.Connector) error {
+
+	fileName := resource.Name + ".yaml"
+	content, err := s.EncodeToYaml(resource)
+	if err != nil {
+		return err
+	}
+
+	err = s.WriteFile(s.pathProvider.GetNamespace(), fileName, content, "connectors")
```
fgiorgetti, line 32: A minor request to set "connectors" as a constant.

internal/cmd/skupper/connector/nonkube/connector_create.go
```diff
@@ -2,28 +2,168 @@ package nonkube
...
+			validationErrors = append(validationErrors, fmt.Errorf("connector type is not valid: %s", err))
+		}
+	}
+	if cmd.Flags.Output != "" {
+		ok, err := outputTypeValidator.Evaluate(cmd.Flags.Output)
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("output type is not valid: %s", err))
+		}
+	}
+	if cmd.Flags.Host != "" {
```
fgiorgetti, line 105: I believe that for nonkube sites, the host field must be required.

internal/nonkube/client/fs/path_provider.go
lynnemorrison: done

internal/cmd/skupper/connector/nonkube/connector_create.go
lynnemorrison: @fgiorgetti do we want to default the host name to be the connector name?   Or do we just fail if host is not defined?

skupper connector backend -n west 
the host name would default to "backend"if not set.
also should we do the same for routing key?  

The default for kube is if routing key or selector is not set it defaults to connector name.

fgiorgetti: @lynnemorrison I believe it should just be required, but it is hard to guess what would be the default hostname.
The connector name wouldn't work, as it probably won't be a name that can be resolved by the router container.

For the routingKey, I believe you can use the default behavior we currently have on kube.

lynnemorrison: done

internal/cmd/skupper/connector/nonkube/connector_update.go
nluaces: I think that the List() is not yet included in the Router Access Handler

lynnemorrison: updated,  thanks missed that.

nluaces (approved):

internal/cmd/skupper/connector/nonkube/connector_update.go
```diff
@@ -2,28 +2,175 @@ package nonkube
...
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("connector name is not valid: %s", err))
+		} else {
+			cmd.connectorName = args[0]
+		}
+	}
+
+	// Validate that there is already a connector with this name in the namespace
+	if cmd.connectorName != "" {
+		connector, err := cmd.connectorHandler.Update(cmd.connectorName)
```
fgiorgetti, line 79: This method is called `Update`, but it is actually just retrieving a connector.
Should it be named something else?

internal/nonkube/client/fs/connector_handler.go
```diff
@@ -0,0 +1,120 @@
...
+		var context v2alpha1.Connector
+		if err = s.DecodeYaml(connector, &context); err != nil {
+			return nil, err
+		}
+		connectors = append(connectors, &context)
+	}
+	return connectors, nil
+}
+
+func (s *ConnectorHandler) Update(name string) (*v2alpha1.Connector, error) {
```
fgiorgetti, line 105: Update does not seem like an appropriate name for what this method is doing.
Maybe we could tweak the `Get` signature allowing some options to be passed
that indicates if it must read from runtime first or from input only?

internal/nonkube/client/fs/path_provider.go
```diff
@@ -1,19 +1,17 @@
...
 type PathProvider struct {
 	Namespace string
 }
 
-func (p *PathProvider) getDefaultNamespace() string {
-	return ".local/share/skupper/namespaces/default/input/sources"
-}
-
 func (p *PathProvider) GetNamespace() string {
+	return api.GetHostNamespaceHome(p.Namespace) + "/input/sources"
```
fgiorgetti, line 12: suggest replace "/input/sources" with api.InputSiteStatePath and concatenate with path.Join()

internal/nonkube/client/fs/path_provider.go
```diff
@@ -1,19 +1,17 @@
...
+}
 
-	if p.Namespace == "" {
-		return p.getDefaultNamespace()
-	}
-	return fmt.Sprintf(".local/share/skupper/namespaces/%s/input/sources", p.Namespace)
+func (p *PathProvider) GetRuntimeNamespace() string {
+	return api.GetHostNamespaceHome(p.Namespace) + "/runtime/state"
```
fgiorgetti, line 16: suggest replace "/runtime/state" with api.RuntimeSiteStatePath and concatenate with path.Join()

fgiorgetti (approved):
Other than my comments and minor suggestions above, it is working just fine. 
Thank you!

internal/cmd/skupper/connector/nonkube/connector_update.go
nluaces: I missed this, I agree that Update should update a resource.

internal/nonkube/client/fs/connector_handler.go
nluaces: Following the go client for kubernetes as example, the get command interface could be modified to something like this:
```
Get(name string, opts GetOptions) T

type GetOptions struct {
// runtimeFirst bool
// ...
}
```

lynnemorrison: Will add a new issue to modify the Get command.

internal/nonkube/client/fs/path_provider.go
lynnemorrison: sorry thought I caught all of these.   Fixed

internal/nonkube/client/fs/path_provider.go
lynnemorrison: fixed

internal/cmd/skupper/connector/nonkube/connector_update.go
lynnemorrison: will add new issue to fix

internal/nonkube/client/fs/connector_handler.go
lynnemorrison: https://github.com/skupperproject/skupper/issues/1777

PR #1714, opened 2024-10-11
fixes deadlock found in the network console collector

c-kruse:
exposes the flow pair reconcile process to the collector_job_seconds internal metric to aid in debugging related issues in the future.


bartoval (approved):
💯

PR #1721, opened 2024-10-14
bump k8s

ajssmith:


nluaces (approved):

c-kruse (changes requested):
Looks like the new leader election strategy would require additional privileges. config-sync gets stuck on this:

```
1 leaderelection.go:436] error retrieving resource lock public-lb/skupper-site-leader: leases.coordination.k8s.io "skupper-site-leader" is forbidden: User "system:serviceaccount:public-lb:skupper-router" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "public-lb"
```

c-kruse (commented):
@ajssmith it looks like the controller also needs that rule added. After editing the clustrrole this worked!

With the controller deployments as-is I get a site status of `roles.rbac.authorization.k8s.io "skupper-router" is forbidden: user "system:serviceaccount:skupper:skupper-controller" (groups=["system:serviceaccounts" "system:serviceaccounts:skupper" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:...`

ajssmith:
hmmm, I had added to the role and then convinced myself it was not needed...let me re-check

fgiorgetti (approved):
Worked fine on a recent minikube and on Openshift 4.12.

grs (approved):

PR #1722, opened 2024-10-14
update desc in site crd

pwright:


api/types/crds/skupper_site_crd.yaml
```diff
@@ -81,57 +113,91 @@ spec:
...
+                        description: |
+                          Name of the endpoint.
                       group:
                         type: string
+                        description: |
+                          Group or category to which the endpoint belongs.
                 sitesInNetwork:
                   type: integer
+                  description: |
+                    The number of sites currently active in the network.
```
pwright, line 133: currently active like in v1?

api/types/crds/skupper_site_crd.yaml
```diff
@@ -81,57 +113,91 @@ spec:
...
+                        description: |
+                          The name of the site in the network.
                       namespace:
                         type: string
+                        description: |
+                          Kubernetes namespace where the site is deployed.
                       platform:
                         type: string
+                        description: |
+                          The platform on which the site is running (e.g. Kubernetes, Docker).
```
pwright, line 154: k8s, docker, podman, systemd, do we detect OpenShift?

api/types/crds/skupper_site_crd.yaml
```diff
@@ -81,57 +113,91 @@ spec:
...
                       version:
                         type: string
+                        description: |
+                          The version of Skupper running on the site.
```
pwright, line 158: The version of site software?

api/types/crds/skupper_site_crd.yaml
```diff
@@ -17,27 +17,47 @@ spec:
               properties:
                 routerMode:
                   type: string
+                  description: |
+                    Defines the mode of the router for the site.
+                    - **interior**: Full linking capability across the network.
+                    - **edge**: Limited to linking to an interior site, acts as a leaf node.
```
pwright, line 23: add default

api/types/crds/skupper_site_crd.yaml
pwright: check case

api/types/crds/skupper_site_crd.yaml
```diff
@@ -17,27 +17,47 @@ spec:
...
                 serviceAccount:
                   type: string
+                  description: |
+                    The site uses the specified service account to interact with Kubernetes resources.
```
grs, line 27: Maybe 'If specified' at the start, and then state that if not specified a serviceaccount will be created by the controller.

api/types/crds/skupper_site_crd.yaml
```diff
@@ -17,27 +17,47 @@ spec:
...
+                  description: |
+                    Displays the default certificate issuer currently configured for the site.
                 active:
                   type: boolean
+                  description: |
+                    Indicates whether the site is currently active and operational.
                 status:
                   type: string
+                  description: |
+                    The current operational status of the site (e.g. Active, Failed, Starting).
```
grs, line 60: At present this either has 'OK@ or some error message.

api/types/crds/skupper_site_crd.yaml
grs: I would say simply 'currently in the network' as the meaning of active here is not clear.

pwright:
closing as these changes need to land against refdog

PR #1723, opened 2024-10-14
Fix SecuredAccessManager updatePorts

c-kruse:
When comparing desired and actual state of services, ignore the value of NodePort as this is assigned dynamically by an external controller and is not managed by us.


c-kruse (commented):
@grs  does this approach make sense to you?

It appears to work, but I'm skeptical it is really that straight-forward. Is it okay to be meddling with resource state like this? If I understand correctly there's also informers/watchers that are meant to be managing these resources, so editing them feels a bit dangerous.

After a closer look this test is not doing what I had hoped it would be doing, so I plan to circle back to that if this looks like an okay path forward.

grs:
@c-kruse I think https://github.com/skupperproject/skupper/pull/1730 is a better way to fix the issue

PR #1726, opened 2024-10-16
refine peer link deletion

ajssmith:


pkg/flow/flow_mem_driver.go
```diff
@@ -823,11 +823,15 @@ func (fc *FlowCollector) updateRecord(record interface{}) error {
...
 						}
 
-						// find any links to the peer router and delete
+						// find any links to the peer router and delete the one that is for current
 						for _, eLink := range fc.Links {
 							if eLink.Direction != nil && *eLink.Direction == peerDirection {
 								if routerId == eLink.Parent {
-									fc.deleteRecord(eLink)
+									parts := strings.Split(*eLink.Name, "-")
+									if parts[len(parts)-1]+":0" == current.Parent {
```
c-kruse, line 831: I do not believe this pattern holds for podman sites. The router will choose a random vanflow id prefix for itself if its configured name isn't pod name-like, so I don't think that name suffix matching logic would work.

I'd suggest the method Lynne took above:
Hold on to something like `currentRouterName := normalizeRouterName(*fc.Routers[current.Parent].Name)` then just compare that to eLink.Name. (also maybe nil-check the pointer dereferences - especially if connections are coming and going we may not have parents and/or names)

pkg/flow/flow_mem_driver.go
```diff
@@ -810,25 +810,22 @@ func (fc *FlowCollector) updateRecord(record interface{}) error {
...
 							}
 						}
-
-						// find any links to the peer router and delete
 						for _, eLink := range fc.Links {
-							if eLink.Direction != nil && *eLink.Direction == peerDirection {
-								if routerId == eLink.Parent {
-									fc.deleteRecord(eLink)
-								}
+							if eLink.Direction != nil && *eLink.Direction == peerDirection && peerRouterId == eLink.Parent && *eLink.Name == normalizeRouterName(*fc.Routers[current.Parent].Name) {
```
c-kruse, line 826: Logic looks solid to me!

Still a bit afraid of a panic if eLink.Name is nil, if fc.Routers[current.Parent] is not present (nil) or if fc.Routers[current.Parent].Name is nil. All three of those should be easy to contrive in the test we have (even if we don't actually commit those test cases.) I am not so sure how likely it'd be to occur in even the most troubled real networks, but I'd rather not be surprised.

c-kruse (approved):
Thank you @ajssmith!

lynnemorrison (approved):

PR #1728, opened 2024-10-16
Run go tests with race detection enabled

c-kruse:
Updates makefile and CI to use the go test `-race` flag, fixes a race in the fs watcher test harness, and skips compat tests that depend on an external container engine.


fgiorgetti (approved):
LGTM!

grs (approved):

nluaces (approved):

Issue #1720, opened 2024-10-14
[v2] SecuredAccessManager - securedaccess endpoints missing port


c-kruse:
**Describe the bug**
With a secured access of type nodePort the endpoints sometimes end up with port set to zero. I think this is a simple case of the manager squashing the nodePort field during the reconcile step - maybe even a slight data race.

```
[christian@rh skupper-ci-playground] (fasterci *)$ k get securedaccesses.skupper.io skupper-router-1 -oyaml | yq '{"spec": .spec, "statusEndpoints": .status.endpoints}'
spec:
  accessType: nodeport
  certificate: skupper-site-server
  issuer: skupper-site-ca
  ports:
    - name: inter-router
      port: 55671
      protocol: TCP
      targetPort: 55671
    - name: edge
      port: 45671
      protocol: TCP
      targetPort: 45671
  selector:
    skupper.io/component: router
    skupper.io/group: skupper-router
statusEndpoints:
  - host: host.skupper-dev.testing
    name: inter-router
    port: "0"
  - host: host.skupper-dev.testing
    name: edge
    port: "0"
[christian@rh skupper-ci-playground] (fasterci *)$ ^C
[christian@rh skupper-ci-playground] (fasterci *)$ k get svc skupper-router-1
NAME               TYPE       CLUSTER-IP     EXTERNAL-IP   PORT(S)                           AGE
skupper-router-1   NodePort   10.96.39.145   <none>        55671:31216/TCP,45671:31147/TCP   7m19s
```
**How To Reproduce**

enable nodeport access types and create a site with accessType: nodeport. Wait around a few minutes and check once in a while.

PR #1730, opened 2024-10-17
improve handling of nodeport service

grs:
* only return nodeport endpoints when port has been assigned
* fix reconciliation of ports for nodeport services

### DIFF
--- a/pkg/kube/securedaccess/access.go
+++ b/pkg/kube/securedaccess/access.go
@@ -424,11 +424,13 @@ func updatePorts(spec *corev1.ServiceSpec, desired []skupperv1alpha1.SecuredAcce
 	var ports []corev1.ServicePort
 	for _, actual := range spec.Ports {
 		if port, ok := expected[actual.Name]; ok {
-			ports = append(ports, port)
-			delete(expected, actual.Name)
-			if actual != port {
+			if equivalentPorts(port, actual) {
+				ports = append(ports, actual)
+			} else {
+				ports = append(ports, port)
 				changed = true
 			}
+			delete(expected, actual.Name)
 		} else {
 			changed = true
 		}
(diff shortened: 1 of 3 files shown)


c-kruse (approved):
@grs This looks to have fixed the behavior!

I am still a little concerned about what I was seeing and don't totally follow exactly how it is working - I know that is is not totally obvious. I suspect endpoints could be returned from "RealiseAndResolve" maybe once before the manager's copy of the service gets "clobbered" by that checkService step of reconcile - that may work okay since in practice NodePorts are not going to change.

grs:
@c-kruse I was a little tunnel minded on my return yesterday after some days away from this, and was focused purely on the issue #1720 where the endpoints in the SecuredAccess status (and from there copied into site status) have a 0 port. This change is really just dealing with that. If the nodePort is zero for a nodeport service then there is no valid endpoint to report.

A separate potential issue is the one that your patch addresses, which is whether the current reconciliation of a nodeport service results in thrashing/conflict as it does not consider the system assigned nodePort field. Do you actually see such conflict? So far I have not but will investigate a little more on that. We may end up wanting both changes.

grs:
@c-kruse I have added an alternative fix to the issue your PR addresses, namely fixing the reconciliation of the service ports when the service is of type nodeport.

c-kruse:
@grs thank you! Much more confidence inspiring.

I am able to reliably reproduce that conflict, although it is difficult to observe without some printf debugging - it does not manifest itself quite like I would expect with "thrashing". Maybe the kube api server (kind v1.27) I am using kindly disregards the update zeroing out NodePort but the client returns the caller's value?

In case you're curious and haven't observed for yourself:
```
diff --git a/pkg/kube/securedaccess/access.go b/pkg/kube/securedaccess/access.go
index 19dc7989..5e48f585 100644
--- a/pkg/kube/securedaccess/access.go
+++ b/pkg/kube/securedaccess/access.go
@@ -2,6 +2,7 @@ package securedaccess

 import (
        "context"
+       "encoding/json"
        "errors"
        "fmt"
        "log"
@@ -154,6 +155,16 @@ func (m *SecuredAccessManager) reconcile(sa *skupperv1alpha1.SecuredAccess) erro
                return nil
        }
        updated := false
+       defer func() {
+               pb, _ := json.Marshal(svc.Spec.Ports)
+               fmt.Printf(
+                       "ck/debug: SecuredAccess.reconcile type: %q updated %v svc/%s --namespace %s :: %s\n",
+                       sa.Spec.AccessType,
+                       updated,
+                       svc.Name,
+                       svc.Namespace,
+                       string(pb))
+       }()
        endpoints, resourceErr := m.accessType(sa).RealiseAndResolve(sa, svc)
```
Anyways, this looks correct to me now.

c-kruse (approved):

PR #1735, opened 2024-10-18
fix cleanup of podwatchers

### LINKED ISSUE #1727: [v2] Instability in the maintenance of router configuration
**Describe the bug**
In a network consisting of 4 interior kubernetes sites and an interior podman site, I'm seeing service-connectivity problems and instability in the maintenance of the router configuration, specifically the tcpListener and tcpConnector entities.

When a connector target is deployed, the tcpConnector is created and destroyed repeatedly, dozens of times, in the router local to the target workload.  A tcpListener used by the target workload also thrashes similarly.

It it interesting to note that the `siteId` attribute in the tcpConnector and tcpListener entities in the router toggle between two different values.  One of the values is correct (i.e. the UID of the kubernetes Site CR), and the other is a different UUID that I cannot associate with any other object.

**How To Reproduce**
It is unclear how to portably reproduce this issue.  I am happy to deploy a network that shows the problem for anyone to look at.

**Expected behavior**
I expect the router configuration to change in an orderly, once-when-needed fashion.

**Environment details**
 - Skupper CLI: Not Used
 - Skupper Operator (if applicable): Head of V2 branch
 - Platform: Openshift and Minikube

**Additional context**
None.

grs:
Fixes #1727

### DIFF
--- a/pkg/kube/site/attached_connector.go
+++ b/pkg/kube/site/attached_connector.go
@@ -174,11 +174,17 @@ func (a *AttachedConnector) IncludeNotReady() bool {
 	return false
 }
 
-func (a *AttachedConnector) Id() string {
+func (a *AttachedConnector) Attr() slog.Attr {
 	if definition := a.activeDefinition(); definition != nil {
-		return fmt.Sprintf("AttachedConnector %s/%s", definition.Name, definition.Namespace)
+		return slog.Group("AttachedConnector",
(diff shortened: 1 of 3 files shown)


ted-ross (approved):
I re-ran my test with this fix and all of the bad behaviors are gone.  Merge it!

fgiorgetti (approved):

PR #1749, opened 2024-10-25
Updated rootful namespace home to var lib

### LINKED ISSUE #1689: [v2] location of SKUPPER_HOME is unconventional
**Describe the bug**

`SKUPPER_HOME`, which houses non-static configuration and other non-static content, is rooted on `/usr/local/share/skupper/namespaces/` for the root user.

That location may  not be adequate given the characteristics of the files that reside there.

According to `hier(7)`:

> `/usr   This directory is usually mounted from a separate partition.  It should hold only shareable, read-only data, so that it can be mounted by various machines running Linux.`
> (...)
> `       /usr/share`
> `             This directory contains subdirectories with specific application data, that can be shared among different architectures of the same OS.  Often one finds stuff here that used to live in /usr/doc or /usr/lib or /usr/man.`
> (...)
>  `      /usr/local/share`
> `              Local application data that can be shared among different architectures of the same OS.`

The main point here is the data not being read-only.  Also, the examples of data that is expected to reside there.

**How To Reproduce**

N/A

**Expected behavior**

`/etc` might be a good candidate to host the files, but as some of them are created and updated by the running application, perhaps `/var/lib` may be a better location for them.  Again `hier(7)`:

> `       /var/lib`
> `             Variable state information for programs.`

Examples of similar uses of that are `/var/lib/containers/storage/` and `/var/lib/libvirt/dnsmasq`

**Environment details**

N/A

**Additional context**

See discussion on refdog for start of conversation: https://github.com/skupperproject/refdog/issues/3

fgiorgetti:
* Only rootful namespace home has been updated to /var/lib
* Rootless remain as is assuming that XDG_DATA_HOME seems to be appropriate
   (see: https://specifications.freedesktop.org/basedir-spec/latest/)

Fixes #1689

### DIFF
--- a/cmd/bootstrap/README.md
+++ b/cmd/bootstrap/README.md
@@ -65,7 +65,7 @@ files are stored, like certificates, configurations, the original sources
 (original custom resources used to bootstrap the nonkube site) and
 the runtime files generated during initialization.
 
-Namespaces are stored under ${XDG_DATA_HOME}/.local/share/skupper/namespaces
+Namespaces are stored under ${XDG_DATA_HOME}/skupper/namespaces
 for regular users when XDG_DATA_HOME environment variable is set, or under
 ${HOME}/.local/share/skupper/namespaces when it is not set.
 As the root user, namespaces are stored under: /var/lib/skupper/namespaces.
--- a/cmd/bootstrap/README.md
+++ b/cmd/bootstrap/README.md
@@ -68,7 +68,7 @@ the runtime files generated during initialization.
 Namespaces are stored under ${XDG_DATA_HOME}/.local/share/skupper/namespaces
 for regular users when XDG_DATA_HOME environment variable is set, or under
 ${HOME}/.local/share/skupper/namespaces when it is not set.
-As the root user, namespaces are stored under: /usr/local/share/skupper/namespaces.
+As the root user, namespaces are stored under: /var/lib/skupper/namespaces.
 
 In case the path (-p) flag is omitted, Skupper will try to process
 custom resources stored at the sources directory of the default namespace,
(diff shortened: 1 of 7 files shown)


cmd/bootstrap/README.md
```diff
@@ -68,7 +68,7 @@ the runtime files generated during initialization.
 Namespaces are stored under ${XDG_DATA_HOME}/.local/share/skupper/namespaces
```
hash-d, line 68: I think this should be `${XDG_DATA_HOME}/skupper/namespaces`, instead (ie, without `.local/share`, which is part of the default `${XDG_DATA_HOME}`).

https://specifications.freedesktop.org/basedir-spec/latest/

> $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored. If  $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

cmd/bootstrap/bootstrap.go
```diff
@@ -46,7 +46,7 @@ the runtime files generated during initialization.
 Namespaces are stored under ${XDG_DATA_HOME}/.local/share/skupper/namespaces
```
hash-d, line 46: Same here:` ${XDG_DATA_HOME}/skupper/namespaces`, instead.

hash-d (approved):
The changes themselves look good for me, but I think there is a couple of pre-existing documentation mistakes that could be looked into.

fgiorgetti:
> @fgiorgetti the Path Provider methods should be updated as well or they should be replaced?

@nluaces In PR #1713 , @lynnemorrison already modified the Path Provider to use `api.GetHostNamespaceHome`,
so I believe it should be ok, as soon as it lands.

nluaces (approved):

c-kruse (approved):

PR #1750, opened 2024-10-25
rollback changes to network console api version

c-kruse:
The network console api version is v1alpha1 and should not be conflated with the skupper.io CRD resource versions.


c-kruse:
@grs  we missed this in the v2 -> v2alpha1 migration.

grs (approved):
Sorry! I thought I had pushed those changes to my PR.

PR #1756, opened 2024-10-29
Fixes #1717 - Change the syntax of the 'oneOf' clause

### LINKED ISSUE #1717: [v2] Anomolous CRD-application behavior
**Describe the bug**
When applying CRDs to a cluster, the "connector" CRD always reports "configured" even when it is unchanged.  Other CRDs do not do this.

**How To Reproduce**
Apply the CRD from the v2 branch:
```
kubectl apply -f https://raw.githubusercontent.com/skupperproject/skupper/v2/api/types/crds/skupper_connector_crd.yaml
```

**Expected behavior**
After the first application, I expect the result to be "unchanged"

**Environment details**
 - Skupper CLI: None used
 - Skupper Operator (if applicable): head of the v2 branch
 - Platform: openshift

**Additional context**
None

ted-ross:
This change matches the syntax of the 'oneOf' clause (for validation) to the syntax that is returned when you use the `kubectl get crd connectors.skupper.io -o yaml` command.

Even though the two syntaxes are semantically equivalent (I believe), `kubectl apply -f` detects a mismatch and will patch the CRD with no effect.

### DIFF
--- a/api/types/crds/skupper_connector_crd.yaml
+++ b/api/types/crds/skupper_connector_crd.yaml
@@ -37,12 +37,10 @@ spec:
               - routingKey
               - port
               oneOf:
-                - properties:
-                  required:
-                  - selector
-                - properties:
-                  required:
-                  - host
+              - required:
+                - selector
+              - required:
+                - host
             status:
               type: object
               properties:


grs (approved):
:shipit:

fgiorgetti (approved):
LGTM

PR #1761, opened 2024-10-31
support for tls credentials in connectors and listeners

grs:
(plus some improved test coverage)


pkg/apis/skupper/v2alpha1/types.go
```diff
@@ -486,6 +486,8 @@ type ConnectorSpec struct {
 	Selector        string            `json:"selector,omitempty"`
 	Port            int               `json:"port"`
 	TlsCredentials  string            `json:"tlsCredentials,omitempty"`
+	NoClientAuth    bool              `json:"noClientAuth,omitempty"`
```
c-kruse, line 489: Question for my own education: how have you been doing this?

Are you writing go + generating the crd yaml defintions, vice versa, or is it just a manual process?

grs: I use a manual process, with the go and yaml edited separately. It may be possible to do some generation one way or the other, but I have not used that.

lynnemorrison:
Will these changes require changes to the kube connector/listener commands?

grs:
> Will these changes require changes to the kube connector/listener commands?

Yes, probably having options in the CLI eventually for these new fields will be useful. However I think the naming is still in flux so I would not rush to add anything yet.

pkg/nonkube/api/site_state.go
```diff
@@ -247,8 +247,8 @@ func (s *SiteState) linkMap(sslProfileBasePath string) site.LinkMap {
 	return linkMap
 }
 
-func (s *SiteState) bindings() *site.Bindings {
-	b := site.NewBindings()
+func (s *SiteState) bindings(sslProfileBasePath string) *site.Bindings {
+	b := site.NewBindings(path.Join(sslProfileBasePath, "certificates/bindings"))
```
fgiorgetti, line 251: If you set its path to `certificates/server` instead of `certificates/bindings`, it should work ok.
Even better if you use `api.CertificatesServerPath` instead of the literal string (from: `pkg/nonkube/api/environment.go`).

pkg/qdr/qdr.go
```diff
@@ -482,7 +493,7 @@ type TcpEndpoint struct {
 	Address        string `json:"address,omitempty"`
 	SiteId         string `json:"siteId,omitempty"`
 	SslProfile     string `json:"sslProfile,omitempty"`
-	VerifyHostname *bool  `json:"verifyHostname,omitempty"`
+	VerifyHostname *bool  `json:"verifyHostname"`
```
fgiorgetti, line 496: Since TcpEndpoint is used for both tcpListener and tcpConnector, I believe that `omitempty` flag must
be preserved as the tcpListener does not have the `verifyHostname` property defined.

pkg/nonkube/api/site_state.go
fgiorgetti: Actually, it would be better to apply the following patch to site_state.go:
https://gist.github.com/fgiorgetti/9eb26360bf26741e6e9893067d7b80fb

This way it will use client certificates for Connectors and server certificates for Listeners.
WDYT?

grs: > Actually, it would be better to apply the following patch to site_state.go: https://gist.github.com/fgiorgetti/9eb26360bf26741e6e9893067d7b80fb
> 
> This way it will use client certificates for Connectors and server certificates for Listeners. WDYT?

I expect that the credentials for connectors and listeners will be supplied by the user.

fgiorgetti: And I believe that your expected approach is correct.

So the patch above ensures that the existing mechanism defined through [PROVIDED_CERTIFICATES.md](https://github.com/skupperproject/skupper/blob/v2/cmd/bootstrap/PROVIDED_CERTIFICATES.md),
which allows users to provide their own certificates, is used properly for the respective Listeners and
Connectors, when the tlsCredentials field is populated.

Or do you have concerns with the current structure for user provided certificates?

pkg/qdr/qdr.go
fgiorgetti: I see. The problem I noticed is that when I create a Listener, the respective tcpListener rendered
into the router config, has the following value, which is invalid and seems to be causing problems to the router:

```
                "verifyHostname": null
```

pkg/nonkube/api/site_state.go
grs: From that doc

```Server and client certificates can be provided whenever your site definition contains at least one RouterAccess (resource).

The expected directory names for the server and client certificates, is determined based on the values of RouterAccess.spec.tlsCredentials (optional field), or RouterAccess.name (default).
```

I assumed those locations were for the inter-router certificates and that those for bindings would be supplied elsewhere.

The same secret could be used for both a connector and a listener. Having them under the same path makes that simpler.

The patch above seems to create a new bindings object for every connector/listener which I don't think will work. (Having everything under the same path avoids the need for that also).

pkg/qdr/qdr.go
```diff
@@ -824,9 +835,19 @@ func equivalentHost(a string, b string) bool {
 	}
 }
 
+func (a TcpEndpoint) equivalentVerifyHostname(b TcpEndpoint) bool {
+	if a.VerifyHostname == nil {
+		return b.VerifyHostname == nil
```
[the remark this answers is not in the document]
grs: Fixed

pkg/qdr/qdr.go
grs: Fixed

pkg/nonkube/api/site_state.go
fgiorgetti: Alright, let's keep it as is (using `bindings/`) and I can make further changes to load properly.
Today we also allow "secrets" to be provided, which is currently just used for `Links`. Maybe we
can also extend it to work with Listeners and Connectors as well, in an upcoming PR.

fgiorgetti (approved):

lynnemorrison (approved):

nluaces (approved):

PR #1780, opened 2024-11-08
Upgrade prometheus and oauth-proxy image versions

c-kruse:
Upgrade prometheus image from v2.42.0 to v2.55.1
Upgrade origin-auth-proxy image from 4.14.0 to 4.18.0


c-kruse:
Appears rather low-risk to me.

Vetted that upgraded and fresh sites work. The delta between oauth proxy releases is quite small. For prometheus the delta is larger, but the project is very mature and we use an incredibly small subset of its feature set.

nluaces (approved):

Issue #1705, opened 2024-10-07
[v2] Network Console Collector should purge inferred records when their parent is purged.


c-kruse:
I had missed a few cases that causes some confusing behavior when a site comes and goes:

* inferred processes (site-client/site-server) should be removed with their parent SiteRecord
* process|processgroup|sitepairs similarly should be removed along with the last process record (or just the process/group/site?)
* Addresses with listeners/connectors

PR #1785, opened 2024-11-14
Collector: Fix Collector Address Manager bugs

c-kruse:
Addresses are now deleted when associated listeners and connectors are gone.
Adds test coverage.

Part of: https://github.com/skupperproject/skupper/issues/1705


bartoval (approved):
+1

PR #1786, opened 2024-11-14
Add nonkube site commands and unit tests

lynnemorrison:
Add update, delete and status site CLI commands
Currently create router-access if user configures enable-link-access.
Add unit tests.

### DIFF
--- a/internal/cmd/skupper/site/nonkube/site_create.go
+++ b/internal/cmd/skupper/site/nonkube/site_create.go
@@ -5,6 +5,8 @@ package nonkube
 
 import (
 	"fmt"
+	"net"
+
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common/utils"
 	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
@@ -14,10 +16,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const (
-	SiteConfigNameKey string = "name"
-)
-
 type CmdSiteCreate struct {
 	siteHandler             *fs.SiteHandler
 	routerAccessHandler     *fs.RouterAccessHandler
@@ -25,7 +23,7 @@ type CmdSiteCreate struct {
 	Flags                   *common.CommandSiteCreateFlags
 	options                 map[string]string
 	siteName                string
-	linkAccessType          string
+	linkAccessEnabled       bool
 	output                  string
 	namespace               string
 	bindHost                string
@@ -48,12 +46,10 @@ func (cmd *CmdSiteCreate) NewClient(cobraCommand *cobra.Command, args []string) 
 }
 
 func (cmd *CmdSiteCreate) ValidateInput(args []string) []error {
-
 	var validationErrors []error
-
-	if cmd.Flags.ServiceAccount != "" {
-		fmt.Println("Warning: --service-account flag is not supported on this platform")
-	}
+	hostStringValidator := validator.NewHostStringValidator()
+	resourceStringValidator := validator.NewResourceStringValidator()
+	outputTypeValidator := validator.NewOptionValidator(common.OutputTypes)
 
 	if cmd.CobraCmd != nil && cmd.CobraCmd.Flag(common.FlagNameContext) != nil && cmd.CobraCmd.Flag(common.FlagNameContext).Value.String() != "" {
 		fmt.Println("Warning: --context flag is not supported on this platform")
@@ -62,10 +58,6 @@ func (cmd *CmdSiteCreate) ValidateInput(args []string) []error {
 	if cmd.CobraCmd != nil && cmd.CobraCmd.Flag(common.FlagNameKubeconfig) != nil && cmd.CobraCmd.Flag(common.FlagNameKubeconfig).Value.String() != "" {
 		fmt.Println("Warning: --kubeconfig flag is not supported on this platform")
 	}
-
-	resourceStringValidator := validator.NewResourceStringValidator()
-	linkAccessTypeValidator := validator.NewOptionValidator(common.LinkAccessTypes)
-	outputTypeValidator := validator.NewOptionValidator(common.OutputTypes)
 
 	if len(args) == 0 || args[0] == "" {
 		validationErrors = append(validationErrors, fmt.Errorf("site name must not be empty"))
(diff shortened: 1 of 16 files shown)


internal/cmd/skupper/site/nonkube/site_delete.go
```diff
@@ -2,29 +2,82 @@ package nonkube
...
+	} else {
+		ok, err := resourceStringValidator.Evaluate(args[0])
+		if !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("site name is not valid: %s", err))
+		} else {
+			cmd.siteName = args[0]
+		}
+	}
+
+	return validationErrors
```
fgiorgetti, line 62: No error is returned when deleting an invalid site name.

internal/cmd/skupper/site/nonkube/site_update.go
```diff
@@ -2,33 +2,236 @@ package nonkube
...
+		if site == nil || err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("site %s must exist in namespace %s to be updated", cmd.siteName, cmd.namespace))
+		} else {
+			// save existing values
+
+		}
+
+		routerAccessName := "router-access-" + cmd.siteName
+		routerAccess, err := cmd.routerAccessHandler.Update(routerAccessName)
+		if err == nil && routerAccess != nil {
```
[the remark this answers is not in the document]
lynnemorrison: @fgiorgetti I was trying to allow a way the user could remove routerAccess with an update command.
If with an updated command the user doesn't set the enableLinkAccess then routerAccess is removed.  

Do you think it is better to leave routerAccess if user had already configured it with the create command?
Once configured with the create or an update command the only way to delete routerAccess would be then for the user to delete the site.

internal/cmd/skupper/site/nonkube/site_create.go
```diff
@@ -146,7 +134,7 @@ func (cmd *CmdSiteCreate) Run() error {
 		},
 		Spec: v2alpha1.SiteSpec{
 			Settings:   cmd.options,
-			LinkAccess: cmd.linkAccessType,
+			LinkAccess: "default",
```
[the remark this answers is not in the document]
lynnemorrison: Should be removed.

internal/cmd/skupper/site/nonkube/site_delete.go
lynnemorrison: added warning

internal/cmd/skupper/site/nonkube/site_update.go
```diff
@@ -2,33 +2,236 @@ package nonkube
...
+			validationErrors = append(validationErrors, fmt.Errorf("bindhost is not valid: a valid IP address or hostname is expected"))
+		} else {
+			cmd.newSettings.bindHost = cmd.Flags.BindHost
+		}
+	}
+	if cmd.Flags != nil && len(cmd.Flags.SubjectAlternativeNames) != 0 {
+		for _, name := range cmd.Flags.SubjectAlternativeNames {
+			ip := net.ParseIP(name)
+			ok, _ := hostStringValidator.Evaluate(name)
+			if !ok || ip == nil {
```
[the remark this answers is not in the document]
lynnemorrison: fixed

internal/cmd/skupper/site/nonkube/site_update.go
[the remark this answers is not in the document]
fgiorgetti: When I look at the update command for the connector resource, I noticed that I do not need to provide all the flags. So if I omit the routing-key on an update, the previous one is preserved.

I was expecting the same behavior here. If the `--enable-link-access` was set during site crete, but not passed (not changed) during an update, then the original value is preserved (enable-link-access=true).

So what seems to be missing here is to load the original state of the enable-link-access field based on the current input state. If user wants to disable it, they can provide `--enable-link-access=false`. Wdyt?

internal/cmd/skupper/site/nonkube/site_create.go
```diff
@@ -77,29 +69,34 @@ func (cmd *CmdSiteCreate) ValidateInput(args []string) []error {
...
 		if !ok {
-			validationErrors = append(validationErrors, fmt.Errorf("link access type is not valid: %s", err))
+			validationErrors = append(validationErrors, fmt.Errorf("output type is not valid: %s", err))
 		}
 	}
 
-	if !cmd.Flags.EnableLinkAccess && len(cmd.Flags.LinkAccessType) > 0 {
-		validationErrors = append(validationErrors, fmt.Errorf("for the site to work with this type of linkAccess, the --enable-link-access option must be set to true"))
+	if cmd.Flags != nil && cmd.Flags.BindHost == "" && cmd.Flags.EnableLinkAccess {
+		validationErrors = append(validationErrors, fmt.Errorf("bindhost should not be empty"))
```
fgiorgetti, line 82: I am wondering why not set `0.0.0.0` as the default value for bind host?
It is currently required, but if enable-link-access is not true, it is not used for anything.

internal/nonkube/client/fs/site_handler.go
```diff
@@ -34,6 +38,79 @@ func (s *SiteHandler) Add(resource v2alpha1.Site) error {
...
+	var context v2alpha1.Site
+	fileName := name + ".yaml"
+
+	if opts.RuntimeFirst == true {
+		// First read from runtime directory, where output is found after bootstrap
+		// has run.  If no runtime sites try and display configured sites
+		err, file := s.ReadFile(s.pathProvider.GetRuntimeNamespace(), fileName, common.Sites)
+		if err != nil {
+			if opts.LogWarning {
+				fmt.Println("Site not initialized yet")
```
fgiorgetti, line 51: I believe we should write this warning message to stderr.
The reason is that if you do `skupper site status my-site-name -o yaml > site.yaml` the warning message is included into the output file.

internal/nonkube/client/fs/site_handler.go
```diff
@@ -34,6 +38,79 @@ func (s *SiteHandler) Add(resource v2alpha1.Site) error {
...
+
+func (s *SiteHandler) List() ([]*v2alpha1.Site, error) {
+	var sites []*v2alpha1.Site
+
+	// First read from runtime directory, where output is found after bootstrap
+	// has run.  If no runtime sites try and display configured sites
+	path := s.pathProvider.GetRuntimeNamespace()
+	err, files := s.ReadDir(path, common.Sites)
+	if err != nil {
+		fmt.Println("Site not initialized yet")
```
fgiorgetti, line 96: Same thing here, maybe better writing it to stderr.

internal/cmd/skupper/site/nonkube/site_update.go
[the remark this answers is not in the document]
fgiorgetti: Now the router access is being preserved or removed properly, but I noticed that when I run the update
command the "linkAccess: default" is being added, no matter what.

internal/nonkube/client/fs/site_handler.go
lynnemorrison: changed to stderr

internal/nonkube/client/fs/site_handler.go
lynnemorrison: fixed

internal/cmd/skupper/site/nonkube/site_create.go
lynnemorrison: defaulted bind-host

internal/cmd/skupper/site/nonkube/site_update.go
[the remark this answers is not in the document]
lynnemorrison: accidently left linkAccess in the update.   It is no removed.

internal/cmd/skupper/site/nonkube/site_create.go
```diff
@@ -48,12 +46,10 @@ func (cmd *CmdSiteCreate) NewClient(cobraCommand *cobra.Command, args []string)
 }
 
 func (cmd *CmdSiteCreate) ValidateInput(args []string) []error {
-
 	var validationErrors []error
-
-	if cmd.Flags.ServiceAccount != "" {
```
nluaces, line 54: Why is this warning deleted?

nluaces (approved):

internal/cmd/skupper/site/nonkube/site_create.go
lynnemorrison: looks like a mistake.   I'll add it back

nluaces (approved):

fgiorgetti (approved):
LGTM

PR #1795, opened 2024-11-20
Add wait flag to sites commands

nluaces:
#### Wait for a given status
New flag that allows the user to wait for a status (pending, configured, ready). 
- Default behaviour, waiting until "ready" for 30 seconds
- This change affects the commands: `skupper site create` and `skupper site update`
- Example: `skupper site create my-site --wait configured`

#### Wait for the site to be deleted
New flag tat allows the user to wait or not for the site to be deleted.
- Default behaviour, waiting until the site do not exist.
- This change affects the command: `skupper site delete`
- Example: `skupper site delete --wait=false`

These flags are available only for Kubernetes platforms for now.


lynnemorrison (approved):

internal/cmd/skupper/site/kube/site_create.go
```diff
@@ -173,25 +183,37 @@ func (cmd *CmdSiteCreate) WaitUntil() error {
...
 		resource, err := cmd.Client.Sites(cmd.Namespace).Get(context.TODO(), cmd.siteName, metav1.GetOptions{})
 		if err != nil {
 			return err
 		}
 
-		if resource != nil && resource.IsConfigured() {
+		isStatusReached := false
+
+		switch cmd.status {
+		case "pending":
```
grs, line 197: I don't think 'pending' is a state you wait for. It is an indication that no state has been settled yet. One option is to wait until a 'configured' condition has been recorded, whether true or false (if false there is likely to be an error).

nluaces: That makes sense. The wait flag also allows to wait for `configured`, but following the Refdog documentation (https://skupperproject.github.io/refdog/commands/site/create.html), when a user indicates `--wait pending` it could mean that they don't expect to wait at all, right? so in that case I just need to skip the WaitUntil step. Do you agree with that?

grs: I'm disagreeing with the refdog documentation in that I don't think 'pending' should be one of the options to wait. @ssorj

nluaces: I have removed the `pending` value (d917edfd92b141691e186b64e06e94321787702f) from the possible values to include in the `--wait` flag, in order to unblock this pull request until until we find a consensus.

grs (approved):
Looks fine to me. One question: what do we want the behaviour to be if something goes wrong with the configuration (i.e. there is an error)? Do we want to just timeout, or do we want to react to that error condition and exit with the error?

internal/cmd/skupper/common/flags.go
```diff
@@ -70,6 +70,10 @@ for other Kubernetes flavors, loadbalancer is the default.`
 	FlagDescListenerPort = "The port of the local listener"
 	FlagNameListenerHost = "host"
 	FlagDescListenerHost = "The hostname or IP address of the local listener. Clients at this site use the listener host and port to establish connections to the remote service."
+
+	FlagNameWait       = "wait"
+	FlagDescWait       = "Wait for the given status before exiting. Choices: pending, configured, ready"
```
fgiorgetti, line 75: As per the latest discussion, I believe we need to remove "pending" from the description.

fgiorgetti (approved):

internal/cmd/skupper/common/flags.go
nluaces: Replaced by `none`, as discussed.

nluaces:
> Looks fine to me. One question: what do we want the behaviour to be if something goes wrong with the configuration (i.e. there is an error)? Do we want to just timeout, or do we want to react to that error condition and exit with the error?

If the site reaches the selected status with error, the message indicates that and exit with an error. If the status was not reached by any means, it exits with error as well.

PR #1801, opened 2024-11-25
Add version command and unit tests

lynnemorrison:
Add version command for kube/nonkube CLI.     The user can redirect
the output of this command to a file.  
Tested existing code that creates a manifest file and default manifest
with environmental variables.    Both work but there is no current CLI 
command to execute either of these features.


nluaces:
This line 
https://github.com/skupperproject/skupper/blob/b2ebdcc683517af90265a8bb65b31df83be49d0f/Makefile#L132C2-L132C12

has to be changed with the new skupper version command (json format) that generates a file called `manifest.json`, like the old command. 

That make target is going to be used in the `config.yaml` configuration that runs in circle ci.

lynnemorrison:
> This line https://github.com/skupperproject/skupper/blob/b2ebdcc683517af90265a8bb65b31df83be49d0f/Makefile#L132C2-L132C12
> 
> has to be changed with the new skupper version command (json format) that generates a file called `manifest.json`, like the old command.
> 
> That make target is going to be used in the `config.yaml` configuration that runs in circle ci.

@nluaces I updated the cmd/manifest/manifest.go to use the updated code.   Please let me know if this work for you.

nluaces:
@lynnemorrison the manifest.go command needs to be removed. It is going to be replaced by `skupper version -o json`. There are a few things to do:

1. Remove file `cmd/manifest/manifest.go`
2. Replace the make target
```
generate-manifest: build-manifest
	./manifest
```

by these lines:

```
generate-manifest: build-cmd
	./skupper version -o json > manifest.json
	
```
you can test that everything is ok by running:

```
make clean
make generate-manifest
```

3. Remove these other lines
```
build-manifest:
	go build -ldflags="${LDFLAGS}"  -o manifest ./cmd/manifest
```

nluaces:
There are other changes to do in `images.go`:

1) Remove these lines (and their dependencies):
```
	// These constants will be soon deprecated.
	ServiceControllerImageName string = "service-controller:main"
	FlowCollectorImageName     string = "flow-collector:main"
	SiteControllerImageName    string = "site-controller:main"
```
They are V1 images that we are not going to use in V2.

pkg/utils/configs/manifest.go
```diff
@@ -42,7 +45,7 @@ func (manager *ManifestManager) GetDefaultManifestWithEnv() Manifest {
 	}
 }
 
-func (manager *ManifestManager) CreateFile(m Manifest) error {
+func (manager *ManifestManager) CreateFile(m SkupperManifest) error {
```
nluaces, line 48: I think this method can be removed, because no function is calling it.

pkg/images/images.go
```diff
@@ -12,9 +12,10 @@ const (
 	PrometheusServerImageName string = "prometheus:v2.42.0"
 	OauthProxyImageRegistry   string = "quay.io/openshift"
 	OauthProxyImageName       string = "origin-oauth-proxy:4.14.0"
+)
```
nluaces, line 15: `NetworkConsoleCollector` is now `NetworkObserver`

So this line:
```
NetworkConsoleCollectorImageName string = "network-console-collector:v2-latest"
```

should be changed to
```
NetworkObserverImageName string = "network-observer:v2-latest"
```

internal/cmd/skupper/version/nonkube/version.go
```diff
@@ -20,13 +34,56 @@ func NewCmdVersion() *CmdVersion {
...
 
-func (cmd *CmdVersion) InputToOptions() {}
+func (cmd *CmdVersion) Run() error {
+	files := cmd.manifest.GetConfiguredManifest()
+	if cmd.output != "" {
+		encodedOutput, err := utils.Encode(cmd.output, files)
+		if err != nil {
+			return err
+		}
+		fmt.Println(string(encodedOutput))
```
nluaces, line 76: the `string` method can be removed

internal/cmd/skupper/version/kube/version.go
```diff
@@ -28,16 +37,53 @@ func (cmd *CmdVersion) NewClient(cobraCommand *cobra.Command, args []string) {
...
+}
+
+func (cmd *CmdVersion) Run() error {
+	files := cmd.manifest.GetConfiguredManifest()
+	if cmd.output != "" {
+		encodedOutput, err := utils.Encode(cmd.output, files)
+		if err != nil {
+			return err
+		}
+		fmt.Println(string(encodedOutput))
```
nluaces, line 76: the `string` method can be removed

lynnemorrison: done

internal/cmd/skupper/version/nonkube/version.go
lynnemorrison: done

pkg/images/images.go
lynnemorrison: done

pkg/utils/configs/manifest.go
lynnemorrison: done

nluaces (approved):
Looks good to me!

PR #1803, opened 2024-11-26
Initial implementation of skupper diagnose

skitt:
This introduces a framework for diagnostics commands, with an initial implementation of two Kubernetes checks (verifying that the Kubernetes API is accessible, and that the Kubernetes version is supported).

The framework supports simple declaration of Cobra commands constructed from individual diagnostics, and dependencies between diagnostics.

The kind spinner is copied with some adaptations borrowed from the Submariner project.


ssorj:
Hi, Stephen.  We are using "skupper debug" as the place to collect all debugging and troubleshooting operations.  I think this operation should live there as well.  Based on what it's doing, I propose "skupper debug check" as the command name.

I also see that the top level "diagnose" requires a "--kube" to run all the checks.  I think it should instead use the current platform (determined by --platform or SKUPPER_PLATFORM), and thus run all the checks without the extra arg.

As a naming thing, I propose using kube instead of k8s for the short form, since we are already using kube in other places.

skitt:
I’ve updated this to add `skupper debug check` (but no additional checks yet).

PR #1805, opened 2024-11-26
Collector API filtering changes

c-kruse:
* Adds a hasListener attribute to addresses to enable filtering.
* Processes are omitted from the API when the parent site is unknown.
* A process is bound only when there is a listener and connector present.


bartoval (approved):
🙏

PR #1809, opened 2024-11-29
add wait flag for link, listener and connector commands

nluaces:
Continuation of the work done in https://github.com/skupperproject/skupper/pull/1795


internal/cmd/skupper/connector/kube/connector_create.go
```diff
@@ -271,25 +283,54 @@ func (cmd *CmdConnectorCreate) WaitUntil() error {
...
-	if err != nil {
-		return fmt.Errorf("Connector %q not ready yet, check the status for more information\n", cmd.name)
+	if err != nil && connectorCondition == nil {
+		return fmt.Errorf("Connector %q is not %s yet, check the status for more information\n", cmd.name, cmd.status)
+	} else if err != nil && connectorCondition.Status == metav1.ConditionFalse {
+		return fmt.Errorf("Connector %q is %s with errors, check the status for more information\n", cmd.name, cmd.status)
 	}
 
-	fmt.Printf("Connector %q is ready\n", cmd.name)
+	fmt.Printf("Site %q is %s.\n", cmd.name, cmd.status)
```
lynnemorrison, line 334: should this be Connector %q

lynnemorrison (approved):
looks good!!

internal/cmd/skupper/connector/kube/connector_create.go
```diff
@@ -271,25 +283,54 @@ func (cmd *CmdConnectorCreate) WaitUntil() error {
...
 		if err != nil {
 			return err
 		}
 
-		if resource != nil && resource.IsConfigured() {
+		isConditionFound := false
+		isConditionTrue := false
+
+		switch cmd.status {
+		case "ready":
```
grs, line 305: As I recall, for sites the default was to wait for ready unless configured was specified explicitly. Is there a reason this is switched here?

nluaces: I thought that for connectors/listeners the default should be "configured" given that they are not ready unless there is match; otherwise it would timeout always. Am I wrong?

internal/cmd/skupper/connector/kube/connector_create.go
nluaces: yes! thank you!

internal/cmd/skupper/connector/kube/connector_create.go
grs: No, that does indeed make sense!

grs (approved):

Issue #1591, opened 2024-08-14
[V2] Adding Cost attribute to the access token


bartoval:
The cost is an attribute of the link. However, when we create it via access token, the link is automatically created after the access token is Redeemed. Therefore, it is necessary to set this property when an access token is created and pass it to the link

PR #1810, opened 2024-12-02
Renaming of CRD fields

grs:


grs:
Also adds linkCost to AccessToken Spec, fixing #1591

cmd/controller/deploy_cluster_scope.yaml
```diff
@@ -151,8 +151,8 @@ rules:
   - connectors/status
   - attachedconnectors
   - attachedconnectors/status
-  - attachedconnectoranchors
```
nluaces, line 154: These lines need to be updated as well here:

https://github.com/skupperproject/skupper/blob/fc25f23ac8a827c301eb955ff795a82b08366f91/deployments/helm/templates/deploy_controller.yaml#L162

nluaces (approved):

cmd/controller/deploy_cluster_scope.yaml
grs: Fixed

scripts/update-codegen.sh
```diff
@@ -18,30 +20,18 @@ set -o errexit
...
-  [[ -d "${TMP_DEST}/github.com/skupperproject/skupper" ]] && \
-  ! ${VERIFY_ONLY} && rm -rf "${TMP_DEST}"
-}
-trap "cleanup" EXIT SIGINT
+kube::codegen::gen_helpers \
+    --boilerplate "${SCRIPT_ROOT}/scripts/boilerplate.go.txt" \
+    "${SCRIPT_ROOT}/pkg/apis"
 
-API_VERSION=`grep k8s.io/apimachinery go.mod | awk '{print $NF}'`
-go get -d k8s.io/code-generator@${API_VERSION}
```
fgiorgetti, line 35: Maybe this could be kept as an alternative, in case there is not a `./vendor` directory?
After that, we could change it a bit to copy kube-gen.sh to SCRIPT_ROOT, like:
```
cp ${GOPATH}/pkg/mod/k8s.io/code-generator\@${API_VERSION}/kube_codegen.sh ${SCRIPT_ROOT}
```

grs: I believe the tools.go causes the code-generator to be considered a dependency, whether vendored or not. I am mostly copying from the 'example'.  The recent change to golang version  means generate-groups.sh is no longer available.

fgiorgetti: ... and possibly add vendor/ to .gitignore?

grs: I've pushed an update that I believe should find the module if not using vendoring. Also added vendor to .gitignore.

PR #1820, opened 2024-12-05
Change naming of tcpConnector and tcpListener elements in the router config

grs:
This makes it simpler and less brittle to derive the name of the kuberenetes resource with which the config element originates.


bartoval (approved):
thanks @grs ! 🙏

PR #1823, opened 2024-12-06
Fixes grant server updating grant status before site is ready

fgiorgetti:
* AccessGrant status.redeemed count was being increased before generateLinkConfig was able to resolve endpoints, returning the "Could not resolve any endpoints for requested link" error and exhausting the AccessGrant before it is actually consumed
* Fixed null pointer issue found (eventually) while deleting a namespace for an active site

### DIFF
--- a/pkg/kube/grants/grants.go
+++ b/pkg/kube/grants/grants.go
@@ -1,6 +1,7 @@
 package grants
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


fgiorgetti:
@grs I have noticed the problem above while testing an automated setup/teardown scenario.
Basically I got `checkAndUpdateAccessToken` called after the generator (which now writes to a buffer first).
This way, if the generator fails, the accessgrant.status.redeemed count is not increased.
It seems to work, but I am not sure, though, if that is the best to do here.

grs:
> @grs I have noticed the problem above while testing an automated setup/teardown scenario. Basically I got `checkAndUpdateAccessToken` called after the generator (which now writes to a buffer first). This way, if the generator fails, the accessgrant.status.redeemed count is not increased. It seems to work, but I am not sure, though, if that is the best to do here. Maybe the accessgrant should be moved to ready, only after the site endpoints are also defined.

What I've been doing is waiting until both the site and the grant are ready before trying to create a token.

I think you ask two good questions here but I need a bit of time to think about them.

fgiorgetti:
@grs I believe this one can be closed, right?

PR #1824, opened 2024-12-06
Add sourceHost and destHost to flow metrics.

scwhaley:
This change adds sourceHost and destHost to the flow metrics produced by the flow collector. It acomplishes this by adding those labels to the promethus metrics, and changing the aggregation key for metric aggreagation in the flow collector.

This approach may change to one that does not require new labels on the prometheus metric. To be discussed.


Karen-Schoener:
Closing in favor of PR https://github.com/skupperproject/skupper/pull/1857

Issue #1827, opened 2024-12-10
[v2] Access/Audit Logging from network observer


c-kruse:
In v1, the flow-collector logs an event when the collector first adds a vanflow record and again when the collector deletes that record. This compromise kept log noise low enough for the always-on nature of the collector logging, and managed to expose enough information for 1. developers and operators to attempt to debug collector and network issues and 2. to serve as a plausible audit trail answering the question "which service was a skupper network connection or request routed to?".

Moving forward, we should consider a dedicated facility for the later case, in the form of  a configurable "network access logging" option to the network-observer.
* Should be configurable (method and granularity TBD)
* Should emit predictable structured events, with the fullest amount of relevant context known (source and dest host, process, site, proxy host, router trace path, etc.)
* [non functional] Appears familiar to end users who understand traditional network and application load balancer access logs, but are not "skupper nerds."

c-kruse:
@Karen-Schoener this one made me think of you all with the work you've been doing on flow metrics. Would be curious to hear your thoughts sometime. I suspect our horizon for this would be post-2.0, so no rush!

PR #1828, opened 2024-12-10
Collector logging

c-kruse:
Reopens https://github.com/skupperproject/skupper/pull/1783

I believe we've sorted out that this sort of a protocol-level vanflow logging option is useful for debugging different scenarios, but is not a particularly nice solution for users looking for an audit trail for service traffic. I've opened https://github.com/skupperproject/skupper/issues/1827 to tackle that use case.

### DIFF
--- a/cmd/network-observer/profiles.go
+++ b/cmd/network-observer/profiles.go
@@ -15,8 +15,8 @@ var (
 			Match: flowlog.NewRecordTypeSet(
 				vanflow.SiteRecord{}, vanflow.RouterRecord{},
 				vanflow.ProcessRecord{}, vanflow.ConnectorRecord{},
-				vanflow.LinkRecord{}, vanflow.ListenerRecord{},
-				vanflow.RouterAccessRecord{}, vanflow.LogRecord{},
+				vanflow.ListenerRecord{}, vanflow.RouterAccessRecord{},
+				vanflow.LogRecord{},
 			),
 			Strategy: flowlog.RateLimited(1.0, 32),
 		}, {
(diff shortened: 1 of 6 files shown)


c-kruse:
Example network-observer startup logs: https://gist.githubusercontent.com/c-kruse/ac0d7da6c5261a45708b03cf4c0c3730/raw/fe3a60d6a3e94c24a3fccb9c3607ab0c47fc5a65/gistfile1.txt

cmd/network-observer/profiles.go
```diff
@@ -0,0 +1,55 @@
...
+	// loggingProfileMinimal logs 1 vanflow event per second (with bursts up to 32)
+	// reduces Link Record noise to 1 every ~20s.
+	// excludes network flow records
+	loggingProfileMinimal = []flowlog.Rule{
+		{
+			Priority: 5,
+			Match: flowlog.NewRecordTypeSet(
+				vanflow.SiteRecord{}, vanflow.RouterRecord{},
+				vanflow.ProcessRecord{}, vanflow.ConnectorRecord{},
+				vanflow.LinkRecord{}, vanflow.ListenerRecord{},
```
fgiorgetti, line 18: Should LinkRecord be part of both rules (priority 5 and 1)?

fgiorgetti (approved):
Played a bit with the different strategies and they seem to work well.

cmd/network-observer/profiles.go
c-kruse: Ah, there was no real sense in that. Dropped it from the p 5 rule (lowest wins). Thank you!

PR #1831, opened 2024-12-11
Export helm templates as an artifact for releases

nluaces:
**Changes:**
- Moved the helm chart to the directory `charts`, to follow current conventions and to allow the inclusion of more charts.
- Added AppVersion: 2.0.0 (that differs from helm chart version, 1.0.0).
- Exported the two variants of the template scope: namespace, cluster.
- Added additional step in the circle CI config file to include the files as part of the release artifact.
- Deleted deployment yamls to avoid duplication.


nluaces:
@grs this is not done yet, but what do you think, in the case of the cluster scope, about not creating the skupper namespace, and deploy the controller in the namespace that the user has configured?

charts/skupper-setup/Chart.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: v2
 name: skupper-setup
 description:  Helm chart for setting up Skupper.
-version: 0.1.0
\ No newline at end of file
+version: 0.1.0
+appVersion: 2.0.0
```
c-kruse, line 5: Thinking through the mechanics of releasing aloud here.

Today when we cut a release branch and update images.go with the next release tag. What happens in helm land?

I'd propose something like this:

**In "dev" branches - v2/main**
Chart version is chosen as a pre-release semver v2 - so for now 0.1.0-devel
Chart App Version is set to match the tag our CI produces - so v2-latest for now. Along with this, we plumb this field into the default values for skupper image tags.

**On release branch Create**
Drop the prerelease-label from the chart version: `0.1.0-devel` -> `0.1.0`.
Chart App version updated to next planned release: `v2-latest` -> `2.0.0`.

**Preparing for next patch on existing release branch**
Bump version `0.1.0` -> `0.1.1`
Bump app version `2.0.0` -> `2.0.1`

nluaces: good idea!

grs:
> @grs this is not done yet, but what do you think, in the case of the cluster scope, about not creating the skupper namespace, and deploy the controller in the namespace that the user has configured?

Makes sense to me.

grs (approved):

c-kruse (approved):

nluaces:
@grs I have tried to remove the skupper namespace, and the creation of the ClusterRoleBinding failed because the associated ServiceAccount requires an explicit namespace to be defined. I will leave it as it is for now.

grs:
> @grs I have tried to remove the skupper namespace, and the creation of the ClusterRoleBinding failed because the associated ServiceAccount requires an explicit namespace to be defined. I will leave it as it is for now.

Presumably that can be templated at some point(?) so that it also takes the user specified namespace if provided. What you have already is a big improvement though, so lets get that in and we can enhance at a later stage if needed. 

:shipit:

PR #1833, opened 2024-12-11
Fixes internal console-auth implementation

c-kruse:
When console-auth is set to internal, load user credentials at startup and validate HTTP basic auth requests using the preloaded values. Consequently, adding or rotating user credentials now requires the flow-collector container to be restarted.


grs (approved):

nluaces (approved):

ajssmith (approved):
lgtm thanks!

PR #1835, opened 2024-12-12
show running pods version

nluaces:
The user can deploy declarative any version of the skupper controller/network-observer without specifying environment variables. For that reason, with this change the kube flavor of the version command will check the skupper components that are running (in ~all namespaces~ the current namespace) to get the version that the component is actually running.

The priority order is as follows:
  Running pod > Environment variable > Constant (images.go)

Affected components:
- controller
- kube-adaptor
- router
- network-observer

### DIFF
--- a/internal/cmd/skupper/version/kube/version.go
+++ b/internal/cmd/skupper/version/kube/version.go
@@ -1,7 +1,9 @@
 package kube
 
 import (
+	"context"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"text/tabwriter"
 
@@ -38,6 +40,7 @@ func (cmd *CmdVersion) NewClient(cobraCommand *cobra.Command, args []string) {
 
 	if err == nil {
 		cmd.namespace = cli.Namespace
+		cmd.KubeClient = cli.GetKubeClient()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/cmd/skupper/version/kube/version.go
```diff
@@ -58,10 +61,27 @@ func (cmd *CmdVersion) ValidateInput(args []string) []error {
 }
 
 func (cmd *CmdVersion) InputToOptions() {
+
+	mapRunningPods := make(map[string]string)
+
+	if cmd.KubeClient != nil {
+		// search for running pods in all namespaces
+		runningPodList, err := cmd.KubeClient.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{LabelSelector: "app.kubernetes.io/part-of=skupper"})
+		if err != nil {
```
lynnemorrison, line 70: I think the network-observer and network-observer-prometheus does not have the label 
app.kubernetes.io/part-of=skupper  
but it does have 
app.kubernetes.io/part-of: skupper-network-observer

pkg/images/image_utils.go
```diff
@@ -225,23 +225,69 @@ func GetImage(imageNames map[string]string, imageRegistry string, enableSHA bool
...
 	names := make(map[string]string)
 	switch component {
 	case "router":
 		// skupper router has two components
-		names[RouterImageEnvKey] = RouterImageName
-		names[AdaptorImageEnvKey] = AdaptorImageName
-		registry = GetImageRegistry()
+		runningImage := runningPods["router"]
+		envImage := os.Getenv(RouterImageEnvKey)
+		defaultImage := RouterImageName
```
lynnemorrison, line 238: do we need defaultImage variable.   Can we just assign names[RuterImageEnvKey] = RouterImageName

pkg/images/image_utils.go
```diff
@@ -253,31 +299,54 @@ func GetImages(component string, enableSHA bool) []SkupperImage {
...
 	var image string
 
 	switch component {
 	case "router":
-		image = os.Getenv(RouterImageEnvKey)
-		if image == "" {
-			image = RouterImageName
+		runningImage := runningPods["router"]
+		envImage := os.Getenv(RouterImageEnvKey)
+		defaultImage := RouterImageName
```
lynnemorrison, line 316: same comment here about defaultImage

lynnemorrison (changes requested):
this is a good change!!

pkg/images/image_utils.go
nluaces: solved! (aeb4da27f4b1bbf947326063d45f14e3a6655504)

pkg/images/image_utils.go
nluaces: solved! (aeb4da27f4b1bbf947326063d45f14e3a6655504)

pkg/utils/configs/manifest.go
```diff
@@ -42,14 +43,14 @@ func (manager *ManifestManager) GetDefaultManifestWithEnv() Manifest {
...
+func getSkupperConfiguredImages(components []string, enableSHA bool, runningPods map[string]string) SkupperManifest {
 	var manifest SkupperManifest
 
 	for _, component := range components {
 		var image SkupperComponent
 		image.Component = component
-		image.Version = images.GetImageVersion(component)
-		image.Images = images.GetImages(component, enableSHA)
+		image.Version = images.GetImageVersion(component, runningPods)
+		image.Images = images.GetImages(component, enableSHA, runningPods)
```
grs, line 53: Minor personal preference: I think it would be cleaner to do a two stage lookup here, checking the pods first then falling back to image utils. That way the image utils could be left to do its one thing, i.e. determine the configured images.

grs (approved):

pkg/utils/configs/manifest.go
nluaces: done

grs (approved):
Thank you!

lynnemorrison (approved):

fgiorgetti (approved):
LGTM

Issue #1836, opened 2024-12-12
[v2] Use Consistent Network Observer API "Slashiness"


c-kruse:
**Is your feature request related to a problem? Please describe.**

The network observer API spec has URIs defined with a trailing slash, and the implementation adds 301 redirects from the "bare" URI to the "slash-y" URI. I.e. `/sites` 301s to `/sites/`. We have decided to align the "slashiness" of the API to the skupper console's current convention, dropping the slash to avoid unneeded redirects.

**Describe the solution you'd like**

* URIs should not end in a trailing slash.
* Accessing resources using a trailing slash will redirect to the desired un-slashed URI.

PR #1837, opened 2024-12-12
Update Network Observer API Route Slash Conventions

c-kruse:
Previously the network-observer API had opted to include trailing slashes in resource URIs (e.g. `sites/`) and had 301 redirects from `sites` to `sites/`. This change reverses the previous convention. Now the API prefers `sites` over `sites/`, and will redirect from the later to the former.

Closes https://github.com/skupperproject/skupper/issues/1836


bartoval (approved):
+1

PR #1846, opened 2024-12-13
Add generate option to commands 

lynnemorrison:
Add generate option to site/listener/connector commands for kube and nonkube
Remove output flag from same commands.

### DIFF
--- a/internal/cmd/skupper/connector/connector.go
+++ b/internal/cmd/skupper/connector/connector.go
@@ -25,6 +25,7 @@ skupper connector status my-connector`,
 	cmd.AddCommand(CmdConnectorStatusFactory(config.GetPlatform()))
 	cmd.AddCommand(CmdConnectorUpdateFactory(config.GetPlatform()))
 	cmd.AddCommand(CmdConnectorDeleteFactory(config.GetPlatform()))
+	cmd.AddCommand(CmdConnectorGenerateFactory(config.GetPlatform()))
 
 	return cmd
 }
@@ -49,7 +50,6 @@ skupper connector create backend 8080 --workload deployment/backend`,
 	cmd.Flags().StringVar(&cmdFlags.Host, common.FlagNameHost, "", common.FlagDescHost)
 	cmd.Flags().StringVar(&cmdFlags.TlsCredentials, common.FlagNameTlsCredentials, "", common.FlagDescTlsCredentials)
 	cmd.Flags().StringVar(&cmdFlags.ConnectorType, common.FlagNameConnectorType, "tcp", common.FlagDescConnectorType)
-	cmd.Flags().StringVarP(&cmdFlags.Output, common.FlagNameOutput, "o", "", common.FlagDescOutput)
 	if configuredPlatform == types.PlatformKubernetes {
 		cmd.Flags().BoolVarP(&cmdFlags.IncludeNotReadyPods, common.FlagNameIncludeNotReadyPods, "i", false, common.FlagDescIncludeNotRead)
 		cmd.Flags().StringVarP(&cmdFlags.Selector, common.FlagNameSelector, "s", "", common.FlagDescSelector)
@@ -110,7 +110,6 @@ func CmdConnectorUpdateFactory(configuredPlatform types.Platform) *cobra.Command
 	cmd.Flags().StringVar(&cmdFlags.Host, common.FlagNameHost, "", common.FlagDescHost)
 	cmd.Flags().StringVar(&cmdFlags.TlsCredentials, common.FlagNameTlsCredentials, "", common.FlagDescTlsCredentials)
 	cmd.Flags().StringVar(&cmdFlags.ConnectorType, common.FlagNameConnectorType, "tcp", common.FlagDescConnectorType)
-	cmd.Flags().StringVarP(&cmdFlags.Output, common.FlagNameOutput, "o", "", common.FlagDescOutput)
 	cmd.Flags().IntVar(&cmdFlags.Port, common.FlagNameConnectorPort, 0, common.FlagDescConnectorPort)
 	if configuredPlatform == types.PlatformKubernetes {
 		cmd.Flags().BoolVarP(&cmdFlags.IncludeNotReadyPods, common.FlagNameIncludeNotReadyPods, "i", false, common.FlagDescIncludeNotRead)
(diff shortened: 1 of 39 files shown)


nluaces:
I have found an error testing skupper site generate:

```
 $ skupper site generate west --enable-link-access
 
Warning: --bind-host flag is not supported on this platform
kind: Site
metadata:
  name: west
  namespace: default
spec:
  linkAccess: default
```

This warning should not be printed.

nluaces:
sorry, my mistake

lynnemorrison:
> I have found an error testing skupper site generate:
> 
> ```
>  $ skupper site generate west --enable-link-access
>  
> Warning: --bind-host flag is not supported on this platform
> apiVersion: skupper.io/v2alpha1
> kind: Site
> metadata:
>   name: west
>   namespace: default
> spec:
>   linkAccess: default
> ```
> 
> This warning should not be printed.

looks like the warning has been removed from the site create, but was left in generate code.    You can't set binding-host on nonkube platform anyway.   I'm wondering if we should also remove:

if cmd.Flags != nil && cmd.Flags.SubjectAlternativeNames != nil && len(cmd.Flags.SubjectAlternativeNames) > 0 {
		fmt.Println("Warning: --subject-alternative-names flag is not supported on this platform")
	}

nluaces:
> looks like the warning has been removed from the site create, but was left in generate code. You can't set binding-host on nonkube platform anyway

Originally, the warning was made to let the user know that flag was not meant for non-kube commands, and by default had an empty value.

Later, there was another change that put a default value in that flag.

Before doing the release, I had to remove the warning from site create because that warning was always printed in non-kube environments.

I think you can do the same in this case, removing the warning, otherwise is going to be printed. 

> @nluaces I'm wondering if we should also remove:
>if cmd.Flags != nil && cmd.Flags.SubjectAlternativeNames != nil && >len(cmd.Flags.SubjectAlternativeNames) > 0 {
>fmt.Println("Warning: --subject-alternative-names flag is not supported on this platform")
>}

It has to be removed if there is a value by default in that flag, otherwise I think it can remain, but I don't have a strong opinion about this, so if you think that is better to remove it as well I'm fine with it.

internal/cmd/skupper/connector/connector.go
```diff
@@ -164,3 +163,42 @@ func CmdConnectorDeleteFactory(configuredPlatform types.Platform) *cobra.Command
...
 	return cmd
 }
+
+func CmdConnectorGenerateFactory(configuredPlatform types.Platform) *cobra.Command {
+	kubeCommand := kube.NewCmdConnectorGenerate()
+	nonKubeCommand := nonkube.NewCmdConnectorGenerate()
+
+	cmdConnectorGenerateDesc := common.SkupperCmdDescription{
+		Use:   "generate <name> <port>",
+		Short: "Generate a connector",
```
nluaces, line 173: Suggestion: The `short` description should be: "Generate a connector resource"

internal/cmd/skupper/connector/kube/connector_generate.go
```diff
@@ -0,0 +1,250 @@
...
+	KubeClient          kubernetes.Interface
+}
+
+func NewCmdConnectorGenerate() *CmdConnectorGenerate {
+
+	return &CmdConnectorGenerate{}
+}
+
+func (cmd *CmdConnectorGenerate) NewClient(cobraCommand *cobra.Command, args []string) {
+	cli, err := client.NewClient(cobraCommand.Flag("namespace").Value.String(), cobraCommand.Flag("context").Value.String(), cobraCommand.Flag("kubeconfig").Value.String())
```
nluaces, line 45: The generate commands need to work in a environment where there is no kubernetes installed because their main purpose is to just generate a resource. Hence, all the checks that depends on existing resources (like checking if a site or workload exists) don't apply to the `generate` commands. The only validations that have to be maintained are the ones that check if the values introduced by the user are incorrect.

nluaces:
tested:
- Hello world example (ok)
- generate commands in a running kubernetes environments (ok)
- generate commands without a kubernetes environment (needs to be fixed)

Note: the link generate command needs to have a kubernetes environment given that its behavior is different: it creates a credential resource and a secret in kubernetes in order to be able to generate a link resource.

internal/cmd/skupper/connector/kube/connector_generate.go
lynnemorrison: thanks, I wasn't sure about that, will remove.

internal/cmd/skupper/connector/connector.go
lynnemorrison: done

internal/cmd/skupper/connector/kube/connector_generate.go
lynnemorrison: @nluaces for workload option once you remove getting the values from k8 I need to put something in the selector field.
```
kind: Connector
metadata:
  name: backup
  namespace: default
spec:
  port: 8080
  routingKey: backup
  selector: ?????
  type: tcp
```
I'm verifying the user doesn't specify more than one host/selector/workload 
but if user specifies workload what do you think I should put in the selector field:   
options:
   1.  just put in what user specified   <deployment/backend>
 
   2. specify where I get the selector from    
        if deployment:
        selector: <deployment.Spec,Selector.MatchLabels>
   3. ????

nluaces: @lynnemorrison if there is no kubernetes environment we can't obtain the selector of a given workload, because this has to work just as a generator of resources.

My suggestion is that if we don't have a selector, to do not specify the selector field in the resource.

cc: @ssorj

lynnemorrison: I've removed all kubernetes calls from the generate commands.

internal/cmd/skupper/connector/kube/connector_generate.go
```diff
@@ -42,12 +38,10 @@ func NewCmdConnectorGenerate() *CmdConnectorGenerate {
 }
 
 func (cmd *CmdConnectorGenerate) NewClient(cobraCommand *cobra.Command, args []string) {
-	cli, err := client.NewClient(cobraCommand.Flag("namespace").Value.String(), cobraCommand.Flag("context").Value.String(), cobraCommand.Flag("kubeconfig").Value.String())
+	cli, err := client.NewGenerateCmdClient(cobraCommand.Flag("namespace").Value.String())
```
nluaces, line 41: This is not necessary, I suggest to just keep
`cmd.namespace = cli.Namespace` in the NewClient method. 

If generate methods are not checking with a client it is fine to just not call them.

internal/cmd/skupper/listener/kube/listener_generate.go
```diff
@@ -39,12 +36,10 @@ func NewCmdListenerGenerate() *CmdListenerGenerate {
 }
 
 func (cmd *CmdListenerGenerate) NewClient(cobraCommand *cobra.Command, args []string) {
-	cli, err := client.NewClient(cobraCommand.Flag("namespace").Value.String(), cobraCommand.Flag("context").Value.String(), cobraCommand.Flag("kubeconfig").Value.String())
+	cli, err := client.NewGenerateCmdClient(cobraCommand.Flag("namespace").Value.String())
```
nluaces, line 39: This is not necessary, I suggest to just keep
cmd.namespace = cli.Namespace in the NewClient method.

If generate methods are not checking with a client it is fine to just not call them.

internal/cmd/skupper/site/kube/site_generate.go
```diff
@@ -38,11 +37,9 @@ func NewCmdSiteGenerate() *CmdSiteGenerate {
 }
 
 func (cmd *CmdSiteGenerate) NewClient(cobraCommand *cobra.Command, args []string) {
-	cli, err := client.NewClient(cobraCommand.Flag("namespace").Value.String(), cobraCommand.Flag("context").Value.String(), cobraCommand.Flag("kubeconfig").Value.String())
+	cli, err := client.NewGenerateCmdClient(cobraCommand.Flag("namespace").Value.String())
```
nluaces, line 40: This is not necessary, I suggest to just keep
cmd.namespace = cli.Namespace in the NewClient method.

If generate methods are not checking with a client it is fine to just not call them.

internal/kube/client/client.go
```diff
@@ -134,3 +134,15 @@ func NewClient(namespace string, context string, kubeConfigPath string) (*KubeCl
 
 	return c, nil
 }
+
+func NewGenerateCmdClient(namespace string) (*KubeClient, error) {
```
nluaces, line 138: Following my previous comments, this function is not necessary, because it is returning a KubeClient that has not been initialized. It is better to just not use the KubeClient at all.

lynnemorrison: removed

internal/cmd/skupper/site/kube/site_generate.go
lynnemorrison: removed function and only set namespace in NewClient

internal/cmd/skupper/listener/kube/listener_generate.go
lynnemorrison: removed function call and only set namespace in NewClient

nluaces (approved):
Looks good to me. Thank you.

nluaces:
@lynnemorrison there are some merge conflicts, I will merge this pull request after their resolution

nluaces:
there are still some merge conflicts to solve

PR #1847, opened 2024-12-16
Prevent crashes while checking for listeners on namespaces without sites

fgiorgetti:


grs (commented):
Sorry, this was a regression from one of my commits. I believe the correct fix is here: https://github.com/skupperproject/skupper/pull/1848

(CheckListener is suppoed to be called even where there is no site yet).

PR #1854, opened 2024-12-17
[v1] Add sourceHost and destHost to flow metrics (with improvements)

Karen-Schoener:
Fixed an issue with PR https://github.com/skupperproject/skupper/pull/1824.

Tested skupper console with this fix.

Tested with external vm-client and external vm-server.  
Verified that in/out metrics displayed as expected in the skupper console.

Tested with hello-world frontend/backend deployments.
Verified that in/out metrics displayed as expected in the skupper console.

### DIFF
--- a/pkg/flow/collector.go
+++ b/pkg/flow/collector.go
@@ -75,37 +75,37 @@ func (fc *FlowCollector) NewMetrics(reg prometheus.Registerer) *collectorMetrics
 				Name: "flows_total",
 				Help: "Total Flows",
 			},
-			[]string{"sourceSite", "destSite", "address", "protocol", "direction", "sourceProcess", "destProcess"}),
+			[]string{"sourceSite", "destSite", "address", "protocol", "direction", "sourceProcess", "destProcess", "sourceHost", "destHost"}),
 		octets: prometheus.NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "octets_total",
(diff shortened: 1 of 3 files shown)


Karen-Schoener:
Hi @ajssmith @c-kruse When you have a minute, I had a question on the 2 minute wait that I noticed in needForSiteProcess.
https://github.com/skupperproject/skupper/blob/main/pkg/flow/flow_mem_driver.go#L2799

I was just curious why a 2 minute wait is needed here...   

Btw, as a next step, I'll work on an approach that does not add new labels.

Thanks, Karen

c-kruse:
@Karen-Schoener, Andy probably has a better answer for you here, but  I believe that delay is intended to guard against the collector prematurely associating flows with a site-process or site-client when a new process record could be en-route from a controller.

One hundred and twenty seconds is probably a bit much, but the collector very much receives vanflow messages in wild orders (especially from different event sources - like is the case with ProcessRecords and FlowRecords which come from the "Controller" part of pkg/flow and FlowRecords that come from the skupper router.) Because these site server and site client processes hang around (don't tend to get cleaned up until the site goes away), and because flow pairs get permanently associated with the process, we were trying to be careful to avoid this.

Karen-Schoener:
Closing in favor of PR https://github.com/skupperproject/skupper/pull/1857

PR #1855, opened 2024-12-17
Network observer chart

c-kruse:
Adds a network-observer chart that
* Runs a single pod with network-observer, prometheus, and one of two reverse proxy containers.
* Has options for service type, ingress and openshift routes
* Obligatory TLS (from pod to edge - LB, Ingress, Route) using either openshift service certs, user-provided cert or skupper issued cert.
* A couple of manifest.yamls generated from the chart.


charts/network-observer/values.yaml
```diff
@@ -0,0 +1,150 @@
...
+  tag: ""
+
+# prometheus configures the promehteus container image
+prometheus:
+  repository: "docker.io/prom/prometheus"
+  pullPolicy: IfNotPresent
+  tag: "v3.0.1"
+  # securityContext for prom container.
+  # With openshift SCC MUST be overwritten to null
+  # With native k8s, MUST be set
```
c-kruse, line 24: This bit had me pulling my hair out. Someone please help.

charts/network-observer/README.md
```diff
@@ -0,0 +1,42 @@
...
+### Authorization
+
+The network observer pod contains a reverse proxy that handles authorization
+and TLS termination for the read only application that binds only to localhost.
+When authorization strategy is "basic", nginx is configured as the proxy, and
+can be configured with user-provided htpasswd file contents or a secret name.
+When authorization strategy is "openshift" an oauth2 proxy is used instead, and
+is configured to use the cluster identity provider for authorization. Openshift
+auth only works with ingress type Route.
+
```
nluaces, line 42: Should we mention here or somewhere how to access the console?

c-kruse: I just added a _little_ more, but that still pretty much assumes you know what you're doing. Is the section on ingress kind of what you meant?

Since it changes so much based on configuration we could put a templated output in NOTES.txt so that when it gets installed through helm it pops up with instructions to get to the Route|Ingress|LoadBalancer Service|NodePort Service or how to port-forward the ClusterIP service.

nluaces: The NOTES.txt thing is a good idea!

charts/network-observer/README.md
fgiorgetti, line 1: It would be nice to have a usage section here, based on the Makefile targets.
Another one that might be too obvious, so if you think it is useless, let's keep it as is.

fgiorgetti (approved):
It works great. Tested using minikube with generic template, using
a custom htpasswd file and using openshift auth strategy.

charts/network-observer/README.md
c-kruse: That's a good idea! Thinking it through aloud here, I'm not sure exactly what to do for that yet. I did not really envision anyone but us + our CI running those make targets to tag on to a release.

As an end state I think it may make sense to have something like this documented for a quickstart: `kubectl apply -f https://github.com/skupperproject/skupper/releases/download/{{ Latest Release }}/skupper-network-observer-generic.yaml` Or maybe even somewhere cleaner ` kubectl apply -f https://skupper.io/v2/network-observer-install.yaml`?

fgiorgetti: Looks good.

PR #1856, opened 2024-12-19
bootstrap script: skip namespace argument if it was not provided

nluaces:
That way the behaviour is similar to the `system setup` command


ajssmith (approved):
lgtm, thnx

PR #1875, opened 2025-01-13
add nonkube link status support

### LINKED ISSUE #1871: [v2] [non-kube] skupper link status implementation is missing
**Describe the bug**
Skupper link status implementation is missing

**How To Reproduce**
$skupper link status
command not supported by the selected platform

**Expected behavior**
Show the status defined statically in the runtime resource

lynnemorrison:
add command and unit tests for nonkube link status 

fixes #1871

### DIFF
--- a/internal/cmd/skupper/link/nonkube/link_status.go
+++ b/internal/cmd/skupper/link/nonkube/link_status.go
@@ -2,16 +2,24 @@ package nonkube
 
 import (
 	"fmt"
+<<<<<<< HEAD
+=======
+	"os"
+	"text/tabwriter"
+>>>>>>> 5122e05 (add nonkube link status support)
 
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
+	"github.com/skupperproject/skupper/pkg/utils/validator"
 	"github.com/spf13/cobra"
 )
 
 type CmdLinkStatus struct {
-	CobraCmd  *cobra.Command
-	Flags     *common.CommandLinkStatusFlags
-	Namespace string
-	siteName  string
+	linkHandler *fs.LinkHandler
+	CobraCmd    *cobra.Command
+	Flags       *common.CommandLinkStatusFlags
+	namespace   string
+	linkName    string
 }
 
 func NewCmdLinkStatus() *CmdLinkStatus {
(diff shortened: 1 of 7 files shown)


internal/nonkube/client/fs/path_provider.go
```diff
@@ -12,6 +12,10 @@ func (p *PathProvider) GetNamespace() string {
 	return api.GetHostNamespaceHome(p.Namespace) + "/" + string(api.InputSiteStatePath)
 }
 
-func (p *PathProvider) GetRuntimeNamespace() string {
+func (p *PathProvider) GetRuntimeResourceNamespace() string {
 	return api.GetHostNamespaceHome(p.Namespace) + "/" + string(api.RuntimeSiteStatePath)
 }
+
+func (p *PathProvider) GetRuntimeNamespace() string {
+	return api.GetHostNamespaceHome(p.Namespace) + "/" + string(api.RuntimePath)
```
fgiorgetti, line 20: This is actually retrieving the "static" links generated for a given site (those used by other sites to connect
with it), not the active outgoing links a given site is using.

It seems like `GetRuntimeResourceNamespace()` is what should have been used here, as it retrieves from 
`runtime/resources` instead. With that, I believe `GetRuntimeNamespace` is not needed as it is only used by
link_handler.

lynnemorrison: ok thanks will fix.

internal/cmd/skupper/link/nonkube/link_status.go
```diff
@@ -2,28 +2,129 @@ package nonkube
...
 
-func (cmd *CmdLinkStatus) ValidateInput(args []string) []error { return nil }
-func (cmd *CmdLinkStatus) InputToOptions()                     {}
 func (cmd *CmdLinkStatus) Run() error {
-	return fmt.Errorf("command not supported by the selected platform")
+	opts := fs.GetOptions{RuntimeFirst: true, LogWarning: true}
+	if cmd.linkName == "" {
+		links, err := cmd.linkHandler.List()
+		if links == nil || err != nil {
+			fmt.Println("no links found:")
```
fgiorgetti, line 72: The output when no link is defined has:

```
no links found:
failed to read directory: open /home/fgiorget/.local/share/skupper/namespaces/west/runtime/resources/links: no such file or directory
```

If would be nice to add an extra condition to ignore such errors.

internal/cmd/skupper/link/nonkube/link_status.go
```diff
@@ -2,28 +2,129 @@ package nonkube
...
+			"NAME", "STATUS"))
+		for _, link := range links {
+			status := "Not Ready"
+			if link.IsConfigured() {
+				status = "Ok"
+			}
+			name := link.Name
+			for index, endpoint := range link.Spec.Endpoints {
+				if endpoint.Name == types.InterRouterRole {
+					name = name + "-" + link.Spec.Endpoints[index].Host
```
fgiorgetti, line 87: I belive you just need to show the name here.

internal/cmd/skupper/link/nonkube/link_status.go
```diff
@@ -2,28 +2,129 @@ package nonkube
...
+		sites, err := cmd.siteHandler.List()
+		if sites != nil && err == nil {
+			endPointType := types.InterRouterRole
+			if sites[0].Spec.Edge {
+				endPointType = types.EdgeRole
+			}
+
+			for index, endpoint := range link.Spec.Endpoints {
+				if endpoint.Name == endPointType {
+					endpointName = link.Spec.Endpoints[index].Host + ":" + link.Spec.Endpoints[index].Port
```
fgiorgetti, line 116: I am still unable to see the endpoints.

internal/cmd/skupper/link/nonkube/link_status.go
lynnemorrison: The problem with just showing the name, is that if the user wants to get the detailed output for the  skupper link status then they don't have the full name to use for that command:

skupper link status -n west -p podman
NAME				                STATUS
link-router-access-west-0.0.0.0	Not Ready

skupper link status -n west -p podman link-router-access-west-0.0.0.0
Name:		link-router-access-west
Status:		Not Ready
Cost:		2
TlsCredentials:	link-router-access-west
Endpoint:	0.0.0.0:55671

internal/cmd/skupper/link/nonkube/link_status.go
lynnemorrison: @fgiorgetti what would you like here?  

don't show any output if links not found?     This be the default.   have a flag --detail  or --errors that will show above logs.   Do you want that for all status commands?

internal/cmd/skupper/link/nonkube/link_status.go
lynnemorrison: Do you have a site configured?   I look up the site to determine if it is an edge or inter-router to display the correct endpoints.    If there is no site no endpoints are displayed.    There is also a bug https://github.com/skupperproject/skupper/pull/1880  that causes a problem and needs to be fixed.    This is just incorrect name of the site directory.

internal/cmd/skupper/link/nonkube/link_status.go
lynnemorrison: only showing link name removed the binding-host value.

internal/cmd/skupper/link/nonkube/link_status.go
lynnemorrison: defaulting to use inter-router

internal/cmd/skupper/link/nonkube/link_status.go
fgiorgetti: I believe that in case there is no links defined under `runtime/resources` it should simply say: `no links found`.

internal/cmd/skupper/link/nonkube/link_status.go
fgiorgetti: It works for me now. Only issue I am seeing is that the following message is displayed: `Site not initialized yet`.

lynnemorrison: fixed

internal/cmd/skupper/link/nonkube/link_status.go
fgiorgetti: For sites where no links are defined, along with the error, it also says: "Site not initialized yet".

lynnemorrison: Now if no links are found will only show:
"No links found"

fgiorgetti (approved):

PR #1876, opened 2025-01-13
Cache the configured platform

skitt:
This avoids attempting to read the configuration file etc. dozens of times during startup.

In tests which change the configured platform, the cached value is cleared.

### DIFF
--- a/internal/config/platform.go
+++ b/internal/config/platform.go
@@ -7,11 +7,17 @@ import (
 
 	"github.com/skupperproject/skupper/api/types"
 	"github.com/skupperproject/skupper/internal/utils"
+	"k8s.io/utils/ptr"
 )
 
 var (
-	Platform string
+	Platform           string
+	configuredPlatform *types.Platform
 )
+
+func ClearPlatform() {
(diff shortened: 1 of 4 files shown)


internal/config/platform.go
```diff
@@ -165,14 +175,15 @@ func GetPlatform() types.Platform {
...
-		return types.PlatformDocker
+		configuredPlatform = &platform
 	case types.PlatformSystemd:
-		return types.PlatformSystemd
+		configuredPlatform = &platform
 	default:
-		return types.PlatformKubernetes
+		configuredPlatform = ptr.To(types.PlatformKubernetes)
 	}
+	return *configuredPlatform
```
fgiorgetti, line 186: I believe this last return is not needed, right?

fgiorgetti (commented):
It looks good to me. I would just recommend rebasing on top of main,
as the code still shows the "PlatformInfo" type,  which has been removed.

fgiorgetti:
@skitt this PR was impacted by #1939. Could you update it, then I will merge next.

PR #1878, opened 2025-01-14
add config layout

ajssmith:
Have a common base structure for operator, etc.


nluaces:
How about adding some make targets to generate the CSV and the bundle (or both at he same time)?

nluaces (approved):

PR #1884, opened 2025-01-14
Add check that token filename is not a directory

### LINKED ISSUE #1858: 'skupper token issue' behavior when <filename> is a directory
**Describe the bug**
When a directory rather than a filename is provided, the command does not indicate an error

**How To Reproduce**
create site
mkdir tmp
skupper token issue ./tmp

**Expected behavior**
Command should detect target is not a filename

**Environment details**
 - Skupper CLI: preview-2

lynnemorrison:
check that filename is not a directory and warn on failure

fixes #1858


ajssmith (approved):
lgtm

nluaces (approved):
Looks good to me

Issue #1891, opened 2025-01-16
[v2] Vanflow Spec Conflict in Site Version


c-kruse:
**Describe the bug**
pkg/vanflow.Site has a Version field with vanflow codepoint set to 52. In v1, this codepoint was an unspecified attribute that was only used between the vanflow controller and collector on SITE records.

In v2, the router has become involved in producing SITE records in the case of non-kube sites, and codepoint 52 has been given to an unrelated attribute. This puts us in an awkward spot.

Replace Site Version with the existing VFLOW_ATTRIBUTE_BUILD_VERSION = 32 attribute. Sites that use different version attributes (either the errant 52 or the desired 32) will not correctly observe each other's versions.

PR #1893, opened 2025-01-17
Fix Unspecified Vanflow SITE Attributes

c-kruse:
Previously the vanflow SITE record was using an unspecified attributes to denote the site version and policy status. This breaking change moves the SITE Version attribute to the specified buildVersion attribute (code point 32) in the spec and removes the now-deprecated Policy attribute. Versions running the old unspecified version attribute (2.0.0-previews 1 and 2) will not observe the site version from sites with the corrected attribute and vice versa. The Policy attribute has been unused since 1.x.

### DIFF
--- a/pkg/kube/flow/status.go
+++ b/pkg/kube/flow/status.go
@@ -514,7 +514,6 @@ func asSiteInfo(site vanflow.SiteRecord) network.SiteInfo {
 		Namespace: dref(site.Namespace),
 		Platform:  dref(site.Platform),
 		Version:   dref(site.Version),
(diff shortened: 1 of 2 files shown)


c-kruse:
Fixes https://github.com/skupperproject/skupper/issues/1891

This is an unfortunate breaking change, but the consequences should be pretty benign. The network-observer API will label incompatible sites as version "unknown", and skupper site status will have a blank (`-`) version for incompatible sites.

bartoval (approved):

PR #1898, opened 2025-01-21
Fix flaky tests

grs:


nluaces (approved):
Thank you for taking care of the retry_test. 

Looks good to me.

PR #1901, opened 2025-01-22
update links from /v2/ to /main/

pwright:
@nluaces  does this make sense? or too broad?


nluaces (approved):
Thank you Paul, looks good to me. I will add @c-kruse to double-check the network observer part.

c-kruse (changes requested):
Thanks Paul! Just those two false positives made it in, but otherwise looks good to me!

pwright:
@c-kruse apols, this is an old PR, pinging you again on it.

c-kruse (approved):

PR #1910, opened 2025-01-22
Ammed network observer chart presets

c-kruse:
Drops the openshift variant of the network observer until openshift auth can be declaratively configured.

Adds an unsecured option for demo environments.

Adds a httpbasic option WITHOUT a static user defined.

Updates the nginx configuration so that it can be dynamically reloaded from kuberentes secrets.


fgiorgetti (approved):
LGTM

ajssmith (approved):
lgtm2

PR #1920, opened 2025-01-24
Updates go build version.Version definition

c-kruse:
Reflect the move from pkg/version.Version to internal/version.Version in the go build flags so that skupper components can correctly report their version.


grs (approved):
Sorry, I missed this!

PR #1923, opened 2025-01-24
Considers modified versions as valid

fgiorgetti:
CI executions in branch v1 are now building the CLI on top of the v1-dev-release tag.

This tag is considered as a valid version by `pkg.utils.Version.ParseVersion`.

With that, the version determined by the CI execution is detected as something like:
`1.0.0-dev-release-commit_hash-modified`.

This is causing existing CLI tests to fail and this PR is probably not the best, but seems
to help with local and CI executions, against changes done on top of the v1 branch.

### DIFF
--- a/pkg/utils/version.go
+++ b/pkg/utils/version.go
@@ -96,7 +96,8 @@ func IsValidFor(actual string, minimum string) bool {
 	}
 	va := ParseVersion(actual)
 	vb := ParseVersion(minimum)
-	return va.IsUndefined() || !va.LessRecentThan(vb)
+	isModified := strings.Contains(va.Qualifier, "-modified")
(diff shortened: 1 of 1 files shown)


fgiorgetti:
@grs @ajssmith @nluaces @c-kruse An alternative would be to change the `ParseVersion` method.
Any thoughts?

grs:
The simplest solution might be to just add a tag to the v1 branch. If it is something that can't be interpreted as a version, then the version checking should be as it was on main. (Alternatively we can regularly tag it to keep incrementing the compatability)

fgiorgetti:
@grs That is true. If we tag it as `v1.8-dev-release`, I believe that the problem is solved
and we do not need to keep tagging afterwards.

PR #1926, opened 2025-01-28
V2 site improvements

grs:
Allows multiple controllers to operate on a cluster. To avoid conflicts, sites can be annotated with `skupper.io/controller=<controller-namespace>/,controller-name>`. This is required for controllers that are not the default controller on the cluster. Namespace scoped controllers always require this annotation on the sites they are to control. Cluster scoped controllers do not, but if running more than one of these, all but one should be set to explicitly require the annotation using the `-sites-require-annotation` option to the controller (or the `SITES_REQUIRE_ANNOTATION` environment variable).

The controller will now update the status of the site to indicate that it is the controller in charge. This involves a small change to the site CRD that will need to be reapplied.

Also includes fixes to correctly handle the recovery and deletion of sites where more than one site definition exists in a namespace.


nluaces (approved):

granzoto (approved):
I tested the changes, I had both cluster-wide controller and namespace scoped controller at the same time, and working properly.
Nice job.

ajssmith (approved):
built controller and ran via operator and ns (sans annotation), verified site status for controller

lgtm

internal/kube/controller/config.go
```diff
@@ -36,5 +46,7 @@ func BoundConfig(flags *flag.FlagSet) (*Config, error) {
 	iflag.StringVar(flags, &c.Namespace, "namespace", "NAMESPACE", "", "The Kubernetes namespace scope for the controller")
 	iflag.StringVar(flags, &c.Kubeconfig, "kubeconfig", "KUBECONFIG", "", "A path to the kubeconfig file to use")
 	iflag.StringVar(flags, &c.WatchNamespace, "watch-namespace", "WATCH_NAMESPACE", metav1.NamespaceAll, "The Kubernetes namespace the controller should monitor for controlled resources (will monitor all if not specified)")
+	iflag.StringVar(flags, &c.Name, "name", "CONTROLLER_NAME", "", "A name identifying the controller. If mot specified it will be deduced from the hostname.")
```
fgiorgetti, line 49: Typo: "mot".

internal/kube/controller/controller.go
```diff
@@ -108,9 +152,23 @@ func (c *Controller) init(stopCh <-chan struct{}) error {
...
 	for _, site := range c.siteWatcher.List() {
+		if !c.matches(site) {
+			c.log.Info("Ignoring site as it does not have required annotation",
+				slog.String("name", site.Name),
+				slog.String("namespace", site.Namespace),
+			)
+			continue
+		}
+		if !siteRecovery.IsActive(site) {
+			c.log.Info("Skipping site recovery as it not the active site",
```
fgiorgetti, line 166: typo: "it not"

cmd/controller/main.go
fgiorgetti, line 1: It would be nice to log an initial message saying that the controller instance
will only process sites annotated with: `skupper.io/controller=x/y`, in case
`RequireAnnotation` is set to true.

fgiorgetti:
I have tested a scenario with a controller watching all namespaces and then a separate controller watching
only its given namespace (sk2). Although everything seems to be working ok, I have found a couple of messages
on the controller logs watching all namespaces that concerned me a bit.
@grs Check it out and let me know your thoughts:
https://gist.github.com/fgiorgetti/da86a85dca97fd21f05df767317b0806

grs:
> Although everything seems to be working ok, I have found a couple of messages
> on the controller logs watching all namespaces that concerned me a bit.

Yes, I noticed that after submitting the PR. The issue is that the SecuredAccess and Certificate resources created automatically by the controller also need to be constrained. I will work on a fix for that but would like to get this merged so that the CRD change makes the rc.

fgiorgetti (approved):

Issue #1916, opened 2025-01-23
[v2] Network Observer Routerlinks missing after kube-adapter restarts


c-kruse:
**Describe the bug**
We noticed that /routerlinks were missing from the observer API after the kube-adapter had restarted on various routers in the network.

I suspect this has to do with the network observer collector Graph implementation: the collector component that is used to quickly traverse more complicated relationships between records (in this example, the peer site for a link - Link->RouterAccess->Router->Site.)

The graph should either be tolerant to records coming and going or should be rebuilt after disruptions or on a periodic schedule.

**Steps to reproduce**

* Start a multi-site VAN with the network observer installed at one site.
* Choose one site, and repeatedly kill the kube-adaptor container in the skupper-router deployment until the site disappears from the skupper console topology view. `kubectl exec -it deployments/skupper-router -c kube-adaptor -- sh -c "kill 1"` 
* Allow the skupper-router deployment to recover. When the issue has been reproduced, the site will appear but the router links will be missing.

Example:

![Image](https://github.com/user-attachments/assets/5cc809c0-a919-459b-99ea-f0f39edd0084)

PR #1931, opened 2025-01-28
Fixes race conditions in network observer graph

c-kruse:
Adds logic to re-add child to parent record relationships when a parent record is added. Fixes several sorts of problems that arise when parent records, like a SITE record, are deleted and re-added after a disruption. Also resolves a known race condition (TODO) where the relation between a connector's Host and processes in a site cannot be made unless we already have the router record (to complete the relationship between connector->router->site.)

Fixes https://github.com/skupperproject/skupper/issues/1916

### DIFF
--- a/cmd/network-observer/internal/collector/collector.go
+++ b/cmd/network-observer/internal/collector/collector.go
@@ -117,7 +117,11 @@ func (c *Collector) updateGraph(event changeEvent, stor readonly) {
 	if !ok {
 		return
 	}
-	c.graph.Reindex(entry.Record)
+	if _, ok := event.(addEvent); ok {
+		c.graph.Index(entry.Record)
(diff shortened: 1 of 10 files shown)


c-kruse:
Difficult to totally confidently say this is resolved, but on one VAN with two network observers running, one latest and one with this change, the observer running latest managed to reproduce the issue 3 of 4 attempts, and the one with this change did not reproduce the issue.
(latest left, this change set right)
image

fgiorgetti (approved):
Tried it and I was able to easily reproduce it, as you mentioned using v2-dev.
With this PR I could not reproduce the issue. Everytime the other sites came
back, the links appeared again.

bartoval (approved):
+1000

PR #1933, opened 2025-01-29
Rename Network Observer API parent fields

c-kruse:
Renames Connector and Listener parent to routerId. Renames Process parent and parentName to siteId and siteName. Renames Router parent to siteId.


c-kruse:
image: `quay.io/ckruse/skupper/network-observer:pull-1933`

bartoval (approved):
+1

PR #1936, opened 2025-02-03
add means of configuring site sizing

grs:
This adds a mechanism for configuring the container resources in the skupper-router pod for sites based on a 'size' concept.

Sizes are defined by configmaps with a `skupper.io/site-sizing` label whose value is the name of the size. It is the users responsibility to sure that size names are unique. The valid keys in the map are , `router-cpu-request`, `router-cpu-limit`, `router-memory-request`, `router-memory-limit`, `adaptor-cpu-request`, `adaptor-cpu-limit`, `adaptor-memory-request` and `adaptor-memory-limit`.

A site definition can specify an explicit `size` in the `settings` map within the spec. If a seize has been configured for that name, then any of the defined resource settings within that size will be applied to the skupper-router deployment for the site.

If one of the defined sizes has the annotation `skupper.io/default-site-sizing`, that is considered the default size to be used for any site where there is not an explicit size specified (or the explicit size specified does not match a configured size).

### DIFF
--- a/internal/kube/site/site.go
+++ b/internal/kube/site/site.go
@@ -19,6 +19,7 @@ import (
 	internalclient "github.com/skupperproject/skupper/internal/kube/client"
 	kubeqdr "github.com/skupperproject/skupper/internal/kube/qdr"
 	"github.com/skupperproject/skupper/internal/kube/site/resources"
+	"github.com/skupperproject/skupper/internal/kube/site/sizing"
 	"github.com/skupperproject/skupper/internal/qdr"
 	"github.com/skupperproject/skupper/internal/site"
 	"github.com/skupperproject/skupper/internal/version"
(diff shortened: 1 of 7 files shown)


internal/kube/site/site.go
```diff
@@ -181,9 +184,22 @@ func (s *Site) reconcile(siteDef *skupperv2alpha1.Site, inRecovery bool) error {
...
 	// 3. deployment
+	size, err := s.sizes.GetSizing(s.site)
+	if err != nil {
+		s.logger.Info("Did not retrieve size for site",
+			slog.String("namespace", s.site.Namespace),
+			slog.String("name", s.site.Name),
+			slog.String("reason", err.Error()),
+		)
+	} else {
+		s.logger.Info("Sizing for site",
```
fgiorgetti, line 195: Minor recommendation to log it with Debug as it keeps showing up in the logs every 30 seconds.

fgiorgetti (approved):
Works great! Nice job as usual.

nluaces (approved):

PR #1940, opened 2025-02-03
Process flows more efficiency in network observer

c-kruse:
The network observer handles an in memory buffer of application and transport level flow records and associated state that can be very large. This changes how those collections are traversed for processing. Previously the collections of flow state were copied on each reconcile loop, resulting in high CPU and memory utilization. Now each reconcile job limits itself to copying the relevant portions.


c-kruse:
Example resource usage for the latest `v2-dev` network-observer vs `pull-1940` on a VAN servicing around ~1000 requests per second.

Memory usage -25%
image

CPU usage  - much better.
image

```
$ oc adm top pod
NAME                                          CPU(cores)   MEMORY(bytes)
skupper-router-84dc449fcd-gfzf4               250m         71Mi
pull-1940-network-observer-7dccf945f5-dnf5h   55m          1839Mi
v2-dev-network-observer-648f5f8f6c-vsxl6      701m         2398Mi
...
```

fgiorgetti (approved):

Issue #1932, opened 2025-01-29
[v2] flaky test TestCmdSiteDelete_ValidateInput/kubernetes_flags_are_not_valid_on_this_platform


c-kruse:
**Describe the bug**
Observed this test failing in CircleCI: https://app.circleci.com/pipelines/github/skupperproject/skupper/4581/workflows/fedb06f2-1b8e-44fd-ae90-d21de862d94b/jobs/29968/parallel-runs/0/steps/0-107

```
=== RUN   TestCmdSiteDelete_ValidateInput
File written to /home/circleci/.local/share/skupper/namespaces/test/input/resources/site/my-site.yaml
=== RUN   TestCmdSiteDelete_ValidateInput/site_name_is_not_specified
=== RUN   TestCmdSiteDelete_ValidateInput/site_name_is_not_specified,_all
=== RUN   TestCmdSiteDelete_ValidateInput/site_name_is_nil
=== RUN   TestCmdSiteDelete_ValidateInput/site_name_is_not_valid
=== RUN   TestCmdSiteDelete_ValidateInput/more_than_one_argument_is_specified
=== RUN   TestCmdSiteDelete_ValidateInput/site_doesn't_exist
=== RUN   TestCmdSiteDelete_ValidateInput/kubernetes_flags_are_not_valid_on_this_platform
Warning: --context flag is not supported on this platform
Warning: --kubeconfig flag is not supported on this platform
    validate_input.go:17: assertion failed: error is not nil: site my-site does not exist
--- FAIL: TestCmdSiteDelete_ValidateInput (0.13s)
    --- PASS: TestCmdSiteDelete_ValidateInput/site_name_is_not_specified (0.00s)
    --- PASS: TestCmdSiteDelete_ValidateInput/site_name_is_not_specified,_all (0.00s)
    --- PASS: TestCmdSiteDelete_ValidateInput/site_name_is_nil (0.00s)
    --- PASS: TestCmdSiteDelete_ValidateInput/site_name_is_not_valid (0.00s)
    --- PASS: TestCmdSiteDelete_ValidateInput/more_than_one_argument_is_specified (0.00s)
    --- PASS: TestCmdSiteDelete_ValidateInput/site_doesn't_exist (0.00s)
    --- FAIL: TestCmdSiteDelete_ValidateInput/kubernetes_flags_are_not_valid_on_this_platform (0.00s)
```

[...]

PR #1946, opened 2025-02-06
Replace /.local/share dir by /tmp directory

nluaces:
#1932

### DIFF
--- a/internal/cmd/skupper/site/nonkube/site_delete_test.go
+++ b/internal/cmd/skupper/site/nonkube/site_delete_test.go
@@ -27,9 +27,9 @@ func TestCmdSiteDelete_ValidateInput(t *testing.T) {
 		expectedError     string
 	}
 
-	homeDir, err := os.UserHomeDir()
+	err := os.Setenv("SKUPPER_OUTPUT_PATH", "/tmp/skupper")
 	assert.Check(t, err == nil)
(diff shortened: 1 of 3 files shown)


internal/cmd/skupper/site/nonkube/site_delete_test.go
```diff
@@ -27,9 +27,9 @@ func TestCmdSiteDelete_ValidateInput(t *testing.T) {
 		expectedError     string
 	}
 
-	homeDir, err := os.UserHomeDir()
+	err := os.Setenv("SKUPPER_OUTPUT_PATH", "/tmp/skupper")
```
c-kruse, line 30: I am not super confident this actually addresses the issue - feels to me like it just _moves_ the problem to `/tmp`.
I'd suggest a per-test temp directory so that parallel executions do not clobber each other. the testing package as a method for this: https://pkg.go.dev/testing#T.TempDir

Secondarily, setting environment variables in tests can be dicey as well - lots of low level os thread and global mutable state problems can pop up. Maybe there's a way to factor that out and pass the path in to the command setup rather than pulling from os.Environ? If that's too much of a bear, I think it's okay for now, but if it is still flakey with a random temp dir that's where i'd look.

nluaces: changed!

I think that the refactor of the command would take a little bit of thinking, but if this happens again I would definitely refactor that.

c-kruse (approved):
Thank you!

PR #1960, opened 2025-02-13
imrpoved handing of conflict between multiple controllers

grs:
Introduce a namespace level skupper setting that determines which controller should control that namespace. This allows a simpler way to handle multiple controllers on a cluster.

Namespace level settings are specified through a ConfigMap in the namespace called 'skupper'. The only currently recognised setting has key 'controller'. The value for this is the controller name, optionally qualified by a namespace if the controller is running in a different namespace from the site, indicating the controller that has control over this namespace.

Namespace scoped controllers must always set this configuration. Cluster scoped controllers will assume they can control any namespace that doesn't contain an explicit configuration for a different controller.

If operating multiple cluster scoped controllers on the same controller, all but one of them should set the -require-explicit-control argument to true.

### DIFF
--- a/internal/kube/controller/config.go
+++ b/internal/kube/controller/config.go
@@ -11,21 +11,21 @@ import (
 )
 
 type Config struct {
-	GrantConfig         *grants.GrantConfig
-	SecuredAccessConfig *securedaccess.Config
-	Namespace           string
-	Kubeconfig          string
-	WatchNamespace      string
-	Name                string
-	RequireAnnotation   bool
+	GrantConfig            *grants.GrantConfig
+	SecuredAccessConfig    *securedaccess.Config
+	Namespace              string
+	Kubeconfig             string
+	WatchNamespace         string
+	Name                   string
+	RequireExplicitControl bool
 }
 
 func (c *Config) WatchingAllNamespaces() bool {
 	return c.WatchNamespace == metav1.NamespaceAll
 }
 
-func (c *Config) sitesRequireAnnotation() bool {
-	return !c.WatchingAllNamespaces() || c.RequireAnnotation
+func (c *Config) requireExplicitControl() bool {
+	return !c.WatchingAllNamespaces() || c.RequireExplicitControl
 }
 
 func BoundConfig(flags *flag.FlagSet) (*Config, error) {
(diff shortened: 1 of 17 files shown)


internal/kube/controller/config.go
```diff
@@ -47,6 +47,6 @@ func BoundConfig(flags *flag.FlagSet) (*Config, error) {
 	iflag.StringVar(flags, &c.Kubeconfig, "kubeconfig", "KUBECONFIG", "", "A path to the kubeconfig file to use")
 	iflag.StringVar(flags, &c.WatchNamespace, "watch-namespace", "WATCH_NAMESPACE", metav1.NamespaceAll, "The Kubernetes namespace the controller should monitor for controlled resources (will monitor all if not specified)")
 	iflag.StringVar(flags, &c.Name, "name", "CONTROLLER_NAME", "", "A name identifying the controller. If not specified it will be deduced from the hostname.")
-	iflag.BoolVar(flags, &c.RequireAnnotation, "sites-require-annotation", "SITES_REQUIRE_ANNOTATION", false, "If set, this controller instance will only process sites with an annotation matching the controller's namespace qualified name")
+	iflag.BoolVar(flags, &c.RequireExplicitControl, "require-explicit-control", "REQUIRE_EXPLICIT_CONTROL", false, "If set, this controller instance will only process resources in which there is a ConfigMap named skupper with an entry 'controller' whose value matches the controller's namespace qualified name. Controllers watching a single namespace require that ConfigMap regardless of this setting.")
```
fgiorgetti, line 50: `REQUIRE_EXPLICIT_CONTROL` needs to be added into `config/overlays/namespace/patch-add-env-var-watch-namespace.yaml`

fgiorgetti (commented):
It definitely provides a much better user experience.
One comment is that I noticed an unexpected behavior in the controller watching all namespaces as it updated the loadbalancer IP for the namespace scoped certificate.

```
time=2025-02-14T20:31:00.002Z level=INFO msg="Resolved endpoints for fg-ns/skupper-router: [{inter-router 10.106.253.185 55671 } {edge 10.106.253.185 45671 }]"
time=2025-02-14T20:31:00.002Z level=INFO msg="Changing hosts for Certificate fg-ns/skupper-site-server from [skupper-router skupper-router.fg-ns] to [skupper-router skupper-router.fg-ns 10.106.253.185]"
```

internal/kube/controller/config.go
grs: > `REQUIRE_EXPLICIT_CONTROL` needs to be added into `config/overlays/namespace/patch-add-env-var-watch-namespace.yaml`

That option does not apply to namespaces scoped controllers as they always require explicit control. What I did add was a configmap to grant that control in the namespaces scoped case.

The option there is only required where you want to run multiple cluster scoped controllers (which at present I am assuming is a fringe use case)

nluaces (approved):

grs:
> It definitely provides a much better user experience. One comment is that I noticed an unexpected behavior in the controller watching all namespaces as it updated the loadbalancer IP for the namespace scoped certificate.
> 
> ```
> time=2025-02-14T20:31:00.002Z level=INFO msg="Resolved endpoints for fg-ns/skupper-router: [{inter-router 10.106.253.185 55671 } {edge 10.106.253.185 45671 }]"
> time=2025-02-14T20:31:00.002Z level=INFO msg="Changing hosts for Certificate fg-ns/skupper-site-server from [skupper-router skupper-router.fg-ns] to [skupper-router skupper-router.fg-ns 10.106.253.185]"
> ```

Sorry! I have pushed a fix that should resolve that now! Thanks for catching it!

grs:
> > Namespace level settings are specified through a ConfigMap in the namespace called 'skupper'.
> 
> Just to confirm that there is no issue with this: Unlike the kustomize setup, using the helm chart will install the ConfigMap in the current namespace (the one in which the controller lives), given that the both scopes (namespace, cluster) don't use any `skupper` namespace.

I've added a configmap to the config which gets included in the namespace scoped yaml and in the namespace scoped install from helm chart.

nluaces:
> I've added a configmap to the config which gets included in the namespace scoped yaml and in the namespace scoped install from helm chart.

Indeed, I just checked it just now; I just thought by your description that the configmap needed to be located in a `skupper` namespace. Sorry for my misunderstanding.

PR #1962, opened 2025-02-18
add install and uninstall commands

nluaces:
Added the following commands to skupper:

https://skupperproject.github.io/refdog/commands/system/install.html
https://skupperproject.github.io/refdog/commands/system/uninstall.html

### DIFF
--- a/internal/cmd/skupper/system/nonkube/system_uninstall.go
+++ b/internal/cmd/skupper/system/nonkube/system_uninstall.go
@@ -3,7 +3,9 @@ package nonkube
 import (
 	"errors"
 	"fmt"
+	"github.com/skupperproject/skupper/api/types"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/config"
 	"github.com/skupperproject/skupper/internal/nonkube/bootstrap"
 	"github.com/spf13/cobra"
 )
@@ -11,7 +13,7 @@ import (
 type CmdSystemUninstall struct {
 	CobraCmd         *cobra.Command
 	Namespace        string
-	SystemUninstall  func() error
+	SystemUninstall  func(string) error
 	CheckActiveSites func() (bool, error)
 	Flags            *common.CommandSystemUninstallFlags
 	forceUninstall   bool
@@ -37,6 +39,10 @@ func (cmd *CmdSystemUninstall) ValidateInput(args []string) error {
 		validationErrors = append(validationErrors, fmt.Errorf("this command does not accept arguments"))
 	}
 
+	if config.GetPlatform() != types.PlatformPodman && config.GetPlatform() != types.PlatformDocker {
+		validationErrors = append(validationErrors, fmt.Errorf("the selected platform is not supported by this command. There is nothing to uninstall"))
+	}
+
 	if cmd.Flags != nil && !cmd.Flags.Force {
 		activeSites, err := cmd.CheckActiveSites()
 		if err != nil {
@@ -57,13 +63,13 @@ func (cmd *CmdSystemUninstall) InputToOptions() {
 
 func (cmd *CmdSystemUninstall) Run() error {
 
-	err := cmd.SystemUninstall()
+	err := cmd.SystemUninstall(string(config.GetPlatform()))
 
 	if err != nil {
 		return fmt.Errorf("failed to uninstall : %s", err)
 	}
 
-	fmt.Println("Podman infrastructure for Skupper is now uninstalled")
+	fmt.Printf("Platform %s infrastructure for Skupper is now uninstalled\n", string(config.GetPlatform()))
 
 	return nil
 }
--- a/internal/cmd/skupper/system/nonkube/system_uninstall.go
+++ b/internal/cmd/skupper/system/nonkube/system_uninstall.go
@@ -1,0 +1,71 @@
+package nonkube
+
+import (
+	"errors"
+	"fmt"
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/nonkube/bootstrap"
+	"github.com/spf13/cobra"
+)
+
+type CmdSystemUninstall struct {
+	CobraCmd         *cobra.Command
+	Namespace        string
+	SystemUninstall  func() error
+	CheckActiveSites func() (bool, error)
+	Flags            *common.CommandSystemUninstallFlags
+	forceUninstall   bool
+}
+
+func NewCmdSystemUninstall() *CmdSystemUninstall {
+
+	skupperCmd := CmdSystemUninstall{}
+
+	return &skupperCmd
+}
+
+func (cmd *CmdSystemUninstall) NewClient(cobraCommand *cobra.Command, args []string) {
+	cmd.SystemUninstall = bootstrap.Uninstall
+	cmd.CheckActiveSites = bootstrap.CheckActiveSites
+	cmd.Namespace = cobraCommand.Flag("namespace").Value.String()
+}
+
+func (cmd *CmdSystemUninstall) ValidateInput(args []string) error {
+	var validationErrors []error
+
+	if len(args) > 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("this command does not accept arguments"))
+	}
+
+	if cmd.Flags != nil && !cmd.Flags.Force {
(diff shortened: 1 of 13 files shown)


grs (approved):

internal/nonkube/bootstrap/install.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		return err
+	}
+
+	cli, err := internalclient.NewCompatClient(os.Getenv("CONTAINER_ENDPOINT"), "")
+	if err != nil {
+		return fmt.Errorf("failed to create container client: %v", err)
+	}
+
+	_, err = cli.NetworkCreate(&container.Network{Name: "skupper"})
```
fgiorgetti, line 29: Not needed. We do not create bridge networks in V2.

internal/nonkube/bootstrap/uninstall.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func Uninstall() error {
+
+	cli, err := internalclient.NewCompatClient(os.Getenv("CONTAINER_ENDPOINT"), "")
+	if err != nil {
+		return fmt.Errorf("failed to create container client: %v", err)
+	}
+
+	err = cli.NetworkRemove("skupper")
```
fgiorgetti, line 20: Same here (not needed)

internal/nonkube/common/systemd.go
```diff
@@ -50,6 +55,13 @@ type systemdServiceInfo struct {
 	platform            string
 }
 
+type systemdGobal struct {
```
fgiorgetti, line 58: Typo `Gobal`.

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,50 @@ func IsLingeringEnabled(user string) bool {
...
+	return &systemdGobal{
+		getUid:              os.Getuid,
+		command:             exec.Command,
+		rootSystemdBasePath: rootSystemdBasePath,
+		platform:            platform,
+	}, nil
+}
+
+func (sg *systemdGobal) getCmdEnablePodmanSocket() *exec.Cmd {
+	return sg.command("systemctl", "--user", "enable", "--now", "podman.socket")
```
fgiorgetti, line 287: The systemd commands must take into account the current user id.
If `getUid()` is 0, then do not include the `--user` argument (for all commands).

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,50 @@ func IsLingeringEnabled(user string) bool {
...
+	return sg.command("systemctl", "--user", "disable", "--now", "podman.socket")
+}
+
+func (sg *systemdGobal) Enable() error {
+	err := sg.getCmdEnablePodmanSocket().Run()
+	if err != nil {
+		return err
+	}
+
+	err = sg.getCmdStartPodmanSocket().Run()
```
fgiorgetti, line 304: When `enable --now` is executed, it is already started, so no need to start it again I belive.

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,50 @@ func IsLingeringEnabled(user string) bool {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+func (sg *systemdGobal) Disable() error {
+	err := sg.getCmdDisablePodmanSocket().Run()
```
fgiorgetti, line 314: One thing that I am not 100% sure is whether we really want to disable it here or not.

I believe that if we explain that the podman api will be disabled through the
`skupper system uninstall --help`, then it should be fine to disable it here.

internal/cmd/skupper/system/nonkube/system_uninstall.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (cmd *CmdSystemUninstall) ValidateInput(args []string) error {
+	var validationErrors []error
+
+	if len(args) > 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("this command does not accept arguments"))
+	}
+
+	if cmd.Flags != nil && !cmd.Flags.Force {
```
fgiorgetti, line 40: Install only runs for podman platform, but uninstall runs for docker and linux.

fgiorgetti:
One comment here that might impact refdog.

If the platform is `docker`, I believe it would be useful to validate if the current user is able to communicate with the endpoint as well. At least try to see if the client instance (`NewCompatClient`) works ok, as it already validates if the endpoint is reachable.

In case it is being executed as the `root` user (uid 0), it could also try to run the `docker.socket` service, similarly to what
we are doing with the `podman.socket`.

Thoughts @nluaces @ssorj ?

internal/nonkube/bootstrap/install.go
nluaces: removed

internal/nonkube/common/systemd.go
nluaces: included in the description

internal/cmd/skupper/system/nonkube/system_uninstall.go
nluaces: fixed, now the allowed platforms are only docker and podman

nluaces:
> One comment here that might impact refdog.
> 
> If the platform is `docker`, I believe it would be useful to validate if the current user is able to communicate with the endpoint as well. At least try to see if the client instance (`NewCompatClient`) works ok, as it already validates if the endpoint is reachable.
> 
> In case it is being executed as the `root` user (uid 0), it could also try to run the `docker.socket` service, similarly to what we are doing with the `podman.socket`.
> 
> Thoughts @nluaces @ssorj ?

I think that is a good idea, Refdog is already including these commands for non-kube environments. I have changed the code to enable/disable docker and podman sockets depending on the selected platform. Thanks!

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,59 @@ func IsLingeringEnabled(user string) bool {
...
+		return sg.command("systemctl", "start", sg.platform+".socket")
+	}
+	return sg.command("systemctl", "--user", "start", sg.platform+".socket")
+}
+
+func (sg *systemdGlobal) getCmdDisableSocket() *exec.Cmd {
+	if sg.platform == "docker" || sg.getUid() == 0 {
+		return sg.command("systemctl", "disable", sg.platform+".socket")
+	}
+	return sg.command("systemctl", "--user", "disable", sg.platform+".socket")
```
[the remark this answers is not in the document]
nluaces: done

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,59 @@ func IsLingeringEnabled(user string) bool {
...
+func (sg *systemdGlobal) getCmdStartSocket() *exec.Cmd {
+	if sg.platform == "docker" || sg.getUid() == 0 {
+		return sg.command("systemctl", "start", sg.platform+".socket")
+	}
+	return sg.command("systemctl", "--user", "start", sg.platform+".socket")
+}
+
+func (sg *systemdGlobal) getCmdDisableSocket() *exec.Cmd {
+	if sg.platform == "docker" || sg.getUid() == 0 {
+		return sg.command("systemctl", "disable", sg.platform+".socket")
```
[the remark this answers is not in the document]
nluaces: done

internal/nonkube/common/systemd.go
```diff
@@ -260,3 +272,59 @@ func IsLingeringEnabled(user string) bool {
...
+	return &systemdGlobal{
+		getUid:              os.Getuid,
+		command:             exec.Command,
+		rootSystemdBasePath: rootSystemdBasePath,
+		platform:            platform,
+	}, nil
+}
+
+func (sg *systemdGlobal) getCmdEnableSocket() *exec.Cmd {
+	if sg.platform == "docker" || sg.getUid() == 0 {
```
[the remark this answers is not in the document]
nluaces: >I believe that when uid is 0, you can run the command without checking for the platform.
But in the rootless case, we should only run it for podman. This also applies to start and disable methods.

done
> I also noticed that --now was removed, if that is the idea, then start should also be invoked here.

--now was indeed removed, but after enabling the socket, `start` is invoked here inside the function 
```
func (sg *systemdGlobal) Enable() error
```

fgiorgetti (approved):

fgiorgetti:
thanks @nluaces

Issue #1958, opened 2025-02-13
warn users when they create tokens on sites without link access


pwright:
**Is your feature request related to a problem? Please describe.**

After performing the hello world example using v2 preview 2, you can do the following in east ns, even tho link access is disabled:

```
$ skupper token issue ~/token.yaml
Waiting for token status ...

Grant "east-91a87489-8768-44f8-b911-f9d15e932147" is ready
Token file /home/paulwright/token.yaml created

Transfer this file to a remote site. At the remote site,
create a link to this site using the "skupper token redeem" command:

        skupper token redeem <file>

The token expires after 1 use(s) or after 15m0s.
```

This token can never be redeemed.

**Describe the solution you'd like**

A message, something like:

'Link access is not enabled. This token cannot be redeemed.'

**Describe alternatives you've considered**

Disable 'token issue' for sites that do not have link access enabled. 
would require more code for an edge case (grant server enabled, but also a site without link access)

**Additional context**

Depending on the solution implemented, we should consider updating wording at https://skupperproject.github.io/refdog/commands/token/issue.html

lynnemorrison:
@ssorj @nluaces  What I'm thinking is that in the code when creating the token, we verify that there is an active site before creating the token.   I will also add a check that the site has LinkAccess enabled.    If not enabled, then add a warning and not create the token.

"A site must have link access enabled before a token can be created."

lynnemorrison:
fixed by #1964

PR #1964, opened 2025-02-19
[v2] Add check that linkaccess enabled before allowing token to be created

lynnemorrison:
Added check that linkaccess is enabled in the site before allowing token to be created, otherwise
the token is created but cannot be used and will lead to confusion by the user.

### DIFF
--- a/internal/cmd/skupper/common/utils/status.go
+++ b/internal/cmd/skupper/common/utils/status.go
@@ -21,3 +21,12 @@ func SiteReady(siteList *v2alpha1.SiteList) (bool, string) {
 	}
 	return false, ""
 }
+
+func SiteLinkAccessEnabled(siteList *v2alpha1.SiteList) (bool, string) {
+	for _, s := range siteList.Items {
+		if s.Spec.LinkAccess == "default" {
+			return true, s.Name
+		}
+	}
+	return false, ""
+}
--- a/internal/cmd/skupper/common/utils/status.go
+++ b/internal/cmd/skupper/common/utils/status.go
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"github.com/skupperproject/skupper/internal/utils/validator"
 	"github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 )
 
@@ -22,9 +23,11 @@ func SiteReady(siteList *v2alpha1.SiteList) (bool, string) {
 	return false, ""
 }
 
-func SiteLinkAccessEnabled(siteList *v2alpha1.SiteList) (bool, string) {
+func SiteLinkAccessEnabled(siteList *v2alpha1.SiteList, linkAccessTypes []string) (bool, string) {
+	linkAccessTypeValidator := validator.NewOptionValidator(linkAccessTypes)
 	for _, s := range siteList.Items {
-		if s.Spec.LinkAccess == "default" {
+		ok, _ := linkAccessTypeValidator.Evaluate(s.Spec.LinkAccess)
+		if ok {
 			return true, s.Name
 		}
 	}
--- a/internal/cmd/skupper/token/kube/token_issue.go
+++ b/internal/cmd/skupper/token/kube/token_issue.go
@@ -80,9 +80,9 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			ok, siteName = utils.SiteLinkAccessEnabled(siteList)
+			ok, siteName = utils.SiteLinkAccessEnabled(siteList, common.LinkAccessTypes)
 			if !ok {
-				validationErrors = append(validationErrors, fmt.Errorf("A site must have link access enabled before a token can be created"))
+				validationErrors = append(validationErrors, fmt.Errorf("You must enable link access for this site before you can create a token."))
 			} else {
 				cmd.grantName = siteName + "-" + uuid.New().String()
 			}
--- a/internal/cmd/skupper/token/kube/token_issue.go
+++ b/internal/cmd/skupper/token/kube/token_issue.go
@@ -80,7 +80,12 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			cmd.grantName = siteName + "-" + uuid.New().String()
+			ok, siteName = utils.SiteLinkAccessEnabled(siteList)
+			if !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("A site must have link access enabled before a token can be created"))
+			} else {
+				cmd.grantName = siteName + "-" + uuid.New().String()
+			}
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


pwright:
is there a case for #1958 statement? token is issued with warning? eg token is created now, but cannot be redeemed until site is updated.

lynnemorrison:
> is there a case for #1958 statement? token is issued with warning? eg token is created now, but cannot be redeemed until site is updated.

My feeling was that not create the token if it can't be used.    We currently don't create a token if the site is not in the ready state.  I can change this if others don't agree?

nluaces:
> > is there a case for #1958 statement? token is issued with warning? eg token is created now, but cannot be redeemed until site is updated.
> 
> My feeling was that not create the token if it can't be used. We currently don't create a token if the site is not in the ready state. I can change this if others don't agree?

I have not reviewed this PR yet, but unless there are some context that I'm not aware of; it makes sense to me what Lynne suggests. I personally think that we could help the user better if we don't allow the generation of a token that cannot be redeemed.

internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -80,7 +80,12 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			cmd.grantName = siteName + "-" + uuid.New().String()
+			ok, siteName = utils.SiteLinkAccessEnabled(siteList)
+			if !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("A site must have link access enabled before a token can be created"))
```
pwright, line 85:
```suggestion
				validationErrors = append(validationErrors, fmt.Errorf("You must enable link access for this site before you can create a token."))
```
trying to avoid passive voice

internal/cmd/skupper/token/kube/token_issue_test.go
```diff
@@ -457,6 +484,47 @@ func TestCmdTokenIssue_ValidateInput(t *testing.T) {
 			},
 			expectedError: `link cost is not valid: strconv.Atoi: parsing "Not-valid": invalid syntax`,
 		},
+		{
+			name: "linkaccess is not valid",
```
pwright, line 488:
```suggestion
			name: "link access is not valid",
```

internal/cmd/skupper/token/kube/token_issue_test.go
```diff
@@ -457,6 +484,47 @@ func TestCmdTokenIssue_ValidateInput(t *testing.T) {
...
+											Status: "True",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedError: `A site must have link access enabled before a token can be created`,
```
pwright, line 526:
```suggestion
			expectedError: `You must enable link access for this site before you can create a token.`,
```

pwright (commented):
just some suggestions re. wording

internal/cmd/skupper/token/kube/token_issue.go
nluaces: +1

internal/cmd/skupper/token/kube/token_issue_test.go
nluaces: +1

internal/cmd/skupper/token/kube/token_issue_test.go
nluaces: +1

lynnemorrison: done

internal/cmd/skupper/token/kube/token_issue_test.go
lynnemorrison: done

internal/cmd/skupper/token/kube/token_issue.go
lynnemorrison: done

pwright:
@lynnemorrison I'd love to approve this, but I think @nluaces should approve cos I'm not familiar with the code.

internal/cmd/skupper/common/utils/status.go
```diff
@@ -21,3 +21,12 @@ func SiteReady(siteList *v2alpha1.SiteList) (bool, string) {
 	}
 	return false, ""
 }
+
+func SiteLinkAccessEnabled(siteList *v2alpha1.SiteList) (bool, string) {
+	for _, s := range siteList.Items {
+		if s.Spec.LinkAccess == "default" {
```
nluaces, line 27: This is not correct. The link access type can have 3 different values: [route loadbalancer default]

internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -80,7 +80,12 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			cmd.grantName = siteName + "-" + uuid.New().String()
+			ok, siteName = utils.SiteLinkAccessEnabled(siteList)
+			if !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("You must enable link access for this site before you can create a token."))
```
nluaces, line 85: I have realized now that while the message is correct, the user must recreate the site to enable the link access; and maybe with this message they could think that the site can be modified.

@pwright do you think that we should say something like "You must have enabled link access for this site before you could create a token."?

nluaces:
This is a suggested change related to token redeem command, but I think it can be added in this pull request:

```
skupper token redeem ~/my-token.yaml -n east
Waiting for token status ...
Token "west-e66c64a0-7763-48a9-a271-b33ad7acf514" has been redeemed
You can now safely delete /home/nluaces/my-token.yaml
```

Could you delete the part `You can now safely delete...` given that the token could be potentially used more than one time if `--redemptions-allowed` is greater than 1? Thanks!

internal/cmd/skupper/token/kube/token_issue.go
lynnemorrison: I agree this is better wording, fixed

internal/cmd/skupper/common/utils/status.go
lynnemorrison: correct,  sorry missed that.

internal/cmd/skupper/token/kube/token_issue.go
pwright: @nluaces I'm not sure i understand, I've performed `skupper site  update --enable-link-access` as per https://skupperproject.github.io/refdog/commands/site/update.html - are you saying this doesn't always work?

pwright:
> Could you delete the part `You can now safely delete...` given that the token could be potentially used more than one time if `--redemptions-allowed` is greater than 1? Thanks!

It's a weird one.. If the token only has one redemption - then there's no advantage to deleting it, ie it has no value.
If the token has two redemptions, and two copies have been sent to two remote sites, we do want the first site to exercise the redemption to delete after redeeming, ie to prevent inadvertent second usage.

how about 'Consider deleting...  if it is no longer required.'?

internal/cmd/skupper/token/kube/token_issue.go
nluaces: @pwright You are completely right, it does always work, thanks for pointing out that.
@lynnemorrison I apologize, the site can be updated and for that reason the wording suggested by Paul is the correct one. Sorry for the confusion.

nluaces (approved):
Looks good to me, thank you.

PR #1967, opened 2025-02-20
Added s390x architecture to the Makefile

rajnish-jauhari:
Please review this PR. This is to ensure the skupper-tests image is build for s390x.
Also, please build the following images for s390x and push it to the quay :
iperf3
nghttp2
hey
cc: @hash-d

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ TEST_IMAGE := quay.io/skupper/skupper-tests
 TEST_BINARIES_FOLDER := ${PWD}/test/integration/bin
 DOCKER := docker
 LDFLAGS := -X github.com/skupperproject/skupper/pkg/version.Version=${VERSION}
-PLATFORMS ?= linux/amd64,linux/arm64
+PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x 
 GOOS ?= linux
 GOARCH ?= amd64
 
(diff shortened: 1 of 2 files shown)


hash-d:
@rajnish-jauhari, while I recommended making the changes to `skupper-tests` image earlier today, only now I realize that that change to `PLATFORMS` on the `Makefile` will impact all Skupper images, not only `skupper-tests`, effectively adding S390x to the list of platforms of Skupper upstream.

I think that requires a bit of project discussion.  I'll tag some people on this PR.  We may also bring this up on some meeting or on the mailing list.

@grs, @ssorj, do you guys have an opinion on this?

hash-d (changes requested):
@rajnish-jauhari, as we talked on Slack, this was discussed on the community meeting.  The decision was that it is ok to add s390x, but that it only makes sense to do that on the control-plane if the data-plane has its own image available for s390x.

Can you work on a PR to make quay.io/skupper/skupper-router:main available for s390x?  Once that's done, we can move forward with this one.

rajnish-jauhari:
Hi @hash-d , this is the PR for skupper-router:main - https://github.com/skupperproject/skupper-router/pull/1788
It is yet to be merged but once it gets merged I will let you know so that we can move forward with this one

PR #1968, opened 2025-02-20
fix several helm chart documentation errors

c-kruse:
Fixes the post-install notes on accessing the network observer. Corrects the network observer README installation instructions and other errors.

### DIFF
--- a/charts/network-observer/README.md
+++ b/charts/network-observer/README.md
@@ -21,7 +21,7 @@ observer is to be installed in.
 To deploy the Skupper Network Observer to a namespace using Helm
 
 ```
-helm install skupper-network-observer oci://quay.io/skupper/helm/network-observer
+helm install skupper-network-observer oci://quay.io/skupper/helm/network-observer --devel
 ```
 
 ## Non-Helm Usage with preconfigured manifest yaml
--- a/charts/network-observer/README.md
+++ b/charts/network-observer/README.md
@@ -18,10 +18,10 @@ observer is to be installed in.
 
 ## Usage
 
-To deploy the Skupper Network Observer to a namsapce using Helm
+To deploy the Skupper Network Observer to a namespace using Helm
 
 ```
-helm install skupper-network-observer .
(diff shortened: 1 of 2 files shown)


charts/network-observer/README.md
```diff
@@ -18,10 +18,10 @@ observer is to be installed in.
 
 ## Usage
 
-To deploy the Skupper Network Observer to a namsapce using Helm
+To deploy the Skupper Network Observer to a namespace using Helm
 
 ```
-helm install skupper-network-observer .
+helm install skupper-network-observer oci://quay.io/skupper/helm/network-observer
```
nluaces, line 24: I have tried this, but seems like it is still necessary to provide a version:

```
helm install skupper-network-observer oci://quay.io/skupper/helm/network-observer
Error: INSTALLATION FAILED: Could not locate a version matching provided version string 
```

c-kruse: That is correct. What do you think the thing to do here is?

Pre: 2.0.0 we will get this error unless we specify `--version 2.0.0-rc1` (or 0.1.0-dev - which I'd be content to change up to v2-dev now that we've agreed to allow ourselves grace in following semver for these charts)

Post 2.0.0 this command would get you the latest release.

I could see either living with a lie a while longer or hard coding a version for now and coming back to remove it later.

nluaces: Leaving the README prepared for Post 2.0.0 makes more sense than the Pre 2.0.0 option. 

But how about...
```
helm install skupper oci://quay.io/skupper/helm/skupper --devel
```
...for now?


I don't have a strong opinion about this, I think I'm just trying to save us some github issues/requests about updating READMEs! 😁

nluaces (approved):

charts/network-observer/README.md
c-kruse: :1st_place_medal: I like the --devel flag better than a hard-coded version. Thanks for finding that!

PR #1974, opened 2025-02-25
add link generate command for non-kube environments

nluaces:


internal/cmd/skupper/link/link.go
```diff
@@ -41,16 +41,16 @@ output needs to be applied in the site in which we want to create the link.`,
 
 	cmd := common.ConfigureCobraCommand(configuredPlatform, cmdLinkGenerateDesc, kubeCommand, nonKubeCommand)
 	cmdFlags := common.CommandLinkGenerateFlags{}
-	cmd.Flags().StringVar(&cmdFlags.TlsCredentials, common.FlagNameTlsCredentials, "", common.FlagDescTlsCredentials)
-	cmd.Flags().StringVar(&cmdFlags.Cost, common.FlagNameCost, "1", common.FlagDescCost)
-	cmd.Flags().StringVarP(&cmdFlags.Output, common.FlagNameOutput, "o", "yaml", common.FlagDescOutput)
-	cmd.Flags().BoolVar(&cmdFlags.GenerateCredential, common.FlagNameGenerateCredential, true, common.FlagDescGenerateCredential)
-	cmd.Flags().DurationVar(&cmdFlags.Timeout, common.FlagNameTimeout, 60*time.Second, common.FlagDescTimeout)
-
+	if configuredPlatform == common.PlatformKubernetes {
```
fgiorgetti, line 44: I believe we need two filtering arguments, when using non-kubernetes platforms.
They're both optional and if not provided, it works as is now.

But we could help filtering static links by: `name` (router access name) or `host` (endpoint host).

nluaces: done, let me know if that works for you.

internal/cmd/skupper/link/nonkube/link_generate.go
```diff
@@ -4,30 +4,113 @@ Copyright © 2024 Skupper Team <skupper@googlegroups.com>
...
+
+	siteState, err := siteStateLoader.Load()
+	if err != nil {
+		validationErrors = append(validationErrors, fmt.Errorf("there is no active site in this namespace"))
+	} else {
+
+		hasRouterAccess := false
+		if siteState.RouterAccesses != nil && len(siteState.RouterAccesses) > 0 {
+			for _, access := range siteState.RouterAccesses {
+				if strings.HasPrefix(access.Name, "router-access") {
```
fgiorgetti, line 63: The resource names are not meant to be prefixed as "router-access". The file names were prefixed
as such some time ago. But now, if you rebase on top of main, the resources are serialized through file
names prefixed with the kind, which in this case would be `RouterAccess`.
But just the filename, not the resource name.

I believe that an accurate solution would be to add a method into the SiteState type that iterates through
the `RouterAccesses`, and for each RouterAccess, it then iterates through their roles and returning true
if a RouterAccess is defined with the roles: `inter-router` or `edge`.

Similar to the existing method: `HasRouterAccess`.

internal/cmd/skupper/link/nonkube/link_generate.go
```diff
@@ -4,30 +4,113 @@ Copyright © 2024 Skupper Team <skupper@googlegroups.com>
...
+		fmt.Println("no tokens found")
+		return nil
+	}
+
+	for _, tokenFile := range tokenFiles {
+		file, errFile := os.ReadFile(hostTokenPath + "/" + tokenFile)
+		if errFile != nil {
+			return fmt.Errorf("error reading file %s: %s", hostTokenPath+"/"+tokenFile, errFile)
+		}
+		fmt.Println(string(file))
```
fgiorgetti, line 111: It should only print the first, then break.

internal/cmd/skupper/link/nonkube/link_generate.go
nluaces: Got it! thanks

internal/cmd/skupper/link/nonkube/link_generate.go
nluaces: done

fgiorgetti (approved):
LGTM! Thank you.

PR #1975, opened 2025-02-26
Adds a kind-based development cluster setup script

c-kruse:
Adds a new script ./scripts/kind-dev-cluster that tries to ensure a kind cluster is configured with a cluster-scoped skupper controller running development images. Presently attempts to pull images from the host's docker image storage (`make docker-build`) or the oci-archives directory (`make multiarch-oci`). Optionally configures metallb.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,15 @@ commands:
       - run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
       - run: sudo chmod +x kubectl
       - run: sudo mv kubectl /usr/local/bin
+  kind-install:
+    description: Install kind
+    steps:
+      - run: curl -Lo kind https://github.com/kubernetes-sigs/kind/releases/download/v0.24.0/kind-linux-amd64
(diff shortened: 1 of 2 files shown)


c-kruse:
Hia! Most of y'all probably have something similar to this. Thought I'd clean mine up and solicit suggestions. Sounds like we're going to need something like this for CI sooner than later. @grs @fgiorgetti @nluaces @ajssmith.

c-kruse:
@rafaelvzago interested to hear from you how this behaves with the e2e test suite.

One place I felt they "chafed" was with linking up kubeconfigs from this script with the hard coded $HOME/.kube/config values in the e2e suites host_vars. Some of my personal preferences leaked in to this script [here](https://github.com/skupperproject/skupper/pull/1975/files#diff-a9d45e280509b03ea701dbf2ad2106eeb1ddf84d7ab45b22dd6463da5191061cR208-R210) where I prefer creating a separate kubeconfig files unless KUBECONFIG is not explicitly set (I tend to reserve ~/.kube/config for other purposes.) If I am alone in working this way I'm glad to default to ~/.kube/config here, but I suppose that only really solves for the use case of a single-cluster test.

nluaces (approved):

Issue #1930, opened 2025-01-28
Editting a Listener and changing the spec.host field creates a new service, but does not remove the old one


granzoto:
**Describe the bug**
Create a listener using spec.host equals to backend-ha, then edits the listener and changes its value to backend.
Check that a new service named backend is created, but the old one, named backend-ha was not removed

**How To Reproduce**
Create two skupper sites
Create an accessgrant and an accesstoken, establish the connection
Deploy a workload ( backend ) in one site
Create the proper connector and listener for the service/workload

From the listener definition :
`spec:`
`  host: backend-ha`

services Created:
`NAME                   TYPE           CLUSTER-IP       EXTERNAL-IP      PORT(S) `
`backend-ha        ClusterIP    X.X.X.X                 <none>           8080/TCP  `

Edit the listener and change the spec.host field.

This should start the creation of a new service, and also remove the old service

New host address in the listener definition :
`spec:`
`  host: backend`

The new service is created as expected, but the old one remains untouched
`NAME                   TYPE           CLUSTER-IP       EXTERNAL-IP      PORT(S)`
`backend              ClusterIP      X.X.X.X              <none>           8080/TCP`
`backend-ha        ClusterIP      Y.Y.Y.Y               <none>           8080/TCP`

**Expected behavior**
The new service should be created and the old one should be removed

**Environment details**
* Skupper Version : v2-dev
* Platform : Minikube

lynnemorrison:
Should you be using listener update command vs listener create again to make a change?

PR #1976, opened 2025-02-26
delete redundant service when listener host is changed

grs:
fixes issue #1930


grs:
Note only the second commit is of relevance to this PR. However I based it on the work for a previous PR #1972 as it touches some of the same lines changed in that PR. Once that PR is merged, I will rebase this one and it will only contain the second commit. Pushing early to allow for more time for review if desired.

fgiorgetti (approved):

Issue #1979, opened 2025-02-26
[v2] CRD Model+Spec Inconsistencies


c-kruse:
**Describe the bug**
I've found a few inconsistencies between ./pkg/apis **model** and the CRDs as written in ./config/crd **spec**. It's not totally clear to me which classes of these are important to get right, and which as-is.

In general, these fall into one of two categoreis: properties in the model that are not in the spec and inconsistencies between a spec field being "required" vs the model having a json:"omitempty" tag.

- [ ] AttachedConnectorSpec.UseClientCert defined in model is not included in the spec.
- [ ] AttachedConnectorSpec.Selector is indicated to be required in the spec but has omitempty in the model (indicates not required?)
- [ ] LinkSpec.Endpoints is indicated to be required in the spec but has omitempty in the model.
- [ ] RouterAccessSpec.Options is defined in the model but is not included in the spec.
- [ ] RouterAccessRole.Name, RouterAccessRole.Port are indicated to be required in the model but not in the spec.
- [ ] RouterAccessSpec.GenerateTLSCredentials and Issuer are indicated to be required in the model but not in the spec.
- [ ] SecuredAccessSpec.Options is defined in the model but not in the spec.
- [ ] SecuredAccessPort.Name and Port are indicated to be required in the model but not in the spec.

Other types of inconsistencies:
- [ ] The pattern used in the spec for status.conditions.type appears inconsistent with the k8s meta/v1.Condition.Type validation spec. I think there is an errant space in character class 8. Not sure how/if this could actually manifest a problem.
```
ours: pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][- A-Za-z0-9_.]*)?[A-Za-z0-9])$
kube: pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$
```
- [ ] AccessGrantStatus.Url Code and Ca indicated required in model, not in spec.
- [ ] Status.Message Many/all Resource Status models embed v2alpha1.Status which as a message field missing from many crd specs.
- [ ] (ConnectorStatus|AttachedConnectorStatus).PodDetails.Name and Ip model indicates these are required fields but the spec does not.
[...]

PR #1981, opened 2025-02-27
align struct definitions and crds

grs:
Fixes for the issues listed in #1979. However, to clarify:

> Many/all Resource Status models embed v2alpha1.Status which as a message field missing from many crd specs.

I only found AttachedConnector and AttachedConnectorBinding to be missing this field. This PR adds it to both of those. Did I miss any others?

[Also, a somewhat philosophical (not to mention pedantic!) note. I personally feel that the omitempty is not an ideal proxy for whether the field is required. If omitempty is specified then clearly the field is optional. I would argue the reverse is not necessarily true, though, i.e. the fact that a field is not marked as omitempty doesn't necessarily mean it is required. To me it just means its important enough that showing an empty value would be useful (e.g. url and ca in AccessGrantStatus). This is not really a practical concern however, since these resources will be viewed by the user through the API server rather than what is generated from the golang structs. For non-kube platforms the situation is slightly different of course, and the output from the types might indeed be viewed by users, but there is probably also less need for the 'important' fields not to be populated there.  Anyway, I have I think marked all the optional fields with omitempty.]

### DIFF
--- a/config/crd/bases/skupper_access_grant_crd.yaml
+++ b/config/crd/bases/skupper_access_grant_crd.yaml
@@ -75,7 +75,7 @@ spec:
                         type: string
                       type:
                         maxLength: 316
-                        pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][- A-Za-z0-9_.]*)?[A-Za-z0-9])$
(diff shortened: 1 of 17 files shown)


c-kruse:
Thank you @grs!

> I only found AttachedConnector and AttachedConnectorBinding to be missing this field. This PR adds it to both of those. Did I miss any others?

Double checking now I also see just those two. I must have just been hasty to generalize here.

> Also, a somewhat philosophical (not to mention pedantic!) note

I agree entirely! Just wanted to be sure to raise the issue as I was not certain I understood fully what the potential ramifications would be.

fgiorgetti (approved):
It looks good to me.
Validated against e2e tests and manual CLI execution.

Issue #1969, opened 2025-02-20
[v2] Produce Network Observer Static Manifest YAML with Openshift Auth


c-kruse:
The skupper network-observer helm chart supports OpenShift authentication by way of introducing the [oauth-proxy] container in front of the observer deployment. Presently this is accomplished by generating the required cookie secret using a helm template on install here: https://github.com/skupperproject/skupper/blob/b872a1a06cb3ebada8a414baa4f4ad9aa806e600/charts/network-observer/templates/_deployment.yaml#L57

In order to produce a secure static yaml manifest file with this configuration enabled, it would require a new Job and associated RBAC resources in addition to the regular network observer deployment resources.

The purpose of this new Job would be only to generate a suitable [cookie secret] for the [oauth-proxy] to use.

## Proposed solution
* Add a new subcommand to either the network-observer, the skupper CLI or even the kube-adaptor that will ensure a suitable cookie-secret is present in a kubernetes secret. https://github.com/skupperproject/skupper/pull/1987
* Add a new Job resource (as a helm pre-install hook?) and associated RBAC resources to the network-observer chart that invokes that new command to generate the secret.
* Update the chart to mount that secret to the proxy container and update the proxy command to pull that secret from a file instead of *argv.

[oauth-proxy]: https://github.com/openshift/oauth-proxy
[cookie secret]: https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview/#generating-a-cookie-secret

c-kruse:
Finished!

PR #1991, opened 2025-03-05
Automate secret generation in network-observer chart.

c-kruse:
Updates the skupper network-observer Chart with an optional Job to provision secrets to facilitate either of the auth options.

When auth.strategy=openshift and auth.openshift.createCookieSecret are set, includes a pre-install Job to generate an appropriate oauth2 session cookie secret and store it in a secret for the oauth proxy to use for signing session cookies.

When auth.strategy=basic and auth.basic.create are set, includes a pre-install Job to generate a htpasswd file containing the user "skupper" and a random password in plain-text for a user to retrieve from the secret.

Finishes https://github.com/skupperproject/skupper/issues/1969


c-kruse:
Depends on https://github.com/skupperproject/skupper/pull/1988

ajssmith (approved):
lgtm

PR #1993, opened 2025-03-10
Comments and coverage with some refactoring

grs:


c-kruse (approved):
Have been over this a few times now. Let's go ahead with it!

PR #2001, opened 2025-03-12
Add README to skupper chart

c-kruse:
Adds a README.md file to the generated skupper helm chart explaining the purpose and usage of the chart. Including this inside of the chart directory allows it to integrate better with the helm tooling that expects to find documentation packaged along with the chart.

### DIFF
--- a/charts/skupper/README.md
+++ b/charts/skupper/README.md
@@ -37,9 +37,13 @@ the `--skip-crds` flag with `helm install`.
 ### Image Overrides
 
 The chart exposes overrides for the three images required to run a skupper site.
-* `controllerImage`
-* `kubeAdaptorImage`
-* `routerImage`
+
+Example values.yaml file:
+```
+controllerImage:    examplemirror.acme.com/skupper/controller:2.0.0
+kubeAdaptorImage:   examplemirror.acme.com/skupper/kube-adaptor:2.0.0
+routerImage:        examplemirror.acme.com/skupper/skupper-router:3.3.0
+```
 
 ## Alternative Installation Methods
 
--- a/charts/skupper/README.md
+++ b/charts/skupper/README.md
@@ -1,0 +1,63 @@
+# Skupper
+
+Skupper is a layer 7 service interconnect. It enables secure communications
+across Kubernetes clusters with no VPNs or special firewall rules.
+
+This chart installs the [Skupper](https://skupper.io) version 2 controller for
+[Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh)
+package manager.
+
+
+## Prerequisites
+
+- Kubernetes 1.25+
+- Helm 3
+
+## Usage
+
+A cluster-scoped skupper controller can be deployed with
+```
+helm install skupper oci://quay.io/skupper/helm/skupper \
+    --namespace skupper \
+    --create-namespace
+```
+
+To install the controller at a namespace-scope use `--set scope=namespace`
+```
+helm install skupper oci://quay.io/skupper/helm/skupper \
+    --set scope=namespace
+```
+
+### CRDs
+
+By default, the chart will install the Skupper CRDs required for the controller
+to properly function. To install CRDs independently from the Helm chart, use
+the `--skip-crds` flag with `helm install`.
+
+### Image Overrides
+
+The chart exposes overrides for the three images required to run a skupper site.
+* `controllerImage`
+* `kubeAdaptorImage`
+* `routerImage`
+
+## Alternative Installation Methods
+
+In addition to this Helm Chart, Skupper releases static manifest yamls for
+deploying both cluster and namespace-scoped controllers.
+
+```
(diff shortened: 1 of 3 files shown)


charts/skupper/README.md
```diff
@@ -0,0 +1,63 @@
...
+SKUPPER_VERSION=2.0.0
+# Deploys a cluster scoped controller to the 'skupper' namespace.
+kubectl apply -f "https://github.com/skupperproject/skupper/releases/download/$SKUPPER_VERSION/skupper-cluster-scope.yaml"
+# Deploys a namespace scoped controller to the current context namespace.
+kubectl apply -f "https://github.com/skupperproject/skupper/releases/download/$SKUPPER_VERSION/skupper-namespace-scope.yaml"
+```
+
+## Development
+
+This Helm chart is generated from the Makefile at root of the [skupper
```
nluaces, line 59: Maybe we could make a reference to the README file in the /charts directory instead.

charts/skupper/README.md
```diff
@@ -0,0 +1,63 @@
...
+## Prerequisites
+
+- Kubernetes 1.25+
+- Helm 3
+
+## Usage
+
+A cluster-scoped skupper controller can be deployed with
+```
+helm install skupper oci://quay.io/skupper/helm/skupper \
```
nluaces, line 20: Probably I'm missing something because I'm a little bit confused; 1) I don't understand why we recommend to create a specific namespace named `skupper` for this scope (it seems to me that we need that specific name in order to work in a cluster namespace), 2) when we thought before that the most common view for a user was to install the controller in the current namespace. 

The creation of the `skupper` namespace was associated with the restriction in the output of the kustomize files that need a specific namespace for the installation of  a service account and it was chosen "skupper" at that time in order to start working. 

I also remember that in previous iterations of the helm chart, one of the suggested changes was to allow the user to specify their own name for the controller namespace, given that at that time forcing the user to use `skupper` didn't make too much sense.

This is not a big issue, I just want to understand the change of opinion and therefore, this suggestion in the README file.

nluaces: I do think that it depends on the first experience of the user: 
- A user that starts using the helm chart, could find strange that the controller is installed in a skupper namespace.
- A user that starts using the yaml files, they would miss that the controller is not installed in skupper, given the fact that the yaml files always deploy the controller there and they are used to it.

If we recommend `skupper` a new Helm user would think that we need that specific name which is not correct.

pwright: @nluaces 
would a line after this block like the following help?

To install the controller in a different namespace use `--namespace <custom-ns> -- create-namespace`.

nluaces: It could be also: 
```
A cluster-scoped skupper controller can be deployed with:

helm install skupper oci://quay.io/skupper/helm/skupper

Or in case you want the skupper cluster to be deployed in a specific namespace instead of your current namespace:

helm install skupper oci://quay.io/skupper/helm/skupper \
    --namespace <custom-ns> \
    --create-namespace
    
```

I can understand adding opinionated settings in a example repository, but I'm not fully convinced to have them in a README file that explains how to use the helm chart (that is going to be published in ArtifactHub as well).

charts/skupper/README.md
```diff
@@ -0,0 +1,63 @@
...
+the `--skip-crds` flag with `helm install`.
+
+### Image Overrides
+
+The chart exposes overrides for the three images required to run a skupper site.
+* `controllerImage`
+* `kubeAdaptorImage`
+* `routerImage`
+
+## Alternative Installation Methods
```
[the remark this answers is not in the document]
c-kruse: @pwright I think it would be! I am just not sure _where_ to link to. AFAIK that is still a little bit in flux. I don't see a place where we document these manifests, and it's not clear if upstream we plan on pointing people at github release artifacts directly or at some fancy `skupper.io/v2/install/cluster.yaml` type shortcut I've seen @ssorj playing around with.

charts/skupper/README.md
```diff
@@ -0,0 +1,63 @@
...
+to properly function. To install CRDs independently from the Helm chart, use
+the `--skip-crds` flag with `helm install`.
+
+### Image Overrides
+
+The chart exposes overrides for the three images required to run a skupper site.
+* `controllerImage`
+* `kubeAdaptorImage`
+* `routerImage`
+
```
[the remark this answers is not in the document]
c-kruse: How's that look now? I think any Helm user would be able to look at that and readily understand how they could apply that in a way that covers their needs.

charts/skupper/README.md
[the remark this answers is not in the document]
pwright: fair enough

charts/skupper/README.md
c-kruse: I think that's a good spot to land! Thank you both!

nluaces (approved):

PR #2002, opened 2025-03-13
Fixes EventProcessor requeue limit

c-kruse:


grs:
Yes! I think that is correct, and that is why it is not limiting the errors correctly. Thank you!

c-kruse:
@grs rebased on your latest PR so that it could be tidy enough to add a unit test!

grs (approved):

grs:
Very nice, thanks for this!

PR #2015, opened 2025-03-17
version command improvement

nluaces:
- get image SHA concurrently 
- only pull the images that are not existent locally

### DIFF
--- a/internal/images/image_utils.go
+++ b/internal/images/image_utils.go
@@ -1,6 +1,7 @@
 package images
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/skupperproject/skupper/api/types"
 	corev1 "k8s.io/api/core/v1"
@@ -152,7 +153,11 @@ func GetPrometheusImageRegistry() string {
 }
 
 func GetSha(imageName string) string {
-	checkCmd := exec.Command("docker", "inspect", imageName)
+	var output bytes.Buffer
+	checkCmd := exec.Command("docker", "inspect", "--format={{index .RepoDigests 0}}", imageName)
+	checkCmd.Stdout = &output
+	checkCmd.Stderr = &output
+
 	if err := checkCmd.Run(); err != nil {
 		// Pull the image only if it's not present locally
 		pullCmd := exec.Command("docker", "pull", imageName)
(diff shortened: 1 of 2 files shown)


grs (approved):
I might have just done the GetSha concurrently, after having constructed the rest of the data serially, as that way the starting of the goroutines is only required if enableSha is true.

However that is personal preference and what is there is accurate, clear and an improvement. :shipit:

nluaces:
@grs Thanks! I've changed it because I think that your suggestion is better.

grs (approved):

internal/images/image_utils.go
```diff
@@ -151,29 +152,63 @@ func GetPrometheusImageRegistry() string {
...
+
+	for _, image := range imageList {
+
+		waitGroup.Add(1)
+
+		go func(skImage SkupperImage) {
+			defer waitGroup.Done()
+
+			var output bytes.Buffer
+			checkCmd := exec.Command("docker", "inspect", "--format={{index .RepoDigests 0}}", skImage.Name)
```
fgiorgetti, line 168: What do you think of running a `docker pull` before the `inspect`? 
This might be useful when using tags that are constantly overwritten like `v2-dev` or `main`.

fgiorgetti: What about doing it only if "tag" does not start with a number?
This way, when dealing with a particular release, this won't be done all the time.

Maybe not needed at all, but the reason why it might be useful to me would be when someone
is using v2-dev (for example) as the local image version might not have the same digest of the
image actually in use in the cluster.

fgiorgetti (approved):

internal/images/image_utils.go
nluaces: I was thinking something like having an array of always-pull-this-tags = [main, v2-dev]

what do you think?

fgiorgetti: It works for me.

grs: To some extent this code is trying to do two very different things. One is to determine what image to use when configuring a site. The other is to report what is actually in use. I wonder f actually we need to be clearer about those distinct goals, including within the version command itself.

fgiorgetti: That is a good point.

These images are no longer used by the CLI to determine what a given Kubernetes site will use, right?
But of course, it could also report which images a given site (namespace) is using and the respective image
of the owning controller instance (to be evaluated).

For non-kubernetes sites, then the images in use by the CLI (from either ENV vars or the compiled ones),
might differ from what is currently active on a given namespace.

Maybe we need a separate tracker or discussion to address this issue?

nluaces: > Maybe we need a separate tracker or discussion to address this issue?
Done: #2019 
Feel free to add more comments to it.

PR #2016, opened 2025-03-19
handle security policies

bartoval:


c-kruse (approved):
Thank you for sorting this out!

Looks to be working w/ the latest console in a number of configurations I checked.

CI mad about whitespace or something. Running `go fmt ./cmd/network-observer`  should fix whatever it is on about.

PR #2027, opened 2025-03-24
Removes helm charts from GitHub Release Artifacts

c-kruse:
Excludes the helm chart archives from the GitHub Release Artifacts. These charts are published as OCI artifacts to quay.io along with our container images. If in an uncommon case a user needed a Chart as an archive they could download it from quay instead of GH.

Ex: 'helm pull oci://quay.io/skupper/helm/skupper --version 2.0.0'

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -356,10 +356,8 @@ jobs:
       - run: cp ./*.yaml skupper-setup
       - run: helm registry login -u ${QUAY_LOGIN} -p ${QUAY_PASSWORD} quay.io
       - run: make pack-skupper-helm-chart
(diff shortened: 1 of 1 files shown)


c-kruse:
@nluaces would appreciate your take on this one! Glad to try and shove them in a "charts.tar.gz" archive or something if you've got a good reason to keep these handy.

I know @ajssmith has been using the network-observer one downstream.

nluaces:
> Glad to try and shove them in a "charts.tar.gz" archive or something if you've got a good reason to keep these handy.

Probably it is just fine to indicate in the release notes how to install or pull the helm chart with a specific version like you have suggested.

ajssmith (approved):

PR #2037, opened 2025-03-26
remove non-essential short options

### LINKED ISSUE #2012: Remove short options for less common cases
Short options are precious and should be reserved for frequent or highly conventional cases.  We don't know right away which options deserve short options, so it's important not to use them up at the start.

In general, remove any short options not present in the CLI spec: https://skupperproject.github.io/refdog/commands/index.html

Right now, these are the only short options we have in the spec:

```
-n for --namespace
-h for --help
-o for --output (on version, resource status, and resource generate commands)
-f for --filename (on system apply and system delete)
```

nluaces:
Fixes #2012


ajssmith (approved):
lgtm

PR #2039, opened 2025-03-26
fix CLI handling of annotations

### LINKED ISSUE #1544: Skupper cli flag handling for annotations
It can be necessary to use any of the --annotations flags to add annotations that contain commas themselves.

For example if I wanted the annotation "foo=x,y,z" to be applied to my resources I would want skupper init to accept the  `--annotations=foo=x,y,z` flag. Due to the confusing combination of bash string handling, cli flag parsing in spf13/pflag and our implementation I have not sorted out how to get this to work.

lynnemorrison:
Add support for different annotations decoding:

The current implementation separates annotations based on commas and equal signs so an annotation like:
foo=x,y,z   would result in an incorrect annotation like:
    annotations:
      x: ""
      y: ""
      foo: z

or 

test2="value=abc" would result in an incorrect annotation like:
  annotations:
    test2: value
    abc: ""

This fix adds an annotation-separator flag that the user can specify to separate annotations and default to a comma if annotation-separator flag is not specified.  This default allows backwards compatibility with existing code.

With this fix the annotations will now correctly be decoded as:

skupper init --router-pod-annotations **service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags=Environment=dev@skupper-separator@foo=x,y,z** --annotation-separator **@skupper-separator@**

 kubectl get deployment skupper-router -o yaml

 template:
    metadata:
      annotations:
        **foo: a,b,c**
        prometheus.io/port: "9090"
        prometheus.io/scrape: "true"
        **service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: Environment=dev**

fixes #1544

### DIFF
--- a/pkg/site/site.go
+++ b/pkg/site/site.go
@@ -27,6 +27,7 @@ const (
 	SiteConfigRunAsUserKey           string = "run-as-user"
 	SiteConfigRunAsGroupKey          string = "run-as-group"
 	SiteConfigClusterPermissionsKey  string = "cluster-permissions"
+	SiteConfigAnnotationSeparator    string = "annotation-separator"
 
 	// console options
 	SiteConfigConsoleKey               string = "console"
@@ -132,6 +133,7 @@ func WriteSiteConfig(spec types.SiteConfigSpec, namespace string) (*corev1.Confi
 			SiteConfigConsoleUserKey:           "",
 			SiteConfigConsolePasswordKey:       "",
 			SiteConfigIngressKey:               types.IngressLoadBalancerString,
+			SiteConfigAnnotationSeparator:      ",",
 		},
 	}
 	if spec.SkupperName != "" {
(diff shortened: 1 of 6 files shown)


pkg/site/site.go
```diff
@@ -475,6 +480,7 @@ func ReadSiteConfig(siteConfig *corev1.ConfigMap, namespace string, defaultIngre
 	var errs []string
 	var result types.SiteConfig
 	result.Spec.SkupperNamespace = siteConfig.Namespace
+	annotationSeparator, _ := siteConfig.Data[SiteConfigAnnotationSeparator]
 	// TODO: what should the defaults be for name, namespace
```
grs, line 484: Should a default value be applied here is none is set explicitly?

lynnemorrison: done

pkg/site/site.go
```diff
@@ -475,6 +480,11 @@ func ReadSiteConfig(siteConfig *corev1.ConfigMap, namespace string, defaultIngre
 	var errs []string
 	var result types.SiteConfig
 	result.Spec.SkupperNamespace = siteConfig.Namespace
+
+	annotationSeparator := ","
+	if _, ok := siteConfig.Data[SiteConfigAnnotationSeparator]; ok {
```
grs, line 485: Minor nitpick: looks odd to ignore the value and then re-retrieve it

grs (approved):

c-kruse (approved):
Thank you!

pkg/site/site.go
lynnemorrison: fixed

PR #2040, opened 2025-03-31
[kube] Check if credentials already exist before generating a link

### LINKED ISSUE #2005: Link generation fails with name conflict
**Describe the bug**

```
$ skupper link generate
certificates.skupper.io "link-west" already exists
-> 1
```

Generate commands don't create resources.  How can there be a conflict with something already existing?

Also, generated resources should have unique qualifiers: link-west-ac619.

**How To Reproduce**

Run "skupper link generate" twice.

**Environment details**

Skupper fb05630a (main, 14 March 2025)

nluaces:
- Fixes #2005
- Avoids printing the yaml separator (`---`) after printing the last resource, in order to make it compatible with the non-kube resource parser.

### DIFF
--- a/internal/cmd/skupper/link/link.go
+++ b/internal/cmd/skupper/link/link.go
@@ -34,8 +34,8 @@ func CmdLinkGenerateFactory(configuredPlatform common.Platform) *cobra.Command {
 
 	cmdLinkGenerateDesc := common.SkupperCmdDescription{
 		Use:   "generate",
-		Short: "Generate a new link resource and save it as a YAML file, unless explicitly specified otherwise using the `--output` flag.",
-		Long: `Generate a new link resource and save it as a YAML file, unless explicitly specified otherwise using the --output flag. The resultant
+		Short: "Generate a new link resource as a YAML output, unless explicitly specified otherwise using the `--output` flag.",
+		Long: `Generate a new link resource as a YAML output, unless explicitly specified otherwise using the --output flag. The resultant
 output needs to be applied in the site in which we want to create the link.`,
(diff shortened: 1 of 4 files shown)


internal/cmd/skupper/link/link.go
```diff
@@ -34,16 +34,16 @@ func CmdLinkGenerateFactory(configuredPlatform common.Platform) *cobra.Command {
 
 	cmdLinkGenerateDesc := common.SkupperCmdDescription{
 		Use:   "generate",
-		Short: "Generate a new link resource in a yaml file",
-		Long: `Generate a new link resource with the data needed from the target site. The resultant
+		Short: "Generate a new link resource and save it as a YAML file, unless explicitly specified otherwise using the `--output` flag.",
```
fgiorgetti, line 37: It is not behaving as the description says. It is always displaying a YAML output.

nluaces: "unless explicitly specified otherwise using the `--output` flag"?

Did not work with `-o json`?

fgiorgetti: Yes, it works with json, but the description says: `Generate a new link resource and save it as a YAML file`.
Is it still a valid description?

nluaces: My bad, you are right, it is not generating any file. Will change that now.

nluaces: I have updated the description again, given that the -o flag is not available for non-kube.

fgiorgetti (approved):

PR #2043, opened 2025-04-03
check if docker is available in order to get the digest in the version command

### LINKED ISSUE #2029: Make docker optional for the version command
At present, the docker executable is needed in the CLI to get the image digest.

If there is no docker installed, skip the digest images search.

nluaces:
Fixes #2029

### DIFF
--- a/internal/utils/configs/manifest.go
+++ b/internal/utils/configs/manifest.go
@@ -51,16 +51,23 @@ func getSkupperImages(components []string, enableSHA bool, runningPods map[strin
 	_, err := exec.LookPath("docker")
 	if err != nil {
 		enableSHA = false
+		os.Stderr.WriteString("Warning: Docker is not installed. Skipping image digests search.\n")
+	}
+
+	//This map contains all the digests of the skupper images concurrently
+	var imageDigestMap map[string]string
+	if enableSHA {
+		imageDigestMap = images.CreateMapImageDigest(runningPods)
 	}
 
 	for _, component := range components {
 		var image SkupperComponent
 		image.Component = component
 		image.Version = images.GetImageVersion(component)
-		image.Images = images.GetImages(component, enableSHA)
+		image.Images = images.GetImages(component, imageDigestMap)
 		if runningPods[component] != "" {
 			image.Version = images.GetVersionFromTag(runningPods[component])
-			image.Images = GetRunningImages(component, enableSHA, runningPods)
+			image.Images = GetRunningImages(component, runningPods, imageDigestMap)
 		}
 
 		manifest.Components = append(manifest.Components, image)
@@ -76,7 +83,7 @@ func getSkupperDefaultImages() SkupperManifest {
 		var image SkupperComponent
 		image.Component = component
 		image.Version = images.GetImageVersion(component)
-		image.Images = images.GetImages(component, false)
+		image.Images = images.GetImages(component, nil)
 		manifest.Components = append(manifest.Components, image)
 	}
 
@@ -140,7 +147,7 @@ func getEnvironmentVariableMap() *map[string]string {
 	return &envVariables
 }
 
-func GetRunningImages(component string, enableSHA bool, runningPods map[string]string) []images.SkupperImage {
+func GetRunningImages(component string, runningPods map[string]string, digestMap map[string]string) []images.SkupperImage {
 
 	names := make(map[string]string)
 
@@ -159,7 +166,7 @@ func GetRunningImages(component string, enableSHA bool, runningPods map[string]s
 	}
 
 	if names != nil {
-		return images.GetImage(names, "", enableSHA)
+		return images.GetImage(names, "", digestMap)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


internal/utils/configs/manifest.go
```diff
@@ -46,6 +47,12 @@ func (manager *ManifestManager) GetDefaultManifestWithEnv() Manifest {
 func getSkupperImages(components []string, enableSHA bool, runningPods map[string]string) SkupperManifest {
 	var manifest SkupperManifest
 
+	// if Docker is not installed we can not inspect digests from images.
+	_, err := exec.LookPath("docker")
+	if err != nil {
+		enableSHA = false
```
fgiorgetti, line 53: I am wondering if it should not print a warning to stderr, in case docker is not available?
Thoughts?

hash-d (commented):
I don't think the problem on 2.0 is whether the `docker` command is available, but instead what to do in case of errors executing it.

For example, in the case below the command was available, but I have forced it to fail on a bad image:

```
$ SKUPPER_CLI_IMAGE=quay.io/skupper/asdf:1234 skupper version -o json  | jq
jq: parse error: Invalid numeric literal at line 1, column 6
```

The resulting JSON was invalid.  The problem was that the command errors were part of the output:

```
$ SKUPPER_CLI_IMAGE=quay.io/skupper/asdf:1234 skupper version -o json  2> /dev/null | head -3
Error pulling image: exit status 1Error pulling image: exit status 1Error pulling image: exit status 1Error pulling image: exit status 1Error pulling image: exit status 1Error pulling image: exit status 1Error pulling image: exit status 1{
  "components": [
    {
```

That, however, was already fixed on e1acf8d79.

Now, the errors are being shown as the digest:

```
$ skupper version -o json 2> /dev/null | head
2025/04/05 10:25:49 Error pulling image: exit status 1
2025/04/05 10:25:49 Error pulling image: exit status 1
2025/04/05 10:25:49 Error getting image digest: exit status 1
(...)
2025/04/05 10:25:50 Error pulling image: exit status 1
2025/04/05 10:25:50 Error getting image digest: exit status 1
{
  "components": [
    {
      "component": "router",
      "images": [
        {
          "digest": "Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?",
          "name": "quay.io/skupper/kube-adaptor:v2-dev"
        },
        {
```
The JSON is now valid, but its contents may not be.

My suggestion:

- Remove the check for the `docker` binary on `getSkupperImages`; let that condition be treated the same way as errors coming from docker would
- On `AddShaToImages`, change the `log.Printf` lines to include the actual error from stderr
- In case of errors, either leave the `digest` key empty, or not include it at all

This way, we get proper reporting on what failed, and the output remains semantically valid (ie, avoid the case where someone tries to use the digest to download an image, only to find it is actually an error message).

That said, I'd like to take a step back: why do we include the digests in the output in the first place?

That is information that is not baked in the binary; we're using `docker` to derive it at runtime from the tag.

The tag is the thing that is baked in, and that we need to report, and it is already present on the output JSON.

It should be simple for a user of that JSON to use whatever tool they prefer to get the digests from the tags.

I think we should consider removing the digests altogether: it adds complexity, depends too much on the environment, is prone to fail and adds little value; value which is easily achieved by external tools consuming the rest of the output information.

nluaces:
> That said, I'd like to take a step back: why do we include the digests in the output in the first place?
> I think we should consider removing the digests altogether: it adds complexity, depends too much on the environment, is prone to fail and adds little value; value which is easily achieved by external tools consuming the rest of the output information.

The main purpose of the digest was to show in the release the exactly images that were used by default in skupper. This function that now is included in the version command, was originally only thought to generate the manifest.json file in the release.

In the past, users wanted to know the exactly digest of the images that were configured by default with a specific skupper version, this was also requested by the router team.

internal/utils/configs/manifest.go
nluaces: done

nluaces:
@hash-d it was fixed to show an empty digest value in case of error. Regarding your suggestions about the verbose output, I think you need to create first an issue in the Refdog project with that idea:
This is the current expectation: https://skupperproject.github.io/refdog/commands/version.html

Once reflected in Refdog, we can proceed with the changes.

hash-d:
@nluaces, I have created https://github.com/skupperproject/refdog/issues/7 for this.

> The main purpose of the digest was to show in the release the exactly images that were used by default in skupper. This function that now is included in the version command, was originally only thought to generate the manifest.json file in the release.

I'm not sure that objective is fulfilled by the current behavior.  The image that will be used by default is the one with the tag, not necessarily the retrieved SHA.  If the user put that SHA on some gitops repo, that information would get stale once the tag is set to a new version (the router component, for example, used to have 3.2.0 pointing to 3.2.0-1, but it is now 3.2.0-2).

> In the past, users wanted to know the exactly digest of the images that were configured by default with a specific skupper version, this was also requested by the router team.

As I see it, this does not exist; as we do not hardcode digests on the binaries (only tags we do), we cannot tell them what image shas are configured by default with a specific skupper version.

On the refdog issue, I added an alternative to make digest retrieval optional, and explicitly an online operation.

Issue #2050, opened 2025-04-07
Reduce control plane image sizes


ssorj:
Use the technique demonstrated here: https://github.com/skupperproject/skupper-router/pull/1728

I think this is what we want for the cli, controller, network-observer, and kube-adaptor images.

The benefits are faster setup times for users and fewer CVEs to address.

PR #2052, opened 2025-04-07
build the skupper images from scratch

nluaces:
Part of #2050

### DIFF
--- a/Dockerfile.config-sync
+++ b/Dockerfile.config-sync
@@ -12,11 +12,19 @@ RUN go mod download
 COPY . .
 RUN make GOOS=$TARGETOS GOARCH=$TARGETARCH
 
-FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9-minimal
+FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9:latest AS packager
 
 # Create user and group and switch to user's context
-RUN microdnf -y install shadow-utils \
-&& microdnf clean all
+RUN dnf -y --setopt=install_weak_deps=0 --nodocs \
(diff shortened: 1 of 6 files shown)


Dockerfile.config-sync
```diff
@@ -12,11 +12,15 @@ RUN go mod download
...
 
-FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9-minimal
+FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9:latest AS packager
 
 # Create user and group and switch to user's context
-RUN microdnf -y install shadow-utils \
-&& microdnf clean all
+RUN dnf -y --setopt=install_weak_deps=0 --nodocs --installroot /output install coreutils-single shadow-utils && dnf clean all --installroot /output
+
+FROM scratch
```
fgiorgetti, line 20: I believe that the useradd should not be done as part of the scratch layer.
Looking at the router image, I see it being done in the previous layer as:
```
chroot /output useradd --uid 10000 runner
```

fgiorgetti: If we do that earlier can we get rid of some of the other deps above installed under /output?

nluaces: This solution as it is is not valid for our project, probably it is needed to find a detailed solution that fits our current docker files, in line with what you are suggesting.

c-kruse: @nluaces - for posterity and because I'm curious: what about this doesn't work?

I see that the executable we build _do_ have a libc dependency (built with CGO), but I can't really imagine what else. Is it that we expect to exec into containers and have bash and other dependencies?

PR #2053, opened 2025-04-07
Readme updates

pwright:


cmd/controller/README.md
```diff
@@ -0,0 +1,13 @@
+# Skupper controller
+
+Skupper releases static manifest yamls for deploying both cluster and namespace-scoped controllers.
+
+```
+SKUPPER_VERSION=2.0.0
```
nluaces, line 6: aren't we interested in showing how to generate the yaml files that use the main controller version?

pwright: that makes more sense than hardcoding 2.0.0 - I just copied and pasted this, but does following make sense?
```
SKUPPER_VERSION=main
```
Replace `main` with release if required, eg `2.0.0`

nluaces: In this case, it would be `v2-dev-release` instead of `main`[1]

https://github.com/skupperproject/skupper/releases/download/v2-dev-release/skupper-cluster-scope.yaml
https://github.com/skupperproject/skupper/releases/download/v2-dev-release/skupper-namespace-scope.yaml

source: https://github.com/skupperproject/skupper/releases/tag/v2-dev-release

Alternatively, the yaml files can be generated by the user in the local cloned repo by running:
```
make generate-skupper-deployment-cluster-scoped
make generate-skupper-deployment-namespace-scoped
```

But if you prefer to not add those commands it's fine by me.


[1] Why? because the former main release contains the v1 version.

nluaces (approved):

PR #2062, opened 2025-04-09
Add output flag support for site command

lynnemorrison:
The site status command was missing the output flag, that allowed user to display site resource as yaml or json.
Also added "Sites in Network" parameter, which is currently not defined in refDog.


nluaces:
> Also added "Sites in Network" parameter, which is currently not defined in refDog.

If I'm not wrong, I think we don't need to add to the skupper site status all the sections that you can see using kubectl. @ssorj thoughts?

nluaces:
@lynnemorrison do you mind to keep the output status command as it is now and just include in this PR the `--output` flag?

We can address later what to show in the status command output in this issue: #2237

Thank you!

nluaces (approved):

PR #2066, opened 2025-04-09
Enable HA on e2e tests

granzoto:
Adds HA test to e2e test set


tests/Makefile
```diff
@@ -102,4 +102,4 @@ test-subset: create-venv
 
 # Run a subset of tests (comma-separated list) for CI
 ci-tests: TESTS=$(TESTS_CI)
-ci-tests: test-subset
```
rafaelvzago, line 105: I think this is just a new empty line. Correct?

granzoto: Yes, the test-subset is in the latest line in the file

rafaelvzago (approved):

PR #2067, opened 2025-04-09
Add test targets for running specific, e2e, subset, and CI tests

### LINKED ISSUE #2063: Feature Request: Add a Top-Level make test Target
Currently, there isn't a straightforward way to execute the tests located in the `tests/` subdirectory directly from the top-level `Makefile`. Developers need to navigate into the `tests/` directory and then run `make` to execute the test targets. This adds an extra step and can be inconvenient, especially when wanting to run tests frequently or as part of a larger build/test workflow from the project root.

I would like to add a new target to the main `Makefile` that, when executed, navigates to the `tests/` subdirectory and runs the targets defined within the `tests/Makefile`. Ideally, this new target would allow specifying which targets in the `tests/Makefile` to execute (e.g., `make test-all`, `make test-unit`). A default target could also be defined to run a standard set of tests.

One alternative is to always `cd tests/` and run `make` from there. However, this requires the developer to manually change directories. Another alternative could be to copy the test targets into the main `Makefile`, but this would lead to duplication and make maintenance more difficult as the test logic evolves.

**Additional context**
This feature would improve the developer experience by providing a more convenient way to run tests. It would also make it easier to integrate test execution into automated build and CI/CD pipelines triggered from the project root. A possible implementation could involve using the `-C` flag of the `make` command to change the directory before executing the targets in the `tests/Makefile`. For example, a target named `test` in the main `Makefile` could be defined as:

```makefile
test:
	$(MAKE) -C tests/
```

Or to run a specific target within the `tests/Makefile`:

```
e2e-tests:
	$(MAKE) -C tests/ unit
```

This approach leverages the existing Makefile within the tests/ directory and avoids duplicating any test definitions.
```makefile
test-unit:
	$(MAKE) -C tests/ unit
```

This approach leverages the existing `Makefile` within the `tests/` directory and avoids duplicating any test definitions.

rafaelvzago:
Add test targets for running specific, e2e, subset, and CI tests.

Closes #2063

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,3 +183,19 @@ clean:
 		cover.out oci-archives bundle bundle.Dockerfile \
 		skupper-*.tgz artifacthub-repo.yml \
 		network-observer-*.tgz  skupper-*-scope.yaml
+
+# Run a specific test
+test:
+	$(MAKE) -C tests/ test TEST="$(TEST)"
+
+# Run all e2e tests
+test-e2e:
+	$(MAKE) -C tests/ e2e-tests
+
+# Run a subset of tests in parallel
+test-subset:
+	$(MAKE) -C tests/ test-subset TESTS="$(TESTS)"
+
+# Run CI tests
+test-ci:
+	$(MAKE) -C tests/ ci-tests


Makefile
```diff
@@ -183,3 +183,19 @@ clean:
 		cover.out oci-archives bundle bundle.Dockerfile \
 		skupper-*.tgz artifacthub-repo.yml \
 		network-observer-*.tgz  skupper-*-scope.yaml
+
+# Run a specific test
+test:
```
fgiorgetti, line 188: A `test` targe already exists. I believe this one can be removed as I do not see
an advantage of keeping it, considering that you also introduced `test-subset`.

Makefile
```diff
@@ -183,3 +183,19 @@ clean:
...
+	$(MAKE) -C tests/ test TEST="$(TEST)"
+
+# Run all e2e tests
+test-e2e:
+	$(MAKE) -C tests/ e2e-tests
```
fgiorgetti, line 193: Actually why don't we bring in only one new target, which could be this one: `test-e2e`.
A single target could do the work of invoking subtests if a variable is provided.

This would make it clearer as we'd have a single entrypoint to run e2e tests.

Makefile
rafaelvzago: DOne under: 4cf39555a81086798f6bb787ceef1354227a4224

Makefile
rafaelvzago: Done under: 4cf39555a81086798f6bb787ceef1354227a4224

tests/README.md
```diff
@@ -133,6 +139,36 @@ make test-subset TESTS=hello-world,attached-connector
...
+### Available Commands
+
+For example:
+
+```bash
+# Run the ci-tests subset from the root
+make tests-e2e ci-tests
+
+# Run all E2E tests from the root
+make tests-e2e e2e-tests
```
fgiorgetti, line 161: Should these be: `test-e2e`, instead?

tests/README.md
```diff
@@ -133,6 +139,36 @@ make test-subset TESTS=hello-world,attached-connector
...
+
+# Run a specific test from the root
+make test-e2e TEST_TYPE=test TEST_OPTIONS="TEST=hello-world"
+
+# Run a specific test with a subset of tests from the root
+make test-e2e TEST_TYPE=test-subset TEST_OPTIONS="TESTS=hello-world,attached-connector"
```
fgiorgetti, line 167: The reason for my first comment is that it feels too complex to run samples like this one...

Makefile
```diff
@@ -26,6 +26,10 @@ DOCKER := docker
 SKOPEO := skopeo
 PODMAN := podman
 
+# The default e2e test type is set to "ci-tests" if not provided
+TEST_TYPE ?= ci-tests
+TEST_OPTIONS ?= ""
+
```
fgiorgetti, line 32: What do you think of making it simpler, by introducing just one variable, like `E2E_OPTIONS ?= ci-tests` ?
With that, the `test-e2e` target below, could be simplified into something like:

```
$(MAKE) -C tests/ $(E2E_OPTIONS)
```

And the execution could be done as:

```
make test-e2e
make test-e2e E2E_OPTIONS="e2e-tests"
make test-e2e E2E_OPTIONS="test TEST=hello-world"
```

PR #2072, opened 2025-04-15
Cleanup of internal/certs/certs.go and other related files.

ganeshmurthy:
Added unit test certs_test.go

### DIFF
--- a/internal/certs/certs_test.go
+++ b/internal/certs/certs_test.go
@@ -1,0 +1,76 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


internal/certs/certs_test.go
```diff
@@ -0,0 +1,76 @@
...
+	cert, err := DecodeCertificate(data)
+
+	if err != nil {
+		t.Error("Error decoding certificate")
+	}
+
+	notBefore := time.Now()
+	expiration := 5 * 365 * 24 * time.Hour
+	notAfter := notBefore.Add(expiration)
+	date_match := cert.NotAfter.Year() == notAfter.Year() && cert.NotAfter.YearDay() == notAfter.YearDay()
```
c-kruse, line 44: Out of an abundance of caution, I'd suggest logic like this where we pick a window of time that we expect the certificate's expiry to be. Comparing calendar date and year is an opportunity for this to flake.

```
notAfterTolerance := time.Hour
assert.Assert(t, notAfter.Add(-1 * notAfterTolernance / 2).Before(cert.NotAfter))
assert.Assert(t, cert.NotAfter.Before(notAfter.Add(notAfterTolernance / 2)))
```

c-kruse (approved):
Very nice changes @ganeshmurthy!

Just the one note on tricks for making assertions about go times.

PR #2080, opened 2025-04-22
update generators

ajssmith:
Take a approach similar to kind-dev-cluster. Scripts generate resources and reduce kustomize dependencies across chart, deploy yamls, and operator.


nluaces (approved):

fgiorgetti (approved):
Minor suggestion only (feel free to ignore) and I have also noticed the same difference pointed out by @c-kruse.

Other than that, nice stuff. I really like it.

PR #2085, opened 2025-04-22
Refactor qdr.Agent methods to encourage type-safety

### LINKED ISSUE #2081: qdr router configuration sharp edges
**Describe the bug**
When adding a numeric field to a router configuration entity, the standard `convert` function that takes the configuration entity as a go type and converts it into a map[string]interface to pass to the amqp client incidentally converts a fixed point integer into a floating point number because of JSON subtleties. This hits a corner case in the skupper router that has no validation and causes the field to be cleared/defaulted.

We have encountered this before, and have worked-around by adding a per-type conversion function (qdr.asRecord, qdr.asListenerRecord) to do the mapping while preserving types.

**Proposed solution**

Let's do away with the convert function, and make the status-quo something more difficult to get wrong.

c-kruse:
Enforces a per-type implementation to map each router configuration entity's go type to a Record type (map[string]any.) Replaces the use of the qdr.convert function, which can cause subtle bugs when working with numeric types.

Fixes #2081

### DIFF
--- a/internal/qdr/amqp_mgmt.go
+++ b/internal/qdr/amqp_mgmt.go
@@ -82,6 +82,10 @@ func getSiteMetadataString(siteId string, version string) string {
 	return string(metadata)
 }
 
+type recordType interface {
+	toRecord() Record
+}
+
 type Record map[string]interface{}
 
 func (r Record) AsString(field string) string {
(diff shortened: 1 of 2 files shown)


fgiorgetti (approved):
Neat.

nluaces (approved):

PR #2088, opened 2025-04-24
update e2e tests readme

pwright:


tests/README.md
```diff
@@ -81,6 +81,19 @@ The Makefile will automatically:
 
 ### 2. Run Test
 
+**Prerequisites**
+
+* A cluster with Skupper installed cluster-wide
+* Your kubeconfig is set to that cluster
+
```
rafaelvzago, line 88: Before start the cluster, it's necessary to build the images, in this case since you are using podman images for your cluster should be:

```bash
# Under the project root folder
make podman-build
```

pwright: how about:

If you are testing changes to Skupper code, create the images you want to test by running the following command in the project root folder:
```
make podman-build
```

rafaelvzago (approved):

PR #2091, opened 2025-04-28
Improves CLI error messages if CRDs are not installed

### LINKED ISSUE #2083: Improve the error message when the CRDs are missing
Improve the error message when the CRDs are missing

```
$ skupper site create west
the server could not find the requested resource (post sites.skupper.io) 
```

We want something along the lines of:

```
The Skupper CRDs are not yet installed.  To install them, do X.
``

### PR DESCRIPTION
**Pull request by lynnemorrison**
Improve the error message the CLI prints if user tries to create a resource but the CRDs have not been installed yet.

fixes #2083

### CODE REVIEW

**Comment by nluaces** [comment]
There are a few cases in which the command code has been modified in this PR but the message is not shown:
```
$skupper link generate
there is no skupper site in this namespace
there is no active skupper site in this namespace

$skupper link status
there is no skupper site available

$skupper token issue ~/token1.yaml
A site must exist in namespace default before a token can be created
```
This is because the error is caught after the input validation.


There are other cases in which the error has not been catched:
```
$skupper site delete
the server could not find the requested resource (get sites.skupper.io)
```

I suggest to check that the correspondent CRD for the command is installed in the validation function as the first step.

nluaces:
(@lynnemorrison my user handle in github is `nluaces` you are addressing another person by mistake)

nluaces:
This is a nice change, there are two things that we could include in this pr:
1) show this errors for the remaining generate commands: site generate, listener generate, connector generate
2) include this case in the unit tests given that we can mock this error

nluaces:
> For the generate commands we do not create a kubeclient since it is not needed for this command.
> `cli, err := client.NewClient(cobraCommand.Flag("namespace").Value.String(), cobraCommand.Flag("context").Value.String(), cobraCommand.Flag("kubeconfig").Value.String())`

You are right. I was testing `skupper link generate command` and it checked the CRDs, this was the exception because it can create credentials and secrets using kubeclient.

I take that back, leave the generate commands as they are. 

We only need to add the absence of CRDs case in the unit tests.

nluaces (approved):

PR #2092, opened 2025-04-29
update to go 1.24

ajssmith:


nluaces (approved):

c-kruse (commented):
Didn't see this coming!

https://tip.golang.org/doc/go1.24#vet

Looks like bumping the language version in go.mod to 1.24 invites some new static analysis on `go test`.

We should definitely go through and fix these violations. Until then, our options are:
- bump to go 1.23 instead.
- add `-vet=off` to our TESTFLAGS in Makefile:6

ajssmith:
Yes, working through those issues now. Leaning towards cleaning them up and getting to 1.24. I will push an update soon...

Issue #2007, opened 2025-03-17
Token redeem for system sites


ssorj:
```
$ export SKUPPER_PLATFORM=podman

$ skupper token redeem ~/token.yaml
command not supported by the selected platform
-> 1
```

Tokens are the primary method for linking sites, so it's a priority to support them.  *Redemption* for system sites is particularly useful (issuing tokens is less important for now).

PR #2115, opened 2025-05-12
Token redeem command for nonkube environments

nluaces:
Solves #2007


internal/cmd/skupper/token/nonkube/token_redeem.go
```diff
@@ -1,30 +1,96 @@
...
+		} else {
+			cmd.fileName = args[0]
+		}
+	}
+
+	// Validate if a token file exists
+	if cmd.fileName != "" {
+		_, err := os.Stat(cmd.fileName)
+		if err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("token file does not exist: %s", err))
```
c-kruse, line 65: Suggest "cannot open token file" or something to that effect. "does not exist" is not strictly correct.

internal/cmd/skupper/token/nonkube/token_redeem.go
```diff
@@ -1,30 +1,96 @@
...
+	// get data from the token file
+	var accessToken v2alpha1.AccessToken
+	var tokenFile []byte
+	tokenFile, err := os.ReadFile(cmd.fileName)
+	if err != nil {
+		err = fmt.Errorf("unable to read token file - %v", err)
+		return err
+	}
+	resource := json.NewSerializerWithOptions(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, json.SerializerOptions{Yaml: true})
+	resource.Decode(tokenFile, nil, &accessToken)
```
c-kruse, line 82: Is there a way to do error handling here? i.e. `skupper token redeem go.mod` should probably fail because a go.mod file isn't a v2alpha1.AccessToken.

internal/cmd/skupper/token/nonkube/token_redeem.go
nluaces: fixed

internal/cmd/skupper/token/nonkube/token_redeem.go
nluaces: right! thanks!

c-kruse (approved):
Thanks!

fgiorgetti:
@nluaces My understanding is that the `token redeem` command should actually redeem the AccessToken and place
a `Link` and a `Secret` pair into the `input/resources` directory instead of placing the AccessToken itself.
Therefore a system site can be started or reloaded as many times as needed.
Thoughts?

nluaces:
@fgiorgetti what you are saying makes total sense; it needs an active site in the runtime directory before redeeming the token.

I think I was focusing on creating all the resources (site, token) at once. I will change that.

internal/cmd/skupper/token/nonkube/token_redeem.go
```diff
@@ -1,30 +1,133 @@
...
+		}
+	}
+
+	// Validate there is already a site defined in the namespace before a token can be redeemed
+	pathProvider := fs.PathProvider{Namespace: cmd.Namespace}
+	siteStateLoader := &nonkubecommon.FileSystemSiteStateLoader{
+		Path: pathProvider.GetRuntimeNamespace(),
+	}
+
+	siteState, err := siteStateLoader.Load()
```
fgiorgetti, line 70: I believe this could be avoided. The reason, as I can see, for a site to exist is just to provide a site name
as the subject when redeeming the AccessToken. But I believe that, instead, if we could pass the namespace
name as an alternative, we should be good, then there is no need to require an active site to exist in advance.

PR #2117, opened 2025-05-12
Add codespell support (config, workflow to detect/not fix) without fixes

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Note that there was a fix in a data-structure and potentially API for `indicies` -- please advise if internal and thus ok to be fixed or better be skipped?

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -3,4 +3,4 @@
 skip = .git*,go.sum,.codespellrc
 check-hidden = true
 # ignore-regex = 
-# ignore-words-list =
+ignore-words-list = dne,haa
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,go.sum,.codespellrc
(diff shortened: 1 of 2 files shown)


ted-ross:
If the maintainers decide to adopt this change, it should probably be split into two pull requests:  one for the configuration change, and another with the spelling fixes.

This appears to me to be adding significant value.

c-kruse:
Thanks for this @yarikoptic.

I clearly need a spell checker in CI! Like @ted-ross suggested, I'd be interested in seeing this change without the corrections - if only to understand what we'll see when introducing a typo.

yarikoptic:
ok, here now only CI without fixes, rebased on top of current main

yarikoptic:
note: by mistake (was intended for https://github.com/openai/codex) pushed commit for pinning actions... up to you, could do similar one here

c-kruse (approved):
Thank you for the PR!

fgiorgetti (approved):

PR #2121, opened 2025-05-14
System controller

fgiorgetti:
# Introducing: System Controller

Its initial goal is to keep the runtime resource statuses updated, so that
users can easily determine the current VAN state.

The system controller can run as a standalone process, built using: `make system-controller`,
or as a container (depending on the CONTAINER_ENGINE environment variable, default set to podman),
by executing `cmd/system-controller/system-controller.sh`.

## How it works?

A single instance of the `system-controller` can be executed per-user and it watches all
Skupper namespaces defined under the given user, monitoring the state of a namespace and
controlling the lifecycle of the Collector Lite, based on the availability of the respective
router process (per namespace).

The Collector Lite has been subtly modified to work with Kubernetes and System sites, allowing
the `skupper-network-status` ConfigMap to be updated, regardless of the platform where it is
running.

Once the ConfigMap is updated by the Collector Lite, the System Controller will parse the
current VAN state and update the respective resources under `runtime/resources` for a namespace.

If the Router is brought down, for any reason, the Collector Lite is stopped and runtime resources
are updated back to the initial state (Configured).

### NamespacesHandler

Main controller handler triggered by the creation or removal of a namespace.
It creates a NamespaceController thread per namespace.

### NamespaceController

A thread created to control the components needed for each existing namespace.

### RouterConfigHandler

Detects when the router configuration has been created, meaning that a namespace has
been initialized, triggering the RouterStateHandler.

### RouterStateHandler

Run a Heartbeats client that informs when the router state changes (up or down), informing
the CollectorLifecycleHandler about it.

### CollectorLifecycleHandler

Controls the lifecycle of the Collector Lite based on router availability.
It is triggered by the RouterStateHandler component.

### NetworkStatusHandler

Processes changes done to the `skupper-network-status` ConfigMap and updates the statuses
of the respective `runtime/resources`.

## Related changes needed

Here is a list of all changes needed to existing components.

## internal/cmd

Some of the nonkube command implementations where using `api.GetHostSiteInternalPath`. This is OK
when you're only dealing with the actual path on the host machine, but since the System Controller
can run as a container, the commands needed to be modified to use `api.GetInternalOutputPath` instead.

There is another advantage on this change for testing purposes, as instead of setting the `SKUPPER_OUTPUT_PATH`
environment variable to customize where test resources are expected to be placed, you can set the `XDG_DATA_HOME`.
The `SKUPPER_OUTPUT_PATH` is meant to be used inside containers only, when the output path is mounted into
the container, while the `XDG_DATA_HOME` is always validated on rootless cases.

Another related change needed for testing purposes, was done to the allow a `DefaultRootDataHome` to be specified,
which allows the Skupper data home directory to be customized when using rootful mode. Previously it was set
statically to `/var/lib`. Now the `internal/cmd` tests have been tweaked to use both.

### internal/kube/flow -> internal/flow

An abstraction of the `StatusSync` has been extracted and moved from `internal/kube/flow` into `internal/flow`.
It now depends on the `StatusSyncClient` implementations, which are meant to provide mechanisms for manipulating
the `skupper-network-status` ConfigMap, regardless of the platform where it is running.

### internal/fs/watcher.go

The FileWatcher implementation has been refactored and improved as it was not yet been used internally.
Along with that, a new set of tests have been created to ensure a reliable behavior.

### internal/kube/controller/controller.go

The `extractSiteRecords` generic function has been exported moved to the `internal/network` package.

### internal/kube/site/site.go

The `getLinkRecordsForSite` generic function has been exported and moved to the `internal/network` package.

### internal/nonkube/bundle

Bundle generation now needs to update the TCP port allocated dynamically by the `router_free_port.py` tool
into the runtime resource `RouterAccess-skupper-local.yaml`, as this resource is used by the System Controller
to identify the listener port used to communicate with the Router.

Another change needed is that, now, the generated `Site ID` must be updated into the Site.metadata.uid under
`runtime/resources`.

### internal/nonkube/client/fs

* The ConfigMap kind is now supported
* BaseCustomResourceHandler and PathProvider now work inside a container


c-kruse (commented):
Looks really nice so far!

grs (approved):

pkg/vanflow/eventsource/manager.go
```diff
@@ -94,6 +101,8 @@ func (m *Manager) sendRecords(ctx context.Context) {
 			msg, err := record.Encode()
 			if err != nil {
 				m.logger.Error("skipping record message after encoding error:", slog.Any("error", err))
+				recordJson, _ := json.MarshalIndent(record, "", "  ")
+				m.logger.Error(string(recordJson))
```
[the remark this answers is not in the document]
fgiorgetti: This is certainly something that should not be here :)
I possibly added it for debugging purposes at some point, and forgot it here.
I am removing it now. Thanks for catching it.

internal/nonkube/flow/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/skupperproject/skupper/pkg/vanflow/session"
+	"github.com/skupperproject/skupper/pkg/vanflow/store"
+)
+
+type ControllerConfig struct {
+	Factory session.ContainerFactory
+	Site    vanflow.SiteRecord
+}
+
+func NewController(cfg ControllerConfig) *Controller {
```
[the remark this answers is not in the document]
c-kruse: Great. Yeah there should be plenty of ways to establish router liveness without adding a new vanflow event source.

We shouldn't have any problems omitting it. The types of things we miss out on would be future enhancements. Like the podman controller in v1 shared PROCESS records for the other running podman containers that could help the collector fill out some details on where application traffic was coming from/going to. I'm not so sure that's totally relevant now that we use the host network namespace, but we could eventually end up wanting some context from the site host - and this would be a way to get it there.

Nothing I can think of would stop us from adding it in later, though!

internal/nonkube/flow/collector.go
```diff
@@ -0,0 +1,131 @@
...
+	statusSync := flow.NewStatusSync(factory, nil, statusSyncClient, types.NetworkStatusConfigMapName)
+	go statusSync.Run(ctx)
+	go func() {
+		<-ctx.Done()
+		_ = client.Delete(cm.Name, true)
+	}()
+	return nil
+}
+
+func getLocalTLSConfig(namespace string) (*tls.Config, error) {
```
c-kruse, line 101: Is there a way to reuse what we've got in the runtime package?

c-kruse (approved):
Great work!

internal/fs/watcher.go
```diff
@@ -0,0 +1,280 @@
+package fs
```
ganeshmurthy, line 1: Can this file be called file_watcher. go ? Also can the name of package name be expanded from fs to something more descriptive ?

cmd/system-controller/system-controller.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/sh
+
+set -Ceu
```
ganeshmurthy, line 3: Can we use one or all of these set options - https://github.com/skupperproject/skupper-router/blob/main/.github/scripts/compile.sh#L23 ?

internal/nonkube/flow/collector.go
fgiorgetti: Nice catch.

internal/fs/watcher.go
fgiorgetti: Do you think that `filesystem/file_watcher.go` would be a more accurate name?

ganeshmurthy: > Do you think that `filesystem/file_watcher.go` would be a more accurate name?

Yes, that would be most excellent !

cmd/system-controller/main.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+
+	"github.com/skupperproject/skupper/internal/nonkube/controller"
+	"github.com/skupperproject/skupper/internal/version"
+	"github.com/skupperproject/skupper/pkg/nonkube/api"
+)
+
+func main() {
+	parseFlags()
+	log.Printf("Version: %s", version.Version)
```
nluaces, line 20: should the system controller version be included in the images.go file like the rest of components?

internal/cmd/skupper/site/nonkube/site_status_test.go
```diff
@@ -26,9 +26,12 @@ func TestCmdSiteStatus_ValidateInput(t *testing.T) {
...
 
-	tmpDir := filepath.Join(t.TempDir(), "/skupper")
-	err := os.Setenv("SKUPPER_OUTPUT_PATH", tmpDir)
-	assert.Check(t, err == nil)
+	if os.Getuid() == 0 {
+		api.DefaultRootDataHome = t.TempDir()
+	} else {
+		t.Setenv("XDG_DATA_HOME", t.TempDir())
+	}
+	tmpDir := api.GetDataHome()
```
nluaces, line 34: I guess these changes have to be replicated in the rest of CLI unit tests, right? (in a subsequent PR)

cmd/system-controller/system-controller.sh
fgiorgetti: I think we can set it to: `set -Ceufxo`, in order to keep it Posix shell compliant. WDYT?

cmd/system-controller/main.go
fgiorgetti: Nice catch @nluaces ! I will add them.

internal/cmd/skupper/site/nonkube/site_status_test.go
fgiorgetti: I think, I have already done that:

```
internal/cmd/skupper/link/nonkube/link_generate_test.go
internal/cmd/skupper/site/nonkube/site_update_test.go
internal/cmd/skupper/site/nonkube/site_delete_test.go
internal/cmd/skupper/site/nonkube/site_status_test.go
internal/cmd/skupper/version/nonkube/version_test.go
```

But if something is missing, we can add them later. WDYT?

nluaces: Oh! indeed, I meant that _I_ have to adapt those because I have a pending pull request relying heavily on SKUPPER_OUTPUT_PATH

fgiorgetti: Thank's Noe!

Makefile
```diff
@@ -57,6 +57,10 @@ build-doc-generator: generate-doc
 generate-doc: $(call pkgdeps,./internal/cmd/generate-doc)
 	GOOS=${GOOS} GOARCH=${GOARCH} go build -ldflags="${LDFLAGS}"  -o $@ ./internal/cmd/generate-doc
 
+build-system-controller: system-controller
```
nluaces, line 60: `system-controller` needs to be added in the `all` target, right?

internal/nonkube/client/fs/configmap_handler.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return &context, nil
+}
+
+func (s *ConfigMapHandler) Delete(name string, runtime bool) error {
+	fileName := name + ".yaml"
+
+	if err := s.DeleteFile(s.getPath(runtime), fileName, common.ConfigMaps); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
```
nluaces, line 98: probably we need to return the error even if the file does not exist, (I'm modifying the rest of the handlers in other PR)

nluaces (approved):
Tested in local, looks good to me!

internal/nonkube/flow/collector.go
```diff
@@ -0,0 +1,111 @@
...
+	tlsCert := runtime.GetRuntimeTlsCert(namespace, "skupper-local-client")
+	config, err := tlsCert.GetTlsConfig()
+	if err == nil {
+		config.MinVersion = tls.VersionTLS13
+	}
+	return config, nil
```
c-kruse, line 102: Some missed error handling here!

Also curious about the meddling with tls configuration MinVersion here and in GetTlsConfig(). The default in the latter (MinVersion = tls.VersionTLS10) feels likely to draw attention - although I suppose this is intended to mostly run in a trusted context over loopback.

fgiorgetti: Yes, the reason why I have kept the original one to use TLS10, is just to avoid an extra impact to the kube-adaptor.
I believe we could raise a separate issue and PR to update it as well, at least it is easier to test and evaluate an
eventual impact.
I will fix this missing error handling.

PR #2122, opened 2025-05-14
Fixes Certificates Manager State Bug

c-kruse:
Recently an optimization was made to reduce the number of updates to v2alpha1.Certificate resources for updating readiness status. This optimization introduced a bug that was preventing the controller from recovering Certificate state after a restart - manifesting in Site status errors. This change keeps the optimization in place, but fixes the manager state tracking logic.

Follow up for https://github.com/skupperproject/skupper/pull/1993

### DIFF
--- a/internal/kube/certificates/mgr.go
+++ b/internal/kube/certificates/mgr.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


c-kruse:
@grs could use some help double checking my assumptions about what was going on here. After controller restarts we've been seeing this. I'm pretty sure i've tracked it down to this.
```
k get sites -A
NAMESPACE   NAME   STATUS   SITES IN NETWORK   MESSAGE
default     east   Error    2                  certificates.skupper.io "skupper-site-ca" already exists
west        west   Error    2                  certificates.skupper.io "skupper-site-ca" already exists
```

nluaces (approved):

PR #2130, opened 2025-05-19
Added s390x arch in Makefiles to build multi arch images and skipped some tests on s390x because of image unavailability

rajnish-jauhari:
Added s390x arch in Makefiles to build multi arch images and 
Skipping below tests on s390x because of image unavailability :
- TestGateway/local-gateway-service
- TestBookinfo
- TestMongo
- TestHipsterShop
- TestPostgres

cc: @hash-d

### DIFF
--- a/test/integration/examples/bookinfo_test.go
+++ b/test/integration/examples/bookinfo_test.go
@@ -6,12 +6,16 @@ package examples
 import (
 	"context"
 	"testing"
+	"runtime"
 
 	"github.com/skupperproject/skupper/test/integration/examples/bookinfo"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
 func TestBookinfo(t *testing.T) {
+	if runtime.GOARCH == "s390x" {
+		t.Skip("Skipping test on s390x architecture as bookinfo images support is unavailable for s390x")
+	}
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	bookinfo.Run(ctx, t, testRunner)
--- a/test/integration/examples/custom/hipstershop/hipstershop_test.go
+++ b/test/integration/examples/custom/hipstershop/hipstershop_test.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"testing"
+	"runtime"
 
 	"github.com/skupperproject/skupper/test/utils/base"
 	"github.com/skupperproject/skupper/test/utils/constants"
@@ -29,6 +30,9 @@ func TestMain(m *testing.M) {
 }
 
 func TestHipsterShop(t *testing.T) {
+	if runtime.GOARCH == "s390x" {
+		t.Skip("Skipping test on s390x architecture as images unavailble for s390x")
+	}
 	// Cluster needs for hipster shop
 	needs := base.ClusterNeeds{
 		NamespaceId:     "hipster",
--- a/test/integration/performance/postgres_test.go
+++ b/test/integration/performance/postgres_test.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"runtime"
 	"time"
 
 	"github.com/skupperproject/skupper/test/integration/performance/common"
@@ -45,6 +46,9 @@ type postgresSettings struct {
 }
 
 func TestPostgres(t *testing.T) {
+	if runtime.GOARCH == "s390x" {
+		t.Skip("Skipping test on s390x architecture as image unavailable")
+	}
 	// TestPostgres is currently not functional for ARM
 	// https://github.com/skupperproject/skupper/issues/1650
 	common.CheckArch(t)
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ TEST_IMAGE := quay.io/skupper/skupper-tests
 TEST_BINARIES_FOLDER := ${PWD}/test/integration/bin
 DOCKER := docker
 LDFLAGS := -X github.com/skupperproject/skupper/pkg/version.Version=${VERSION}
-PLATFORMS ?= linux/amd64,linux/arm64
+PLATFORMS ?= linux/amd64,linux/arm64,linux/s390x
 GOOS ?= linux
 GOARCH ?= amd64
 
--- a/test/images/Makefile
+++ b/test/images/Makefile
@@ -86,7 +86,7 @@ SKOPEO = skopeo
 # That is required for their use with Openshift 3.11
 FORMAT_OPTIONS = --format docker
 TRANSFORM_OPTIONS = --format v2s2
-PLATFORM = linux/amd64,linux/arm64
+PLATFORM = linux/amd64,linux/arm64,linux/s390x
 
 # Repositories
 MAIN_REPO = quay.io/skupper
--- a/test/integration/acceptance/gateway_test.go
+++ b/test/integration/acceptance/gateway_test.go
@@ -16,6 +16,7 @@ import (
 	"strconv"
 	"testing"
 	"time"
+	"runtime"
 
 	"github.com/skupperproject/skupper/test/integration/acceptance/gateway"
 	"github.com/skupperproject/skupper/test/integration/examples/tcp_echo"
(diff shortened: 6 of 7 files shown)


test/integration/examples/bookinfo_test.go
```diff
@@ -6,12 +6,16 @@ package examples
...
 	"context"
 	"testing"
+	"runtime"
 
 	"github.com/skupperproject/skupper/test/integration/examples/bookinfo"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
 
 func TestBookinfo(t *testing.T) {
+	if runtime.GOARCH == "s390x" {
```
hash-d: Checking `runtime.GOARCH` is ok for `testLocalGatewayService`, because that's a local test.  For `TestBookInfo`, however, you should be inspecting the cluster's architecture, not the local machine.

Check how that's done for ARM, you can adapt that for Z.

hash-d (changes requested):
For the tests that cannot be run on Z because of missing images, the cluster's architecture should be checked, instead of the local test machine.

If you're running from an x86 VM pointing to a Z cluster, for example, those tests would not be skipped and the test would fail.

rajnish-jauhari:
@hash-d made the changes as requested by you

test/integration/examples/custom/hipstershop/hipstershop_test.go
```diff
@@ -29,6 +31,13 @@ func TestMain(m *testing.M) {
 }
 
 func TestHipsterShop(t *testing.T) {
+	
+	//Skipping test on s390x architecture as images unavailable for s390x
+	cluster := base.GetClusterContext()
+	if err := arch.SkipOnlyS390x(t, cluster); err != nil {
+	t.Fatal(err)
```
hash-d: The check failed because of incorrect Go formatting.  This is one such case (missing indent).  Please fix these so the check passes.

test/utils/arch/k8s.go
```diff
@@ -53,3 +53,38 @@ func Skip(t *testing.T, clusters ...*base.ClusterContext) error {
 	}
 	return err
 }
+
+// CheckOnlyS390x skips ONLY if the architecture is s390x
+func CheckOnlyS390x(clusters ...*base.ClusterContext) (err error, skip bool) {
```
hash-d, line 58: It would have been best to generalize Check(), as listed on its TODO, but this works fine

hash-d (changes requested):
Please make the check pass, and we should be good to merge.

test/integration/examples/mongodb/mongo.go
```diff
@@ -41,9 +39,6 @@ func RunTests(ctx context.Context, t *testing.T, r *base.ClusterTestRunnerBase)
 	_, err = k8s.WaitForSkupperServiceToBeCreatedAndReadyToUse(pubCluster1.Namespace, pubCluster1.VanClient.KubeClient, "mongo-c")
 	assert.Assert(t, err)
 
-	// Make sure you have a minimum of three members on the replicaset
-	// https://www.mongodb.com/docs/manual/tutorial/deploy-replica-set/
-	// https://www.mongodb.com/docs/manual/core/replica-set-architecture-three-members/
```
hash-d, line 46: Why did you remove these comments?

test/integration/examples/mongodb/mongo.go
```diff
@@ -63,17 +58,18 @@ func RunTests(ctx context.Context, t *testing.T, r *base.ClusterTestRunnerBase)
 	`)
 	assert.Assert(t, err)
 
-	// Let's wait until the election is settled, for a maximum of 5 min
-	log.Print("Waiting for the Mongo replicaset primary member election to be settled")
+	log.Print("Waiting for Mongo replicaset primary election")
```
hash-d: Here, too, why make these changes?  What's the relation of them with the PR?

test/integration/examples/mongodb/mongo.go
```diff
@@ -82,12 +78,14 @@ func RunTests(ctx context.Context, t *testing.T, r *base.ClusterTestRunnerBase)
 
 	_, err = k8s.CreateTestJob(pubCluster1.Namespace, pubCluster1.VanClient.KubeClient, jobName, jobCmd)
 	assert.Assert(t, err)
-	job, err := k8s.WaitForJob(pubCluster1.Namespace, pubCluster1.VanClient.KubeClient, jobName, constants.ImagePullingAndResourceCreationTimeout)
-	jobLogs, _ := k8s.GetJobsLogs(pubCluster1.Namespace, pubCluster1.VanClient.KubeClient, jobName, true)
-	t.Logf("%s logs:", jobName)
-	t.Logf("%s", jobLogs)
```
hash-d, line 88: Why removing those logs?

test/integration/examples/bookinfo_test.go
```diff
@@ -8,11 +8,54 @@ import (
...
+	pub1, err := testRunner.GetPublicContext(1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	prv1, err := testRunner.GetPrivateContext(1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
hash-d: The changes above are completely changing how the test behaves, why are you doing that?  A shared `ClusterTestRunnerBase` already exists and is created on `test/integration/examples/main_test.go`, and was used by this test, why create a new `ClusterNeeds`?

You probably want to just add `arch.SkipOnlyS390x()` to `Setup()` on `test/integration/examples/bookinfo/bookinfo.go`, instead.

test/integration/performance/postgres_test.go
```diff
@@ -45,6 +46,35 @@ type postgresSettings struct {
 }
 
 func TestPostgres(t *testing.T) {
+	// Skip test if the cluster architecture is s390x
+	needs := base.ClusterNeeds{
+		NamespaceId:     "perf-archcheck",
+		PublicClusters:  1,
+		PrivateClusters: 0,
```
hash-d: Same thing as for Mongo test, here: why a new `ClusterNeeds`?

test/integration/performance/postgres_test.go
```diff
@@ -45,6 +46,35 @@ type postgresSettings struct {
...
+		t.Skipf("%s", err)
+	}
+
+	_, err := r.Build(needs, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Cleanup the temporary namespace used for arch detection
+	defer base.RemoveNamespacesForContexts(r, []int{1}, []int{})
```
hash-d: Why a temporary namespace for arch detection?  The creation and removal of namespaces add time to the test execution; use the actual namespaces, instead.

hash-d: Also, probably not the case here as the focus is on v2, but if new namespaces are introduced in the test by future changes, the temporary ones may be left behind, unaltered.  It's better to test the actual namespaces, in the place where they are created or accessed by the test, to ensure future changes take the verification you're introducing in mind.

test/utils/arch/k8s.go
```diff
@@ -10,19 +10,21 @@ import (
...
+	return labels["beta.kubernetes.io/arch"]
+}
+
 // If the target clusters are found to contain any non-amd64 nodes, return true.
 //
 // Notice that only the cluster is checked to be amd64; if the VM needs checked,
 // use build flags or some other technique.
-//
-// Usage: check first skip; only check err if skip is false.  If skip is true, error
-// will be non-nil, with information on why skipping
```
hash-d, line 19: Why removing the Usage documentation?

test/utils/arch/k8s.go
```diff
@@ -10,19 +10,21 @@ import (
...
-//
-// TODO: make it more granular, allow for hibrid clusters?
-// TODO: allow for list of accepted archs?
```
hash-d, line 22: These TODOs have not been implemented; why removing them?

rajnish-jauhari:
@hash-d all the conflicts have been resolved

test/utils/arch/k8s.go
```diff
@@ -53,3 +53,38 @@ func Skip(t *testing.T, clusters ...*base.ClusterContext) error {
...
+	for _, c := range clusters {
+		list, err := c.VanClient.KubeClient.CoreV1().Nodes().List(ctx, v1.ListOptions{})
+		if err != nil {
+			return err, false
+		}
+		for _, node := range list.Items {
+			arch := node.Labels["beta.kubernetes.io/arch"]
+			if arch == "s390x" {
+				return fmt.Errorf(
+					"at least one cluster node is not s390x -- skipping (%s at %s is %q)",
```
hash-d: I think this should read

> "at least one cluster node **is** s390x -- skipping (%s at %s is %q)",

rajnish-jauhari: it's done

PR #2133, opened 2025-05-19
Implements Certificate Rotation Fundamentals

c-kruse:
Adds a new internal/kube/secrets package that implements the logic for watching kubernetes Secret resources containing the credentials to be used in router sslProfile configuration, and automatically invoking the certificate rotation primitives in router configuration when Secret data changes.

Introduces a new partially backwards incompatible interface between the controller and kube-adaptor in order to accurately associate a given Secret Resource revision with an sslProfile+ordinal pair. Prior versions of the kube-adaptor should work with the updated controller logic, but updated kube-adaptors will wait for an updated controller before starting. This should be safe when upgrading, but potentially annoying in development.

### DIFF
--- a/doc/tls/README.md
+++ b/doc/tls/README.md
@@ -255,7 +255,7 @@ This will apply the link to the private site. The sites should connect.
 **TODO** Support for rotating TLS Credentials is currently under development
 and is not yet fully implemented.
 
-Skupper uses Links to communicate between Sites. Each Link consists of a family
+Skupper uses Links to communicate between Sites. Each Link consists of a set
 of TLS Connections that Skupper routers use for internal coordination and
 proxying application traffic across Sites. Rotating the TLS Credentials used by
 a Link will initiate an active migration where the routers will open new TLS
@@ -268,16 +268,15 @@ connections.
 ### Updating TLS Credentials
 
 Skupper automatically watches for changes to Secrets being used for TLS
-Credentials. When the Secret is updated, Skupper will immediately begin to
-prefer the new set of credentials. For a RouterAccess any new connections will
-use the updated credentials, but existing connections are only terminated when
-they are based on a prior revision that becomes invalid (more below.) For Link
-resources, skupper will proactively begin the migration to the new set of
-credentials.
+Credentials. When the Secret is updated, Skupper immediately prefers the new
+set of credentials. For a RouterAccess any new connections will use the updated
+credentials, but existing connections are only terminated when they are based
+on a prior revision that becomes invalid (more below.) For Link resources,
+skupper will proactively begin the migration to the new set of credentials.
 
 ### Discontinuing Use of Previous TLS Credentials
 
-When TLS Credentials are rotated, Skupper will preserve the connections based
+When TLS Credentials are rotated, Skupper preserves the connections based
 on the previous revision of the credentials to ensure service continuity for
 open application connections. This behavior can be overwritten at the Site with
 the `tls-prior-valid-revisions` setting or at the Secret with the


internal/kube/secrets/manager.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+		if updateSecretChecksum(secret, &state.SecretContentSum) {
+			state.Ordinal += 1
+			changed = true
+		}
+		pv := w.checkPriorValidity(secret)
+		nextOldest := state.Ordinal - pv
+		if pv <= state.Ordinal && nextOldest > state.OldestValidOrdinal {
+			changed = true
+			state.OldestValidOrdinal = nextOldest
```
Karen-Schoener, line 95: Eventually, will a user be able to configure the oldest valid ordinal for a link secret or for a skupper-site-server secret?  Thanks, Karen

c-kruse: Hi @Karen-Schoener.

The plan so far is to not expose the idea of ordinals directly to the user. Instead, the interface is to be:
1. To use an updated set of credentials, directly update the Secret.
2. To force invalidation of previous revisions of a secret/set of tls credentials, apply `tls-prior-valid-revisions` - either to the Site resource .spec.settings, or to the Secret itself using the `skupper.io/tls-prior-valid-revisions` annotation.
   - Setting tls-prior-valid-revisions=0 would mean always set oldestValidOrdinal to the latest ordinal. (i.e. no seamless migration.) Default I've chosen as 2 - so a certificate could be rotated twice before existing connections are disrupted.
   - (Pending getting ordinal/connection info back to the user to confirm the new credentials are in use) I think we'll end up suggesting a workflow like this: Update Secret; Wait for Confirmation that the latest secret is in use; IFF the previous key was compromised then apply the `skupper.io/tls-prior-valid-revisions=0` annotation to the secret to close the old connections (disrupts application traffic.) Clear the `skupper.io/tls-prior-valid-revisions` annotations so that the future migrations will complete seamlessly.

Karen-Schoener (approved):

nluaces (approved):

internal/kube/adaptor/config_init.go
```diff
@@ -2,29 +2,67 @@ package adaptor
...
+	"github.com/cenkalti/backoff/v4"
+	internalclient "github.com/skupperproject/skupper/internal/kube/client"
+	"github.com/skupperproject/skupper/internal/kube/secrets"
+	"github.com/skupperproject/skupper/internal/kube/watchers"
 	"github.com/skupperproject/skupper/internal/qdr"
 )
 
-func InitialiseConfig(client kubernetes.Interface, namespace string, path string, routerConfigMap string) error {
+func InitialiseConfig(cli internalclient.Clients, namespace string, path string, routerConfigMap string) error {
 	ctxt := context.Background()
```
ganeshmurthy, line 23: Should there be a line here that checks to see if the passed in parameter `namespace` is `nil` and if `nil` get it from `cli.GetNamespace()` ?

doc/tls/README.md
```diff
@@ -255,11 +255,45 @@ This will apply the link to the private site. The sites should connect.
...
-[trust model](#default-controller-managed-tls-in-kubernetes) used by Skupper is
-distributed: each Site having its own trust root. Because of this, rotating
-client certificates and Site CAs is an especially complicated exercise.
+Skupper uses Links to communicate between Sites. Each Link consists of a family
+of TLS Connections that Skupper routers use for internal coordination and
+proxying application traffic across Sites. Rotating the TLS Credentials used by
+a Link will initiate an active migration where the routers will open new TLS
+connections to their peers using the updated credentials. Once established, all
+internal communications and any new application traffic will happen over the
+new connections based on the updated credentials. By default, Skupper will
```
ganeshmurthy, line 264:
```suggestion
new Link based on the updated credentials. By default, Skupper will
```

internal/kube/adaptor/config_init.go
```diff
@@ -2,29 +2,67 @@ package adaptor
...
-	current, err := client.CoreV1().ConfigMaps(namespace).Get(ctxt, routerConfigMap, metav1.GetOptions{})
-	if err != nil {
-		return err
+	controller := watchers.NewEventProcessor("config-init", cli)
+	secretsSync := secrets.NewSync(
+		sslSecretsWatcher(namespace, controller),
+		nil,
+		slog.New(slog.Default().Handler()).With(slog.String("component", "kube.secrets")),
+	)
+	stop := make(chan struct{})
```
ganeshmurthy, line 30:
```suggestion
	stopCh := make(chan struct{})
```

internal/qdr/qdr.go
```diff
@@ -371,10 +371,12 @@ type RouterMetadata struct {
...
 type SslProfile struct {
-	Name           string `json:"name,omitempty"`
-	CertFile       string `json:"certFile,omitempty"`
-	PrivateKeyFile string `json:"privateKeyFile,omitempty"`
-	CaCertFile     string `json:"caCertFile,omitempty"`
+	Name               string `json:"name,omitempty"`
+	CertFile           string `json:"certFile,omitempty"`
+	PrivateKeyFile     string `json:"privateKeyFile,omitempty"`
+	CaCertFile         string `json:"caCertFile,omitempty"`
+	Ordinal            uint64 `json:"ordinal,omitempty"`
```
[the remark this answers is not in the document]
ganeshmurthy: Darn ! I missed that when reviewing the router code ! Cool

PR #2143, opened 2025-05-29
Improve uniqueness of system site ids

fgiorgetti:
* Both system bundle and local initialization to include nanoseconds as part of ID (mitigate risk of ID clashes)
* The site entity of the router config was not using an exclusive name on bundle installations


c-kruse (approved):
Looks okay to me!

Existing behavior, but I notice now that SiteID changes with each `skupper system reload` - it is marginal, but I suspect it would be nicer for this to be a consistent identifier. Not sure where we'd get that, though. Something like `sha( host id, user, namespace)` I think would be ideal - not sure what options are for a unique host id are. hostname could easily collide - maybe some MAC or something we generate ourselves on skupper system install?

PR #2148, opened 2025-06-04
Install/uninstall system-controller container

### LINKED ISSUE #2149: Include system-controller container as part of the system install command
Currently in Skupper V2, the system-controller needs to be installed by running a separated script, and the `skupper system install/uninstall` commands just enable the socket for podman/docker.

`skupper system install` command should configure, create and start the system-controller container and,
`skupper system uninstall` command should remove stop and remove said container.

**Additional context**
This implementation should use the libpod library.

nluaces:
Resolves #2149

### DIFF
--- a/internal/nonkube/bootstrap/controller/systemd.go
+++ b/internal/nonkube/bootstrap/controller/systemd.go
@@ -121,12 +121,6 @@ func (s *systemdServiceInfo) Create() error {
 }
 
 func (s *systemdServiceInfo) GetServiceFile() string {
-	if api.IsRunningInContainer() {
-		outputStat, err := os.Stat("/output")
-		if err == nil && outputStat.IsDir() {
-			return path.Join(path.Join("output", string(api.ScriptsPath)), s.GetServiceName())
-		}
-	}
 	if s.GetUid() == 0 {
 		return path.Join(s.rootSystemdBasePath, s.GetServiceName())
 	}


internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +36,179 @@ func Install(platform string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to pull system-controller image: %v", err)
+	}
+	fmt.Printf("Pulled system-controller image: %s\n", images.GetSystemControllerImageName())
+
+	env := map[string]string{
+		"CONTAINER_ENDPOINT":  config.containerEndpoint,
+		"SKUPPER_OUTPUT_PATH": config.hostDataHome,
+		"CONTAINER_ENGINE":    config.containerEngine,
+		"XDG_DATA_HOME":       config.xdgDataHome,
```
fgiorgetti: Not sure XDG_DATA_HOME is needed.

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +36,179 @@ func Install(platform string) error {
...
+			Name:        socketPath,
+			Destination: "/var/run/docker.sock",
+			Mode:        "z",
+			RW:          true,
+		})
+	} else if strings.HasPrefix(config.containerEndpoint, "/") {
+
+		mounts = append(mounts, container.Volume{
+			Name:        config.containerEndpoint,
+			Destination: "/var/run/docker.sock",
```
fgiorgetti: It should also evaluate the containerEndpoint as well as the selected platform or containerEngine to deteremine which one to mount (podman or docker .sock file). The `system-controller.sh` has it.

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +36,179 @@ func Install(platform string) error {
 		return err
 	}
 
+	fmt.Printf("Enabled %s socket \n", platform)
```
fgiorgetti: I believe it should verify whether or not the system-controller container is already defined,
before trying to create the container, or if the container already exists, it should just show a
message and exit.

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +36,179 @@ func Install(platform string) error {
...
+
+	sysControllerContainer := container.Container{
+		Name:        fmt.Sprintf("%s-skupper-system-controller", config.username),
+		Image:       images.GetSystemControllerImageName(),
+		Env:         env,
+		Mounts:      mounts,
+		Annotations: annotations,
+	}
+
+	err = cli.ContainerCreate(&sysControllerContainer)
```
fgiorgetti, line 126: I noticed that if you use `--platform docker` it is not working, as it fails with:
```
failed to configure the environment : Docker installation must be run as root user.
```

But if the docker endpoint it already active, it should try to create the container.
And again, if the container already exists (created with podman, for example), I believe
it should just show a warning as only one instance of the system controller is expected
to run, regardless of the target platform.

internal/nonkube/bootstrap/uninstall.go
```diff
@@ -1,14 +1,44 @@
...
+		return fmt.Errorf("failed to create container client: %v", err)
+	}
+
+	list, err := cli.ContainerList()
+	if err != nil {
+		return err
+	}
+	var systemControllerContainer string
+	for _, container := range list {
+		if strings.HasSuffix(container.Name, "system-controller") {
```
fgiorgetti: Why not using an exact match, based on the expected container name?

internal/nonkube/bootstrap/uninstall.go
```diff
@@ -1,14 +1,44 @@
...
+			systemControllerContainer = container.Name
+			break
+		}
+	}
+
+	err = cli.ContainerStop(systemControllerContainer)
```
fgiorgetti: In the uninstall command, I believe it could also verify if the container exists, before trying to remove it.
If it does not exist, then just keep the existing behavior.

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +36,179 @@ func Install(platform string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create system-controller container: %v", err)
+	}
+	err = cli.ContainerStart(sysControllerContainer.Name)
+	if err != nil {
+		return fmt.Errorf("failed to start system-controller container: %v", err)
+	}
+
```
fgiorgetti, line 134: One thing that is missing to me, is the creation of a systemd service, to control the lifecycle of the 
system controller, and to ensure it will be active after a reboot. That also implies, the removal of the
same service at the uninstall.

internal/nonkube/bootstrap/install.go
nluaces: removed from the list of env. variables, thanks!

internal/nonkube/bootstrap/install.go
nluaces: I missed this! thanks!

internal/nonkube/bootstrap/uninstall.go
nluaces: done

internal/nonkube/bootstrap/uninstall.go
nluaces: done

internal/nonkube/bootstrap/install.go
nluaces: done

internal/nonkube/bootstrap/install.go
nluaces: because docker requires root permissions, I was testing it with: 
`sudo -E skupper system install -p docker`

the thing is that installing the controller with root, makes the system controller container have the name `root-skupper-system-controller`. 

I can modify the functionality to check if there is a container with a system controller image (instead of checking just the name) running. Thoughts?

fgiorgetti: Actually you don't need to be root, I suppose. Instead your user must be part of the docker group.
And running as root is not a solution, because the system-controller running as root will only watch
for rootful namespaces, i.e.: `/var/lib/skupper/namespaces`.

So the important thing is to verify if the current user has permission to create a container, if not, then
we can error out saying that user cannot run docker containers. WDYT?

nluaces: fixed to allow to install a system container in a docker platform without being root.

internal/nonkube/bootstrap/install.go
nluaces: done!

pkg/nonkube/api/environment.go
```diff
@@ -188,3 +188,13 @@ func GetInternalOutputPath(namespace string, internalPath InternalPath) string {
 func GetInternalBundleOutputPath(namespace string, internalPath InternalPath) string {
 	return path.Join(GetDefaultBundleOutputPath(namespace), string(internalPath))
 }
+
+func GetSystemControllerPath() string {
+	dataHome := GetSystemConfigHome()
```
fgiorgetti: Instead of calling `GetSystemConfigHome` you should use `GetHostDataHome`.
This already handles the `rootful` case, as well as when this code is running in a container,
which will be needed once we add the `install` command to the `skupper.v2` Ansible collection.

internal/nonkube/bootstrap/controller/stopsh-container.template
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+set -o errexit
+set -o nounset
+
+{{.ContainerEngine}} stop -t 10 {{.Name}}
```
fgiorgetti: Minor: add new line to the end of file.

internal/nonkube/bootstrap/controller/systemd.go
```diff
@@ -0,0 +1,254 @@
...
+	rootSystemdBasePath string
+	ScriptPath          string
+}
+
+func NewSystemdServiceInfo(systemContainer container.Container, platform string) (SystemdService, error) {
+
+	scriptPath := path.Join(api.GetSystemControllerPath(), "internal", "scripts")
+
+	return &systemdServiceInfo{
+		Name:                systemContainer.Name,
```
fgiorgetti: @nluaces I was thinking that the service could be named just `skupper-controller`.
The container should still be prefixed with the username as you have it.

The reason for this is because today all services are named as: `skupper-<namespace>` so the
common prefix to the service names is `skupper`. WDYT?

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +39,261 @@ func Install(platform string) error {
...
 
+	config, err := configEnvVariables(platform)
+	if err != nil {
+		return err
+	}
+
+	containerName := fmt.Sprintf("%s-skupper-controller", config.username)
+
+	isContainerAlreadyRunningInPodman, err := IsContainerRunning(containerName, types.PlatformPodman)
+	if err != nil {
```
fgiorgetti: This is assuming that both podman and docker are installed on the system, which might not always be true.
Maybe safer to ignore the error if either podman or docker is not available?

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +39,261 @@ func Install(platform string) error {
...
+	if err := os.MkdirAll(api.GetDefaultOutputNamespacesPath(), 0755); err != nil {
+		return nil, fmt.Errorf("Failed to create directory: %v", err)
+	}
+
+	containerEndpoint := os.Getenv("CONTAINER_ENDPOINT")
+	if containerEndpoint == "" {
+		containerEndpoint = containerEndpointDefault
+	}
+
+	err = os.Setenv("CONTAINER_ENDPOINT_DEFAULT", containerEndpointDefault)
```
fgiorgetti: Is it needed to set the env vars: CONTAINER_ENDPOINT_DEFAULT, XDG_DATA_HOME and SKUPPER_OUTPUT_PATH?
I believe these are useful when running the `install` command as a container through a shell script. 
Maybe these are not needed in the go code.

internal/nonkube/bootstrap/install.go
```diff
@@ -16,5 +39,261 @@ func Install(platform string) error {
...
+	controllerConfig.containerEndpoint = containerEndpoint
+
+	return &controllerConfig, nil
+}
+
+func createSystemdService(container container.Container, platform string) error {
+
+	// Creating startup scripts
+	startupArgs := controller.StartupScriptsArgs{
+		Name:     container.Name,
```
fgiorgetti, line 227: My thought is that we could name the service as `skupper-controller` (keeping container name as is).

internal/nonkube/bootstrap/uninstall.go
```diff
@@ -1,14 +1,81 @@
...
 func Uninstall(platform string) error {
 
+	currentUser, err := user.Current()
+	if err != nil {
+		return fmt.Errorf("failed to get current user: %v", err)
+	}
+
+	containerName := fmt.Sprintf("%s-skupper-controller", currentUser.Username)
+
+	isContainerAlreadyRunningInPodman, err := IsContainerRunning(containerName, types.PlatformPodman)
```
fgiorgetti: Similarly, we should ignore errors in case platform is not available (as docker and podman might now always be installed on the system).

internal/nonkube/bootstrap/controller/stopsh-container.template
nluaces: done

internal/nonkube/bootstrap/controller/systemd.go
nluaces: agreed!

internal/nonkube/bootstrap/install.go
nluaces: done

internal/nonkube/bootstrap/install.go
nluaces: done

internal/nonkube/bootstrap/uninstall.go
nluaces: done

internal/nonkube/bootstrap/install.go
nluaces: I think they are:

```
2025/07/10 14:53:00 Version: v2-dev-release-176-g4d059a44
2025/07/10 14:53:00 Skupper System Controller watching /app/.local/share/skupper/namespaces
2025/07/10 14:53:00 Host path /app/.local/share/skupper/namespaces
2025/07/10 14:53:00 Error creating skupper namespaces directory "/app/.local/share/skupper/namespaces": mkdir /app/.local: permission denied
2025/07/10 14:53:01 Version: v2-dev-release-176-g4d059a44
2025/07/10 14:53:01 Skupper System Controller watching /app/.local/share/skupper/namespaces
2025/07/10 14:53:01 Host path /app/.local/share/skupper/namespaces
2025/07/10 14:53:01 Error creating skupper namespaces directory "/app/.local/share/skupper/namespaces": mkdir /app/.local: permission denied
```
When starting the container, it is not watching the external namespaces directory by default.

pkg/nonkube/api/environment.go
nluaces: By doing that change is not working as expected:
```
Jul 10 16:53:01 bash[94575]: time="2025-07-10T16:53:01+02:00" level=warning msg="Requested oom_score_adj=100 is lower than the current one, changing to 200"
Jul 10 16:53:01 podman[94575]: 2025-07-10 16:53:01.328635365 +0200 CEST m=+0.071893630 container init 2fb4e5b9ae67a65b563c1de9208d857f157838e8e38c8a70696ae928188a>
Jul 10 16:53:01 podman[94575]: 2025-07-10 16:53:01.331574946 +0200 CEST m=+0.074833207 container start 2fb4e5b9ae67a65b563c1de9208d857f157838e8e38c8a70696ae928188>
Jul 10 16:53:01 bash[94575]: nluaces-skupper-controller
Jul 10 16:53:01 nluaces-skupper-controller[94598]: 2025/07/10 14:53:01 Version: v2-dev-release-176-g4d059a44
Jul 10 16:53:01 nluaces-skupper-controller[94598]: 2025/07/10 14:53:01 Skupper System Controller watching /app/.local/share/skupper/namespaces
Jul 10 16:53:01 nluaces-skupper-controller[94598]: 2025/07/10 14:53:01 Host path /app/.local/share/skupper/namespaces
Jul 10 16:53:01 nluaces-skupper-controller[94598]: 2025/07/10 14:53:01 Error creating skupper namespaces directory "/app/.local/share/skupper/namespaces": mkdir />
Jul 10 16:53:01 podman[94613]: 2025-07-10 16:53:01.356032682 +0200 CEST m=+0.012858218 container died 2fb4e5b9ae67a65b563c1de9208d857f157838e8e38c8a70696ae928188a>
Jul 10 16:53:01 podman[94613]: 2025-07-10 16:53:01.389921564 +0200 CEST m=+0.046747095 container cleanup 2fb4e5b9ae67a65b563c1de9208d857f157838e8e38c8a70696ae9281>
```
This error comes from the systemd service status.

fgiorgetti: The logs above suggest that something else is wrong, as it should not be watching namespaces under:
`/app/.local/share/skupper/namespaces`.
Maybe for some unrelated reason the `SKUPPER_OUTPUT_PATH` was not set properly in the container?

internal/nonkube/bootstrap/install.go
fgiorgetti: Oh, I mean, it is definitely needed inside the new controller container, but not sure if those env vars should be
set on the host machine while running `skupper system install`.

pkg/nonkube/api/environment.go
nluaces: you were right! fixed

internal/nonkube/bootstrap/install.go
nluaces: got it! fixed

internal/nonkube/bootstrap/controller/systemd.go
```diff
@@ -130,7 +130,7 @@ func (s *systemdServiceInfo) GetServiceFile() string {
 	if s.GetUid() == 0 {
 		return path.Join(s.rootSystemdBasePath, s.GetServiceName())
 	}
-	return path.Join(api.GetSystemConfigHome(), "systemd/user", s.GetServiceName())
+	return path.Join(api.GetHostDataHome(), "systemd/user", s.GetServiceName())
```
fgiorgetti: I believe that this function should return something similar to what `internal/nonkube/common/systemd.go` returns on rootless cases. Example:

```
	if s.GetUid() == 0 {
		return path.Join(s.rootSystemdBasePath, s.GetServiceName())
	}
	return path.Join(api.GetConfigHome(), "systemd/user", s.GetServiceName())
```

As the intention is to get the config home to place the `.service` file.
And this won't be useful if the `install` command itself is running in a container.

fgiorgetti (commented):
I believe this is the last change needed @nluaces.
Other than that, it is working great to me with docker, podman using both rootless and rootful modes.

internal/nonkube/bootstrap/controller/systemd.go
nluaces: Now it is getting the GetConfigHome value instead. Thanks!

fgiorgetti: Along with what you have done, I believe you need to get rid of the if block `api.IsRunningInContainer`,
as it will only be used when running outside of a container anyway.

nluaces: done

PR #2150, opened 2025-06-04
Add link delete command for nonkube

lynnemorrison:
Add command to delete links.

The file "Link-<filename>.yaml" is removed from ~/.local/share/skupper/namesapces/<namespace>/input/resources.
When the site is reloaded the link is removed.

Currently on the specified link is removed, so if the user wants to remove more than one link, the user will need to issue this
command for each link to be removed.

### DIFF
--- a/internal/cmd/skupper/link/nonkube/link_delete_test.go
+++ b/internal/cmd/skupper/link/nonkube/link_delete_test.go
@@ -1,1 +1,197 @@
 package nonkube
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
+	"github.com/skupperproject/skupper/internal/cmd/skupper/common/testutils"
+	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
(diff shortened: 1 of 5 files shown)


internal/cmd/skupper/link/nonkube/link_delete_test.go
```diff
@@ -1 +1,197 @@
...
+	}
+
+	// Add a temp file so Link exists for delete tests
+	LinkResource := v2alpha1.Link{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "skupper.io/v2alpha1",
+			Kind:       "Link",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "link-router-east",
```
fgiorgetti: Maybe just rename it to `router-east` to match the expected file name.

fgiorgetti (approved):
LGTM

internal/cmd/skupper/link/nonkube/link_delete_test.go
lynnemorrison: fixed

nluaces (approved):

PR #2151, opened 2025-06-05
Update golang version

nluaces:
- updated to golang 1.23 to update the dependency `golang.org/x/crypto` to v0.31.0
- fixed errors found with `go vet`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/skupperproject/skupper
 
-go 1.19
+go 1.23
```
ajssmith, line 3: suggest go, toolchain as used here:

https://github.com/skupperproject/skupper/blob/96648003e55932aff63fec301e8da8d700b15f20/go.mod#L3

nluaces: You are right, thanks!

PR #2157, opened 2025-06-11
Add check to make sure directory exists before creating token

### LINKED ISSUE #1989: V2 token issue hangs when invalid file provided
**Describe the bug**
Running `skupper token issue /invalid-dir/token.yaml` causes CLI to hang, then
it times out saying that grant not ready yet, in spite of saying the destination file
is invalid.

**How To Reproduce**
Run token issue providing a filename that is supposed to be stored under an invalid directory.

**Expected behavior**
A correct error message is displayed immediately.

**Environment details**
- Skupper CLI - main

lynnemorrison:
When creating a token and user specifies a directory to put the token in, make sure it exists before 
trying to create the token.

Fixes #1989

### DIFF
--- a/internal/cmd/skupper/token/kube/token_issue.go
+++ b/internal/cmd/skupper/token/kube/token_issue.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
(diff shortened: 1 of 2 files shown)


internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -70,9 +71,15 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("token file name is not valid: %s", err))
 		} else {
-			// check we can use as a filename
-			if _, err := os.ReadDir(args[0]); err == nil {
+			directory, filename := filepath.Split(args[0])
+			if filename == "" {
+				validationErrors = append(validationErrors, fmt.Errorf("token file name is not valid"))
+			} else if directory == "/" {
```
fgiorgetti: As long as the filename is not empty (which is the case here), I believe that `/` can be considered as a valid directory.

Maybe a more generic validation would be to use, like, `os.CreateTemp(directory, "skupper-token.*")` to create a temporary file (removing it later) to make sure the current user has permission to write into `directory`, before proceeding with the token generation procedure.

internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -70,9 +71,18 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("token file name is not valid: %s", err))
 		} else {
-			// check we can use as a filename
-			if _, err := os.ReadDir(args[0]); err == nil {
-				validationErrors = append(validationErrors, fmt.Errorf("token file name is a directory"))
+			// check filename and directory are valid to create the token
+			directory, filename := filepath.Split(args[0])
```
fgiorgetti: Before doing it, it would be safe to run `os.Stat(args[0])` and ensure it is not a directory.
As is, if you pass a directory that is writable to the current user, it hangs then throw an
error saying grant is not ready yet.

lynnemorrison: ok

fgiorgetti (approved):
LGTM

PR #2159, opened 2025-06-17
Fixes #2118: Modified GenerateSecret() in certs.go to return an error…

### LINKED ISSUE #2118: Poor Error handling in package internal/certs
**Describe the bug**
The `certs.GenerateSecret` function takes arguments for the properties of a TLS Certificate to generate, including a CA Secret, and returns just a resuling Secret containing the new TLS Certificate. It is not reasonable to expect this operation to always succeed, and we should return an error and handle it appropriately at the caller.

In many cases this function will `log.Fatal` - which will cause the controller to exit.

**Example**
* create a secret in a namesapce with a running skupper Site containing anything but a valid TLS credentials `kubectl create secret generic invalid-ca --from-literal=tls.crt=not --from-literal=tls.key=valid`
* create a v2alpha1.Certificate resource referencing that secret as a ca.
```
cat << EOF | kubectl apply -f -
kind: Certificate
metadata:
  name: my-site-server
spec:
  ca: invalid-ca
  hosts:
  - skupper-router
  - skupper-router.default
  - 172.18.255.113
  server: true
  subject: skupper-router
EOF
```
* See the skupper controller go into a crash loop.

ganeshmurthy:
… instead of log.Fatal() messages. This would ensure that the controller does not exit

### DIFF
--- a/internal/certs/certs.go
+++ b/internal/certs/certs.go
@@ -21,7 +21,6 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
-	"log"
 	"math/big"
 	"net"
 	"strings"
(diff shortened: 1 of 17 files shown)


internal/certs/certs.go
```diff
@@ -57,31 +56,43 @@ type CertificateAuthority struct {
 
 type CertificateData map[string][]byte
 
-func decodeDataElement(in []byte, name string) []byte {
+func decodeDataElement(in []byte, name string) ([]byte, error) {
 	block, _ := pem.Decode(in)
 	if block == nil {
-		log.Fatal("failed to decode PEM block of type " + name)
+		return nil, fmt.Errorf("unable to decode the Data element of the secret %s", name)
```
c-kruse: I've got a series of error message suggestions.

If we walk through and reproduce the issue (as written), the subject Certificate resource ends up with a status containing just this error message text.
```
# kubectl get certificates.skupper.io my-site-server  -ojsonpath="{.status}"
{
  "conditions": [
    {
      "lastTransitionTime": "2025-06-17T22:03:15Z",
      "message": "unable to decode the Data element of the secret certificate",
      "observedGeneration": 1,
      "reason": "Error",
      "status": "False",
      "type": "Ready"
    }
  ]
}
```

Ideally, I think we'll want to see something that could direct a user more directly to the issue. Something like this: `error reading CA Certificate from Secret 'invalid-ca': failed to read PEM encoded data from 'tls.crt'`

c-kruse:
```suggestion
		return nil, fmt.Errorf("failed to read PEM encoded data from %q", name)
```

internal/certs/certs.go
```diff
@@ -57,31 +56,43 @@ type CertificateAuthority struct {
...
 }
 
-func getCAFromSecret(secret *corev1.Secret) *CertificateAuthority {
+func getCAFromSecret(secret *corev1.Secret) (*CertificateAuthority, error) {
 	if secret == nil || secret.Data == nil {
-		return nil
+		return nil, nil
+	}
+
+	certBytes, err := decodeDataElement(secret.Data["tls.crt"], "certificate")
```
c-kruse: alternatively, we could just reorganize decodeDataElement a bit, or wrap the error it returns here. The point is we want it to be obvious exactly what bit of Secret data we're talking about.
```suggestion
	certBytes, err := decodeDataElement(secret.Data["tls.crt"], "tls.crt")
```

internal/certs/certs.go
```diff
@@ -57,31 +56,43 @@ type CertificateAuthority struct {
...
+		return nil, err
+	}
+
+	cert, err := x509.ParseCertificate(certBytes)
+	if err != nil {
+		return nil, err
 	}
-	cert, err := x509.ParseCertificate(decodeDataElement(secret.Data["tls.crt"], "certificate"))
+
+	privateKeyBytes, err := decodeDataElement(secret.Data["tls.key"], "private key")
```
c-kruse:
```suggestion
	privateKeyBytes, err := decodeDataElement(secret.Data["tls.key"], "tls.key")
```

internal/certs/certs.go
```diff
@@ -90,12 +101,17 @@ func getCAFromSecret(secret *corev1.Secret) *CertificateAuthority {
 // hosts are the host names in the x509 certificate. Comma separated if more than one hostname
 // expiration is when the secret expires, if zero is passed in, the expiration is set to 5 years from now
 // ca is the certificate authority, if nil a ca cert will be created.
-func GenerateSecret(name string, subject string, hosts string, expiration time.Duration, ca *corev1.Secret) corev1.Secret {
-	caCert := getCAFromSecret(ca)
+func GenerateSecret(name string, subject string, hosts string, expiration time.Duration, ca *corev1.Secret) (*corev1.Secret, error) {
+	caCert, err := getCAFromSecret(ca)
+
+	if err != nil {
+		return nil, err
```
c-kruse:
```suggestion
		return nil, fmt.Errorf("error reading CA Certificate from Secret %q: %s", ca.Name, err)
```

nluaces (approved):
LGTM. I suggest to wait for Christian's approval.

PR #2165, opened 2025-06-23
Fixes #2161 - Properly protect only the initial-config listeners.

### LINKED ISSUE #2161: RouterAccess cannot create "normal" ingresses
**Describe the bug**
RouterAccess only honors "inter-router" and "edge" roles for router listeners.  It should also be usable for "normal" role listeners/ingresses.

**How To Reproduce**
Apply the following yaml to a running V2 site:
```
kind: RouterAccess
metadata:
  name: test-access
spec:
  generateTlsCredentials: true
  issuer: skupper-site-ca
  roles:
  - name: normal
    port: 20002
  tlsCredentials: skupper-site-server
```

**Expected behavior**
I expect an ingress to be created for test-access and the router configuration to be updated with a normal-mode listener on port 20002.

The ingress _is_ created, but the listener is not.

**Environment details**
 - Skupper CLI: N/A
 - Skupper Operator (if applicable): main branch (6a81e6f62c06217642aa5172b7a4b6408d208aa4)
 - Platform: Openshift

**Additional context**
None.

ted-ross:
The old code assumed that all normal-role listeners come from the initial configuration. This does not allow for the possibility of creating normal-role RouterAccess resources. This fix makes "protected" status more explicit for listeners.

### DIFF
--- a/internal/qdr/qdr.go
+++ b/internal/qdr/qdr.go
@@ -824,8 +824,14 @@ func (r *RouterConfig) UpdateConfigMap(configmap *corev1.ConfigMap) (bool, error
 
 type ListenerPredicate func(Listener) bool
 
-func IsNotNormalListener(l Listener) bool {
-	return l.Role != "normal" && l.Role != ""
+func IsNotProtectedListener(l Listener) bool {
+	protectedNames := [3]string{"@9090", "amqp", "amqps"}
+	for _, name := range protectedNames {
+		if l.Name == name {
+			return false
+		}
+	}
+	return true
 }
 
 func FilterListeners(in map[string]Listener, predicate ListenerPredicate) map[string]Listener {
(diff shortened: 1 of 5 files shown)


internal/qdr/qdr.go
```diff
@@ -824,8 +824,14 @@ func (r *RouterConfig) UpdateConfigMap(configmap *corev1.ConfigMap) (bool, error
 
 type ListenerPredicate func(Listener) bool
 
-func IsNotNormalListener(l Listener) bool {
-	return l.Role != "normal" && l.Role != ""
+func IsNotProtectedListener(l Listener) bool {
```
ganeshmurthy, line 827:
```suggestion
func IsProtectedListener(l Listener) bool {
```

ganeshmurthy: The function name `IsProtectedListener` makes it much easier for me to understand the code. wdyt ?

ted-ross: I agree, but that function is used in a filter that includes list objects in which the filter function returns true.  I did not change the "polarity" of the original function.

ganeshmurthy (approved):

Issue #2047, opened 2025-04-05
`skupper version` shows no information if the cluster is unreacheable


hash-d:
**Describe the bug**

If `$KUBECONFIG` is set (or `~/.kube/config` exists) and its configured cluster is inaccessible, the `skupper version` command returns an empty output.

```
$ skupper version
COMPONENT	VERSION
$
```

**How To Reproduce**

Pick an existing kubeconfig file and change its `server` key to a closed port on `localhost`, such as `1234`.

Then execute the `skupper version` command and inspect the output.

```
$ skupper version
COMPONENT	VERSION
$ grep server $KUBECONFIG 
    server: https://localhost:1234
```

**Expected behavior**

This needs to be worked in light of #2019:

- if the intention is to show the running versions, an error message should be shown stating that the operation cannot be completed, as the cluster is not accessible.
- for baked-in images and environment variables, the command should return its information, in full, as it does not depend on the cluster.

**Environment details**
 - Skupper CLI: 2.0.0
 - Skupper Operator (if applicable): Not applicable
 - Platform: Not applicable

**Additional context**
Not applicable

Issue #2019, opened 2025-03-20
More explicit output in the version command


nluaces:
**Is your feature request related to a problem? Please describe.**
The version command is doing a few things: 
- Showing which image should to be used when configuring a site (default behaviour).
- Showing which image is configured in with ENV variables.
- Reporting what image is actually running.

**Describe the solution you'd like**
It would be clearer to show which of the previous three situations is actually being shown in the output.

PR #2167, opened 2025-06-23
If cluster doesn't exist return an error message.

lynnemorrison:
Prints and error message instead of just printing headers with no data.

 fixes issue #2047

### DIFF
--- a/internal/cmd/skupper/version/kube/version.go
+++ b/internal/cmd/skupper/version/kube/version.go
@@ -69,13 +69,11 @@ func (cmd *CmdVersion) InputToOptions() {
 	if cmd.KubeClient != nil {
 		// search for running pods in the current namespace
 		runningPodList, err := cmd.KubeClient.CoreV1().Pods(cmd.namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app.kubernetes.io/part-of in (skupper, skupper-network-observer)"})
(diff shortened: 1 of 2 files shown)


nluaces:
It shouldn't show an error, it should show the version of the images that we have by default.

lynnemorrison:
from Issue #2047 it says:

"1. if the intention is to show the running versions, an error message should be shown stating that the operation cannot be completed, as the cluster is not accessible.
2. for baked-in images and environment variables, the command should return its information, in full, as it does not depend on the cluster."

So that is why I changed to an error message.    I will change to just showing the default instead if that is the better solution.

nluaces:
@lynnemorrison the current version command is showing default images if skupper does not find a running container to which retrieve its version. 

These images are being shown as the recommended versions that work with the skupper CLI version (examples: network-observer and its dependencies).

With other platforms that does not require a cluster like podman, the version command shows the default version (if I remember correctly).

In the meantime, I suggest to show the default versions if there is no cluster running for kubernetes environments.

How to show what type of version is an issue that still needs definition https://github.com/skupperproject/skupper/issues/2019

and should be done in a different pull request.

lynnemorrison:
@nluaces Thanks for the clarification, will do!!

nluaces (approved):

PR #2170, opened 2025-06-26
Fixes #2050: Use the scratch image instead of ubi9-minimal in order t…

### LINKED ISSUE #2050: Reduce control plane image sizes

ganeshmurthy:
…o reduce the size of the images

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -13,17 +13,19 @@ COPY . .
 ENV CGO_ENABLED=0
 RUN make build-cli GOARCH=$TARGETARCH
 
-FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9-minimal
+# Use scratch for minimal image size
+FROM scratch
+
 LABEL \
   org.opencontainers.image.title="Skupper CLI" \
   org.opencontainers.image.description="Provide users with the ability to operate their Skupper resources."
 
-# Create user and group and switch to user's context
-RUN microdnf -y install shadow-utils \
-&& microdnf clean all
-RUN useradd --uid 10000 runner
+WORKDIR /app
+
+# Copy the statically linked binary
+COPY --from=builder /go/src/app/skupper .
+
+# Use numeric user ID (no need to create user in scratch)
 USER 10000
 
-WORKDIR /app
-COPY --from=builder /go/src/app/skupper .
 ENTRYPOINT ["/app/skupper"]
--- a/Dockerfile.controller
+++ b/Dockerfile.controller
@@ -10,20 +10,23 @@ COPY go.sum .
 RUN go mod download
 COPY . .
 
+# Build statically linked binary
 ENV CGO_ENABLED=0
 RUN make build-controller GOARCH=$TARGETARCH
 
-FROM --platform=$TARGETPLATFORM registry.access.redhat.com/ubi9-minimal
+# Use scratch for minimal image size
+FROM scratch
+
 LABEL \
   org.opencontainers.image.title="Skupper Controller" \
   org.opencontainers.image.description="Kubernetes controller for operating skupper networks"
 
-# Create user and group and switch to user's context
-RUN microdnf -y install shadow-utils \
-&& microdnf clean all
-RUN useradd --uid 10000 runner
+WORKDIR /app
+
+# Copy the statically linked binary
+COPY --from=builder /go/src/app/controller .
+
+# Use numeric user ID (no need to create user in scratch)
 USER 10000
 
-WORKDIR /app
-COPY --from=builder /go/src/app/controller .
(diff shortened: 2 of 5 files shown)


Dockerfile.controller
```diff
@@ -8,20 +8,55 @@ COPY go.sum .
 RUN go mod download
 COPY . .
 
+# Build statically linked binary
```
skitt, line 13: It’s not statically linked if it needs dynamic libraries ;-).

Dockerfile.cli
```diff
@@ -11,17 +11,51 @@ COPY . .
...
+            lib_path=$(echo "$line" | awk '{print $1}'); \
+        fi; \
+        # Copy the file into the runtime-deps only if path is valid
+        if [ -f "$lib_path" ] && [ "$lib_path" != "" ]; then \
+            dest_dir="/runtime-deps$(dirname "$lib_path")"; \
+            mkdir -p "$dest_dir"; \
+            cp "$lib_path" "$dest_dir/"; \
+            echo "Copied: $lib_path"; \
+        fi; \
+    done
```
skitt: You can rely on the fact that relevant entries are absolute paths:

```suggestion
RUN mkdir -p /runtime-deps && \
    ldd /go/src/app/skupper | grep -o '/\S*' | while IFS= read -r lib_path; do \
        dest_dir="/runtime-deps/${lib_path%/*}"; \
        mkdir -p "$dest_dir"; \
        cp "$lib_path" "$dest_dir"; \
        echo "Copied: $lib_path"; \
    done
```

skitt:
> > The extraction process can be simplified, see the suggestion. I’m wary of taking things this far however because this will produce an image where scanners will only recognise the Go binary; vulnerabilities in the libraries will go undetected.
> 
> Thank you Stephen for looking into this. You make a very good point that Clair scan (or any other scanner) will scan only the go binary and might not find CVEs that might actually affect the binary. I had not thought about that. A better approach would be to also copy the dependencies. I will look into this. Thanks again.

Yup, you could look for the packages that ship the required libraries and only install those (and their dependencies).

Dockerfile.controller
c-kruse: @ganeshmurthy - I believe it IS statically linked

```
$ podman run quay.io/skupper/controller:2.0.0  ldd /app/controller
        not a dynamic executable
```

ganeshmurthy: > @ganeshmurthy - I believe it IS statically linked
> 
> ```
> $ podman run quay.io/skupper/controller:2.0.0  ldd /app/controller
>         not a dynamic executable
> ```

@c-kruse In this PR, if you comment the code that copies over the libraries into the /runtime-deps folder and then deploy the controller, does it run  ?

c-kruse: Yes. Commented out `52: COPY --from=builder /runtime-deps /` here and it started right up - ran a site, etc.

ganeshmurthy: > Yes. Commented out `52: COPY --from=builder /runtime-deps /` here and it started right up - ran a site, etc.

Darn, if I did not copy over that folder, I kept running into a problem that said executable not found. Well, if the container is working normally, it is perfect. I will clean the PR up shortly and see what the CI does. Thank you very much !

ganeshmurthy:
> @ganeshmurthy you might as well drop the commits dealing with `/runtime-deps` entirely…

I will squash all the commits so any commits that create/delete the /runtime-deps will be gone.

BTW, @c-kruse noted that the reason I needed to copy the dependent libs was because of the issue with 'jq' which you fixed, thanks for that as well !

skitt (approved):

ganeshmurthy:
It looks like the end to end test has failed. Can you @granzoto please take a look and let me know if the failure is due to this PR ? Thanks.

nluaces:
> I'll leave it to @fgiorgetti and @nluaces - who may know better if we expect a shell or some other non-skupper executables in these containers somewhere like the system commands or skupper-ansible or something.

I think this concern is valid.

lynnemorrison:
@fgiorgetti podman containers.  Right now I'm just using the   pkg/container/client.go
running ContainerInspect and ContainerLogs.   I have not got the containerExec workig for running the skstat commands.

fgiorgetti:
> @fgiorgetti podman containers. Right now I'm just using the pkg/container/client.go running ContainerInspect and ContainerLogs. I have not got the containerExec workig for running the skstat commands.

If the goal is just to run skstat and skmanage against the router container, I believe we should be in good shape.

PR #2173, opened 2025-07-07
Improve until clauses in e2e tests

granzoto:


tests/e2e/scenarios/expose-pods-by-name/test.yml
```diff
@@ -233,7 +252,9 @@
             label_selectors:
               - internal.skupper.io/listener = true
           register: svc_list
-          until: svc_list.resources | length > 2
+          until:
```
c-kruse, line 296: Where I end up lost for what to do is around here.

So far, the test starts a deployment, waits for pods to be ready, wait for N services to get created (though expose pods by name), and compares pod names with service names. :heavy_check_mark:.

Then the test restarts the deployment and waits for the new replicaset is started. :heavy_check_mark:. Here we wait for the number of services to be what we expect at steady state - which seems to me likely is begging non-deterministic behavior. :x:.

I was thinking we may need something like "service list length is expected AND service list does not contain old pod/service names".

Thoughts?

tests/e2e/scenarios/expose-pods-by-name/test.yml
```diff
@@ -87,7 +92,7 @@
             pod_names_array: "{{ pod_list.resources | map(attribute='metadata.name') | list }}"
           when:
             - pod_list.resources is defined
-            - pod_list.resources | length > 0
+            - pod_list.resources | length == 3
```
fgiorgetti: This task should not move on if the length is != 3, so what do you think about
moving the `pod_list.resources` assertion into a `failed_when` so that if there
are not 3 pods, it fails. This will help test failing fast, if some unexpected
condition happens.

Another benefit is that you don't need to check if `pod_names_array` is defined
on further tasks.

tests/e2e/scenarios/expose-pods-by-name/test.yml
```diff
@@ -107,7 +112,9 @@
             label_selectors:
               - internal.skupper.io/listener = true
           register: svc_list
-          until: svc_list.resources | length > 2
+          until:
+            - svc_list.resources is defined
+            - svc_list.resources | length == 4
```
fgiorgetti, line 120: Same thing here. Maybe you could just fail the task if length != 4, then you will
also fail fast and won't need to assert that `svc_names_array` is defined on other
tasks.

tests/e2e/scenarios/expose-pods-by-name/test.yml
```diff
@@ -202,7 +215,13 @@
             label_selectors:
               - app = backend
           register: pod_list
-          until: pod_list.resources | length > 0 and  pod_list.resources[0].status.phase == "Running"
+          until:
+            - pod_list is defined
+            - pod_list.resources is defined
+            - pod_list.resources | length > 0
```
fgiorgetti: Should this be `length > 0` or `length == X`, where X is the exact number of pods?

tests/e2e/scenarios/expose-pods-by-name/test.yml
```diff
@@ -244,7 +265,7 @@
             svc_names_array: "{{ svc_list.resources | selectattr('metadata.name', 'search', search_character) | map(attribute='metadata.name') | list }}"
           when:
             - svc_list.resources is defined
-            - svc_list.resources | length > 0
+            - svc_list.resources | length == 4
```
fgiorgetti, line 309: Minor suggestion here. I believe that this `when` condition is not needed here, as it is the condition
used in the `until` option of the previous task, so if this condition was not met earlier, the previous
task has already failed.

tests/e2e/scenarios/expose-pods-by-name/test.yml
fgiorgetti: @c-kruse I think I got your point here. Do you mean that the code is currently comparing the current pod names found on east with the svc names found on west, but it is not comparing the old/new pod and service names between themselves?

c-kruse: That's about right! We are (re)capturing the svc names at a point where the list of service names is the correct length with no guarantee that the skupper controller has completed processing the adding/removing of services to match the pods in east.

The order the controller processes events is non-deterministic. So it could end up doing something like this: add service (5), delete service (4), add service (5), delete service (4)... - so this step could end up just not waiting long enough.

tests/e2e/scenarios/expose-pods-by-name/test.yml
granzoto: Ok, failing if number os pods is not the expected

tests/e2e/scenarios/expose-pods-by-name/test.yml
granzoto: moved the check to a previous task

tests/e2e/scenarios/expose-pods-by-name/test.yml
granzoto: moved the check to a previous task

tests/e2e/scenarios/expose-pods-by-name/test.yml
granzoto: Ok, I added two additional checks to the code:
1 - Ensure that the new-pods names are different than the old/previous pods
2 - Ensure that all the old pods were removed after the restart

tests/e2e/scenarios/expose-pods-by-name/test.yml
granzoto: Moved to the exact number of pods

fgiorgetti (approved):
Nice job.

PR #2174, opened 2025-07-07
initial debug dump implementation for nonkube

lynnemorrison:
Initial implementation of debug dump command for nonkube.   
Missing skstat commands for skupper-router containers.


lynnemorrison:
@ajssmith @fgiorget when you get a chance can you take a look at what I have so far.  Please let me know what other items you would like to see in the dump

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+	cmd.secretHandler = fs.NewSecretHandler(cmd.namespace)
+	cmd.configMapHandler = fs.NewConfigMapHandler(cmd.namespace)
+}
+
+func (cmd *CmdDebug) ValidateInput(args []string) error {
+	var validationErrors []error
+	fileStringValidator := validator.NewFilePathStringValidator()
+
+	// Validate dump file name
+	if len(args) < 1 {
```
fgiorgetti, line 66: I believe that the presence of the provided namespace should be validated here.
Today, if you provide a namespace that does not exist, or even if you provide nothing and the 
default namespace does not exist, it tries to dump it.

fgiorgetti: Another problem is that if a namespace is not provided, the file is generated as:
`skupper-dump--20250716180558.tar.gz`

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+		pv, err := utils.RunCommand("skrouterd", "--version")
+		if err == nil {
+			utils.WriteTar("/versions/skrouterd.txt", pv, time.Now(), tw)
+		}
+	}
+
+	manifest, err := utils.RunCommand("skupper", "version", "-o", "yaml", "-n", cmd.namespace)
+	if err == nil {
+		utils.WriteTar("/versions/skupper.yaml", manifest, time.Now(), tw)
+		utils.WriteTar("/versions/skupper.yaml.txt", manifest, time.Now(), tw)
```
fgiorgetti, line 141: Why are we storing both yaml and txt versions of the same content?
I believe it is to give users the ability to open the file from a browser, is it right (just curious)?

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+
+	// compress tar
+	gz := gzip.NewWriter(tarFile)
+	defer gz.Close()
+	tw := tar.NewWriter(gz)
+	defer tw.Close()
+
+	platform := config.GetPlatform()
+
+	if platform == types.PlatformPodman {
```
fgiorgetti: Maybe we could include all of the versions (podman, docker and skrouterd)?
Since you're already ignoring errors, it should be a good thing to include imho.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+	}
+
+	rpath := "/runtime/"
+	inpath := "/input/"
+	intpath := "/internal/"
+	certFiles := []string{"ca.crt", "tls.crt"}
+
+	//input/certs
+	path := inpath + "certs/"
+	certPath := api.GetDefaultOutputPath(cmd.namespace) + "/" + string(api.InputCertificatesPath)
```
fgiorgetti: Not a problem at all, but just a recommendation. You could use this instead:
```go
api.GetInternalOutputPath(cmd.namespace, api.InputCertificatesPath)
```

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+	for x := range certFiles {
+		file, err := os.ReadFile(certPath + "/" + certFiles[x])
```
fgiorgetti: I believe this won't work. The `certs/` directory is supposed to have sub-directories representing
the certificate name and files inside of it as explaing through [PROVIDED_CERTIFICATES.md](https://github.com/skupperproject/skupper/blob/main/cmd/bootstrap/PROVIDED_CERTIFICATES.md#server-and-client-certificates).

My recommendation would be to use `DirectoryReader.ReadDir` which reads all files recursively, based on a
provided filter, so you could have something like:

```go
	dr := &DirectoryReader{}
	files, err := dr.ReadDir(certPath, func(s string) bool {
		return slices.Contains(certFiles, s)
	})
```

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+			if file != nil && err == nil {
+				utils.WriteTar(path+fileName, file, time.Now(), tw)
+			}
+		}
+	}
+
+	// runtime/certs
+	path = rpath + "certs/"
+	certPath = api.GetDefaultOutputPath(cmd.namespace) + "/" + string(api.CertificatesPath)
+	certDirs, err := os.ReadDir(certPath)
```
fgiorgetti: Same recommendation here. Use the `DirectoryReader`.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+
+func (cmd *CmdDebug) Run() error {
+	dumpFile := cmd.fileName
+
+	// Add extension if not present
+	if filepath.Ext(dumpFile) == "" {
+		dumpFile = dumpFile + ".tar.gz"
+	}
+
+	tarFile, err := os.Create(dumpFile)
```
fgiorgetti, line 111: Maybe you could try to use the `internal.utils.Tarball` type, to avoid a dup functionality?

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+					utils.WriteTar(path+fileName, file, time.Now(), tw)
+				}
+			}
+		}
+	}
+
+	// runtime/issuers
+	path = rpath + "issuers/"
+	iPath := api.GetDefaultOutputPath(cmd.namespace) + "/" + string(api.IssuersPath)
+	iDirs, err := os.ReadDir(iPath)
```
fgiorgetti, line 256: Same recommendation here. Use the `DirectoryReader`.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+	}
+
+	// runtime/links
+	path = rpath + "links/"
+	lpath := api.GetDefaultOutputPath(cmd.namespace) + "/" + string(api.RuntimeTokenPath)
+	lopts := fs.GetOptions{ResourcesPath: lpath, RuntimeFirst: true, LogWarning: false}
+	links, err = cmd.linkHandler.List(lopts)
+	if err == nil && links != nil && len(links) != 0 {
+		for _, link := range links {
+			err := utils.WriteObject(link, path+link.Name+link.Spec.Endpoints[0].Host, tw)
```
fgiorgetti: A dash is missing between `link.Name` and `link.Spec.Endpoints[0].Host`.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+		return fmt.Errorf("Unable to save skupper dump details: %w", err)
+	}
+
+	// compress tar
+	gz := gzip.NewWriter(tarFile)
+	defer gz.Close()
+	tw := tar.NewWriter(gz)
+	defer tw.Close()
+
+	platform := config.GetPlatform()
```
fgiorgetti: The platform, for an active namespace, should be loaded using:

```golang
	nsl := &NamespacePlatformLoader{}
	platform, err := nsl.Load("your-namespace")
```

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+	path = rpath + "router/"
+	skrPath := api.GetDefaultOutputPath(cmd.namespace) + "/" + string(api.RouterConfigPath+"/skrouterd.json")
+	//skrPath := pathProvider.GetRuntimeNamespace()+ api.
+	skrouterd, err := os.ReadFile(skrPath)
+	if err == nil && skrouterd != nil {
+		utils.WriteTar(path+"skrouterd.json", skrouterd, time.Now(), tw)
+	}
+
+	//logs and skupper-router statistics
+	cli, err := internalclient.NewCompatClient(os.Getenv("CONTAINER_ENDPOINT"), "")
```
fgiorgetti: Before creating the compat client, set the `SKUPPER_PLATFORM` environment variable, using the `platform` loaded using the `NamespacePlatformLoader`.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +45,355 @@ func NewCmdDebug() *CmdDebug {
...
+				out, err := cli.ContainerExec(rtrContainerName, strings.Split(command, " "))
+				if err == nil {
+					fmt.Println("containerexec: ", err, out)
+				}
+			}
+		*/
+
+		out, err := cli.ContainerExec(rtrContainerName, strings.Split("skstat -c", " "))
+		if err == nil {
+			fmt.Println("containerexec: ", err, out)
```
fgiorgetti: This won't work. As containers are connected to the `host` network, we are using a secure connection
to communicate with the router through a normal listener.

First you need to call `runtime.GetLocalRouterAddress` to determine the effective router address.

Then you can create a skStatCommand variable that includes:

```go
localRouterAddress := runtime.GetLocalRouterAddress("my-namespace")
skStatCommand := []string{
  "-b", localRouterAddress,
  "--ssl-certificate", "/etc/skupper-router/runtime/certs/skupper-local-client/tls.crt",
  "--ssl-key", "/etc/skupper-router/runtime/certs/skupper-local-client/tls.key",
  "--ssl-trustfile", "/etc/skupper-router/runtime/certs/skupper-local-client/ca.crt"
}
```

Then you can use it as a prefix to all skstat commands you're going to run.

internal/cmd/skupper/debug/nonkube/debug_test.go
```diff
@@ -1 +1,355 @@
...
+		{
+			name:  "ok",
+			flags: common.CommandDebugFlags{},
+			args:  []string{"test"},
+		},
+		{
+			name:  "ok default name",
+			flags: common.CommandDebugFlags{},
+			args:  []string{},
+		},
```
fgiorgetti, line 64: Validate if namespace is valid.

internal/cmd/skupper/debug/nonkube/debug_test.go
```diff
@@ -1 +1,355 @@
...
+		namespace string
+		filename  string
+		args      []string
+		flags     common.CommandDebugFlags
+	}
+
+	testTable := []test{
+		{
+			name:      "default name",
+			namespace: "default",
```
fgiorgetti, line 100: Test with an empty namespace too.

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: yes that is what Justin wanted.

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: makes sense,  changing.

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: fixed

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: fixed

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: fixed

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug_test.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug_test.go
lynnemorrison: done

internal/cmd/skupper/debug/nonkube/debug.go
fgiorgetti: This one is still missing.

internal/cmd/skupper/debug/nonkube/debug.go
fgiorgetti: Another thing. The `skstat` is not being executed on `linux` sites.
In that case, instead of executing inside of the container, you have to run the same command directly on the host machine.

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -21,13 +44,372 @@ func NewCmdDebug() *CmdDebug {
...
+				utils.WriteTar(rpath+"logs/"+rtrContainerName+".txt", []byte(logs), time.Now(), tw)
+			}
+
+			ctlContainerName := "system-controller"
+			if container, err := cli.ContainerInspect(ctlContainerName); err == nil {
+				encodedOutput, _ := utils.Encode("yaml", container)
+				utils.WriteTar(rpath+"Container-"+container.Name+".yaml", []byte(encodedOutput), time.Now(), tw)
+			}
+
+			logs, err = cli.ContainerLogs(ctlContainerName)
```
fgiorgetti: If using the `linux` platform, we need to capture logs too.
In that case, you could run something like:

In rootful mode:
```
journalctl -u skupper-<namespace>.service --no-pager --all
```

In rootless mode:
```
journalctl --user -u skupper-<namespace>.service --no-pager --all
```

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: add skstat for linix sites

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: added logs for linux sites

internal/cmd/skupper/debug/nonkube/debug.go
```diff
@@ -363,47 +363,78 @@ func (cmd *CmdDebug) Run() error {
...
-			rtrContainerName := cmd.namespace + "-skupper-router"
-			if container, err := cli.ContainerInspect(rtrContainerName); err == nil {
-				encodedOutput, _ := utils.Encode("yaml", container)
-				utils.WriteTar(rpath+"Container-"+container.Name+".yaml", []byte(encodedOutput), time.Now(), tw)
-			}
-
-			localRouterAddress, err := runtime.GetLocalRouterAddress(cmd.namespace)
+			utils.WriteTar(path+"logs/"+rtrName+".txt", pv, time.Now(), tw)
+		}
+		pv, err = utils.RunCommand("skrouterd", "/bin/skstat", "-c")
```
fgiorgetti: Should this one be `skrouterd -v` ?

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: This is how the kube debug command, unless you feel strongly, I'd just leave it as is.

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: fixed

internal/cmd/skupper/debug/nonkube/debug.go
lynnemorrison: It was accidently left in.   I was just testing before I had the loop below working.

PR #2184, opened 2025-07-18
add missing install

ajssmith:


nluaces (approved):

.circleci/config.yml
```diff
@@ -364,6 +364,8 @@ jobs:
     executor:
       name: go_cimg
     steps:
+      - docker/install-docker
```
nluaces, line 367: maybe we could try to add 
`- setup_remote_docker`
after line 367, to see if with that docker is running.

fgiorgetti: Well spotted.

PR #2191, opened 2025-07-22
Get ROUTER_IMAGE_TAG value from the value set in the Makefile

### LINKED ISSUE #2190: kind-dev-cluster script is getting the wrong ROUTER_IMAGE_TAG
When e2e tests are being executed in a release, the kind-dev-cluster script is not getting the configured router version in the variable `ROUTER_IMAGE_TAG`, and is using instead `SKUPPER_ROUTER_IMAGE` that is not set in the Makefile. 

As a result, the tests are being executed with the skupper controller image that is being released and the main skupper-router version by default.

nluaces:
Fixes #2190


c-kruse (commented):
I am happy with this change.

As noted in the issue, I do not believe that this change alone will close it!

c-kruse (approved):

PR #2198, opened 2025-07-31
Default redemptionsallowed in accessgrant

### LINKED ISSUE #2087: An AccessGrant with the redemptionsAllowed field left empty is useless
**Describe the bug**
If you provide an AccessGrant leaving the redemptionsAllowed field empty, it is accepted and marked ready, but it cannot be used.

**How To Reproduce**
Create an AccessGrant with an empty `spec`.
Then try to generate a token and send it to another site.

**Expected behavior**
A default number of redemptions could be used (i.e. 1) or its status should be set to failure.

**Environment details**
Skupper V2

lynnemorrison:
If accessgrant is received without redemptionsallowed specified, we will default the value to 1.   This will allow the accessgrant
to be functional.

Fixes #2087


internal/kube/grants/grants.go
```diff
@@ -210,8 +210,19 @@ func (g *Grants) checkGrant(key string, grant *skupperv2alpha1.AccessGrant) erro
...
+
 	var err error
 
+	// if RedemptionsAllowed is not set default to 1 so that grant is usable
+	if grant.Spec.RedemptionsAllowed == 0 {
+		grant.Spec.RedemptionsAllowed = 1
+		_, e := g.clients.GetSkupperClient().SkupperV2alpha1().AccessGrants(grant.ObjectMeta.Namespace).Update(context.TODO(), grant, metav1.UpdateOptions{})
+		if e != nil {
+			status = append(status, fmt.Sprintf("Failed updating Redemptions Allowed %s", e))
+		}
```
c-kruse: Suggest that we omit the early update operation here.

I think we want to just update the allowed redemptions value and set changed=true here. That way at the end we do only one update operation (bonus we correctly handle grant state and avoid  the following error.)

> time=2025-08-11T21:16:50.156Z level=INFO msg="[ControllerError] Error while handling AccessGrant west/west-grant: Operation cannot be fulfilled on accessgrants.skupper.io \"west-grant\": the object has been modified; please apply your changes to the latest version and try again"

lynnemorrison: fixed

c-kruse: @lynnemorrison - many apologies. My suggestion did not end up working the way I thought it would. Would approve the previous revision. (I see now that at the end we do an UpdateStatus operation - so redemtions is never updated)

Bonus points if you want to fix that error message. Thinking more carefully about it now, it may make more sense for this to be the first check: i.e. redemptionsallowed needs defaulted? then (UPDATE and return; else continue to do all of the status update checks.

lynnemorrison: changed to your recommendation.   I do not see error log.

c-kruse (approved):
Thanks for bearing with me!

PR #2201, opened 2025-08-01
Add logs for add/delete custom annotations and labels

### LINKED ISSUE #2200: Add log entries for addition and removal of custom annotations and labels
Currently, users don't have a clear way to know exactly when newly added custom annotations and labels were processed by the Skupper controller and applied to their resources,  so they must manually check if the changes have taken effect.

By adding some log entries, we can provide a clear signal to our users, letting them know the exact moment their new labels and annotations are ready to be used. We might also log when custom annotations and labels were removed.

Example of log entries :
time=2025-08-01T14:52:43.654-03:00 level=INFO msg="Found new label and annotation configuration" component=kube.site.labels.registry name=label-annot-all-kinds namespace=label-annot-east
time=2025-08-01T14:53:00.697-03:00 level=INFO msg="Removing label and annotation configuration" component=kube.site.labels.registry name=label-annot-all-kinds namespace=label-annot-east

granzoto:
Fixes #2200

### DIFF
--- a/internal/kube/site/labels/registry.go
+++ b/internal/kube/site/labels/registry.go
@@ -1,6 +1,7 @@
 package labels
 
 import (
+	"log/slog"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -10,6 +11,7 @@ import (
 type LabelsAndAnnotations struct {
 	namespaces          map[string]*Registry
 	controllerNamespace string
+	log                 *slog.Logger
 }
 
 func NewLabelsAndAnnotations(controllerNamespace string) *LabelsAndAnnotations {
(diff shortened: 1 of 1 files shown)


internal/kube/site/labels/registry.go
```diff
@@ -62,18 +71,27 @@ func (l *LabelsAndAnnotations) SetAnnotations(namespace string, name string, kin
...
 	config map[string]*corev1.ConfigMap
+	log    *slog.Logger
 }
 
 func newRegistry() *Registry {
 	return &Registry{
 		config: map[string]*corev1.ConfigMap{},
+		log: slog.New(slog.Default().Handler()).With(
+			slog.String("component", "kube.site.labels.registry"),
+		),
```
c-kruse: I'd just pass the logger in from LabelsAndAnnotations if it's going to be the same instead of constructing a new one. So that'd be like `func newRegistry(log *slog.Logger) *Registry`

granzoto: Ok, reusing the same logger

granzoto:
I moved the "Loading" message to the correct location. Previously it was identifying a new configmap only 1 time

fgiorgetti (approved):
LGTM

Issue #2132, opened 2025-05-19
Controller "Route updated successfully" noise


c-kruse:
**Describe the bug**
The controller logs "Route < namespace >/< route name > updated successfully" several times per minute for each securedaccess with type route.

I believe that it is performing an API operation because the "desired" RouteSpec has .spec.to.weight set to nil, but the actual state contains the defaulted value, .spec.to.weight=100, which casues the comparison logic to indicate there has been an update: https://github.com/skupperproject/skupper/blob/ac97599a90d5cd1ac5ceff5cc1db2121264fc17e/internal/kube/securedaccess/route.go#L139

**How To Reproduce**
Start the skupper controller on openshift with the default settings (grants enabled) and watch it log `Route skupper/skupper-grant-server-https updated successfully`.

**Expected behavior**
The controller should more carefully compare desired and actual state, and only trigger an update when the resource changes.

**Environment details**
skupper 2.0.0
platform openshift

PR #2204, opened 2025-08-04
Specify SecuredAccess Route target weight

c-kruse:
Explicitly sets Route spec.to.weight to 100. Resolves issue #2132


fgiorgetti (approved):
LGTM.
Just for the records, it also works for cases when sites were created without the weight being explicitly set.

PR #2205, opened 2025-08-04
Removes non-canonical CRD specifications

c-kruse:


c-kruse:
Not sure how we ended up with this split. I am pretty confident these have been totally replaced with the crds in config/crds.

nluaces (approved):

fgiorgetti (approved):
That is a good start. Not sure how we ended up with both copies, but I agree that those under `config/crd` represent the latest version.

PR #2206, opened 2025-08-05
Set site status to pending when router is down

fgiorgetti:


lynnemorrison (approved):
tested this fix and it works great.

PR #2208, opened 2025-08-05
Include Documentation in Site CRD

c-kruse:
Adds field descriptions largely copied from refdog documentation.

### DIFF
--- a/config/crd/bases/skupper_site_crd.yaml
+++ b/config/crd/bases/skupper_site_crd.yaml
@@ -80,6 +80,9 @@ spec:
 
                     `routerDataConnections`: Set the number of router worker threads. Minimum 2.
                     `routerLogging`: Set the number of router logging level. Options are "info", "warning", "error".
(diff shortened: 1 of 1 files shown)


config/crd/bases/skupper_site_crd.yaml
```diff
@@ -10,34 +10,99 @@ spec:
...
                 ha:
                   type: boolean
+                  description: |-
+                    Configure the site for high availability (HA). HA sites
+                    have two active routers.
+
+                    Note that Skupper routers are stateless, and they restart
+                    after failure. This already provides a high level of
+                    availability. Enabling HA goes further and reduces the
+                    window of downtime caused by restarts.
```
fgiorgetti, line 61: I know it is not on refdog, but do you think it is worth mentioning that ha = true implies anti-affinity?

fgiorgetti (approved):

pwright:
should defaults be defined outside of descriptions?

PR #2220, opened 2025-08-11
Mark issues and prs as stale, then close them

fgiorgetti:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+    steps:
+      - uses: actions/stale@v9
+        with:
+          exempt-issue-labels: 'bug,enhancement,needs investigation'
+          days-before-issue-stale: 90
```
nluaces: I was thinking that 1 year or 6 months is a period to consider the issue as stale, otherwise this would require for us to at least label properly the issue with `bug`, `needs investigation`, etc, before 90 days, right? 

If it was agreed like so on purpose, I'm not going to oppose.

c-kruse (approved):

nluaces (approved):

ajssmith (approved):

nluaces (approved):

nluaces:
oh! I was wondering why this pr didn't get merged!

PR #2221, opened 2025-08-11
add documentation to attachedconnectorbinding crd

nluaces:


config/crd/bases/skupper_attached_connector_binding_crd.yaml
```diff
@@ -33,10 +47,19 @@ spec:
               type: object
               properties:
                 status:
```
ajssmith, line 49: Suggest adding hasMatchingListener and use text from same in additionalPrinterColumns

config/crd/bases/skupper_attached_connector_binding_crd.yaml
```diff
@@ -33,10 +47,19 @@ spec:
...
+                  description: |-
+                    The current state of the resource.
+                    `Pending`: The resource is being processed.
```
pwright:
```suggestion
                    - `Pending`: The resource is being processed.
```

pwright: Repeat for all items, renders better when displayed to user, eg refdog

pwright (commented):
suggest making unordered list to improve rendering

config/crd/bases/skupper_attached_connector_binding_crd.yaml
nluaces: done

PR #2223, opened 2025-08-11
Update description of a SecuredAccess resource

c-kruse:


config/crd/bases/skupper_secured_access_crd.yaml
```diff
@@ -10,7 +10,14 @@ spec:
       storage: true
       schema:
         openAPIV3Schema:
-          description: "An internal resource used to create secure access to pods"
+          description: |-
+            SecuredAccess is an internal resource used to create secure access to pods.
+
```
pwright, line 15: might be just me, but Skupper needs access to pods, but why would it create an ingress/route?

c-kruse: We could clearly use some help explaining what this thing is! It configures the route or ingress (when access type indicates) because that's the entire point?

This is the core of the logic skupper is using to expose router pods externally: configuring some ingress type, learning about the address by which it will be extremely accessible, then issuing a cert valid for that address. It also gets used for our grants server which needs identical behavior, so it's more generic than just router access.

Again, this is all implementation details and not something useful for users. Not sure what level of detail is appropriate in the docs.

config/crd/bases/skupper_secured_access_crd.yaml
```diff
@@ -10,7 +10,14 @@ spec:
       storage: true
       schema:
         openAPIV3Schema:
-          description: "An internal resource used to create secure access to pods"
+          description: |-
+            SecuredAccess is an internal resource used to create secure access to pods.
```
pwright, line 14: Internal resource for secure access to the site router and grant server. See RouterAccess for user configuration options.

c-kruse: Sorry! Did not love this either
- would prefer leaving it open ended: used for routers and grants servers today, but would get reused for future features if needed.
- referring users to RouterAccess config also seems off: as if the RouterAccess options were applicable to SecuredAccess (they mostly line up quite well). Also we have no good controller docs to point to for grants server configuration options.

PR #2232, opened 2025-08-18
Check that valid site endpoints exist before allowing a token creation

### LINKED ISSUE #2213: CLI: token issue Site "Link Access" validation fails for other access types
**Describe the bug**
The skupper CLI prevents you from issuing tokens when the Site is configured with a linkAccess it does not expect. I understand hiding less common options from CLI users, but I think it would be reasonable to allow this to work.

Skupper has more access types than "loadbalancer" and "route". The default access types configured with the default controller installs are "loadbalancer", "local" and "route". The Site spec linkAccess field adds two options for that: "none" and "default".

**How To Reproduce**
Create a Site with linkAccess="local"
```
kind: Site
metadata:
  name: test
spec:
  linkAccess: local
```

```
$ skupper token issue /dev/null
You must enable link access for this site before you can create a token.
$ echo $?
2
```

**Expected behavior**
One token please.

**Environment details**
 - Skupper CLI: 2.1.0, 2.0.0
 - Skupper Operator (if applicable): 2.1.0, 2.0.0
 - Platform: kubernetes - any flavour

lynnemorrison:
The user created a site with LinkAccess = local, and then tried to create a token with the CLI and it failed.   The CLI expects LinkAccess = default, loadbalancer, or router.  The CLI token issue command has been changed to just verify that the site CR has endpoints, which indicate that the site can be linked to.   A site that has endpoints can issue a token to be used by another site to connect to this site.    

fixes #2213

### DIFF
--- a/internal/cmd/skupper/token/kube/token_issue.go
+++ b/internal/cmd/skupper/token/kube/token_issue.go
@@ -101,7 +101,7 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			ok, siteName = utils.SiteLinkAccessEnabled(siteList, common.LinkAccessTypes)
+			ok, siteName = utils.SiteLinkAccessEndpoints(siteList)
 			if !ok {
 				validationErrors = append(validationErrors, fmt.Errorf("You must enable link access for this site before you can create a token."))
 			} else {
--- a/internal/cmd/skupper/common/utils/status.go
+++ b/internal/cmd/skupper/common/utils/status.go
@@ -1,18 +1,8 @@
 package utils
 
 import (
-	"github.com/skupperproject/skupper/internal/utils/validator"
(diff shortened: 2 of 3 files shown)


internal/cmd/skupper/token/kube/token_issue.go
```diff
@@ -101,7 +101,7 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
-			ok, siteName = utils.SiteLinkAccessEnabled(siteList, common.LinkAccessTypes)
```
nluaces, line 104: If the func 
`SiteLinkAccessEnabled(siteList *v2alpha1.SiteList, linkAccessTypes []string) (bool, string)`
is not used anywhere else, could it be removed?

(same goes for `SiteConfigured(siteList *v2alpha1.SiteList) bool`, not related to this pr, but it is in the same file).

lynnemorrison: makes sense removed

Issue #2166, opened 2025-06-23
RouterAccess resource-deletion does not work


ted-ross:
**Describe the bug**

**How To Reproduce**
Apply the following yaml to a running V2 site:
```
apiVersion: skupper.io/v2alpha1
kind: RouterAccess
metadata:
  name: test-access
spec:
  generateTlsCredentials: true
  issuer: skupper-site-ca
  roles:
  - name: inter-router
    port: 20002
  tlsCredentials: skupper-site-server
```
Observe the creation of the RouterAccess, the Service, and the related ingress.

Now delete the RouterAccess:
```
$ kubectl delete routeraccess test-access
```
Observe that the RouterAccess has been deleted but the Service and the ingress are still in place.

**Expected behavior**
All of the artifacts of the RouterAccess that were created should be removed when the RouterAccess is deleted.

**Environment details**
 - Skupper CLI: N/A
 - Skupper Operator (if applicable): head of main (6a81e6f62c06217642aa5172b7a4b6408d208aa4)
 - Platform: [e.g kubernetes, openshift] Openshift, minikube, etc.

**Additional context**
None.

PR #2233, opened 2025-08-18
Changes SecuredAccess owner to RouterAccess

c-kruse:
Updates RouterAccess logic to set itself as owner of the SecuredAccess it creates. When a RouterAccess is deleted, so will the SecuredAccess and associated Service and ingress resources.

Fixes https://github.com/skupperproject/skupper/issues/2166

### DIFF
--- a/internal/kube/securedaccess/access.go
+++ b/internal/kube/securedaccess/access.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log"
 	"reflect"
-	"slices"
 
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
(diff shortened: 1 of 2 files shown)


internal/kube/securedaccess/access.go
```diff
@@ -639,3 +644,27 @@ func hasSecuredAccessLabel(obj *metav1.ObjectMeta) bool {
...
+	}
+
+	changed := false
+	for i, ref := range meta.OwnerReferences {
+		uid := string(ref.UID)
+		if _, ok := byUID[uid]; ok {
+			delete(byUID, uid)
+			continue
+		}
+		meta.OwnerReferences = slices.Delete(meta.OwnerReferences, i, i+1)
```
fgiorgetti: Probably not a real or valid scenario, but if for any reason `meta.OwnerReferences` has 2 entries to be removed, this is going to crash.

c-kruse: Good catch!

fgiorgetti (approved):
LGTM

PR #2234, opened 2025-08-18
Add validation for namespace in nonkube site create

### LINKED ISSUE #2207: Namespace characters should be validated on system sites
Namespace characters are not being validated for system sites when using the CLI.

Namespace characters should be validated allowing just alphanumeric characters and dashes, like `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`.

lynnemorrison:
When creating a site verify the namespace configured follows proper format.

fixes #2207

### DIFF
--- a/internal/cmd/skupper/connector/nonkube/connector_create_test.go
+++ b/internal/cmd/skupper/connector/nonkube/connector_create_test.go
@@ -15,6 +15,7 @@ import (
 func TestNonKubeCmdConnectorCreate_ValidateInput(t *testing.T) {
 	type test struct {
 		name              string
+		namespace         string
 		args              []string
 		k8sObjects        []runtime.Object
 		skupperObjects    []runtime.Object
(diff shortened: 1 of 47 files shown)


fgiorgetti:
The validation itself looks ok to me, but it has to be applied to all `create`, `delete` and `update` commands in order to close #2207. As is, if someone tries to create a `listener` or any other resource using an invalid namespace, it is still allowed.

fgiorgetti (commented):
Validation seems to work. The validation is still missed in the `token redeem` command.

lynnemorrison:
> Validation seems to work. The validation is still missing in the `token redeem` command.

Sorry about that.   I didn't realize tokens were used in nonkube, I only looked at the token issue command and saw it wasn't implemented and didn't look at the redeem command.

fgiorgetti (approved):

c-kruse (approved):

internal/cmd/skupper/connector/nonkube/connector_create_test.go
```diff
@@ -96,7 +106,6 @@ func TestNonKubeCmdConnectorCreate_ValidateInput(t *testing.T) {
 			flags: &common.CommandConnectorCreateFlags{},
 		},
 		{
-
```
nluaces, line 99: should we have a use case scenario for invalid namespaces?

nluaces (approved):

PR #2235, opened 2025-08-18
Fixes race with Network Status Handler

fgiorgetti:
The NetworkStatusHandler test was failing because the NetworkStatusHandler was expecting
that the configmap was already added, capturing only update events.

But if the path is added to the watcher and the configmap is written in between the startup of
the file watcher and the moment the configmap is flushed to the filesystem, the update event
was not being triggered, and instead it was triggering a create event, which was not captured.


hash-d (approved):
@fgiorgetti, while I have not reviewed the code change (well I have opened it, but I have no idea on context), I have ran the unit tests pointing to your branch 5 times, on an environment where I saw this test fail 6 times in the previous 10 runs.  

There were no failures on this specific test in these runs, and I see no new tests failing after that change.

PR #2244, opened 2025-08-26
Update nonkube uninstall command with force flag

### LINKED ISSUE #2192: skupper system uninstall --force
**Describe the bug**

I thought `skupper system uninstall --force` would clear my environment, removing sites

**How To Reproduce**
$ skupper system uninstall --force
$ podman ps -a
CONTAINER ID  IMAGE                                 COMMAND               CREATED         STATUS                    PORTS                     NAMES
a850c51b6712  quay.io/skupper/skupper-router:3.4.0  /home/skrouterd/b...  49 seconds ago  Up 49 seconds             5671-5672/tcp, 55672/tcp  default-skupper-router

**Expected behavior**

Skupper cli should return a message after performing uninstall (even without --force) outlining the action (or lack of action) taken, eg:

* Controller uninstalled
* Controller not found

Maybe this command should also remove sites?

**Environment details**
 - Skupper CLI: 2.1.0
 - Platform: podman

**Additional context**

lynnemorrison:
Add logs to indicate what is being uninstalled.
Add logs to help user understand when uninstall fails. 
Add code to delete directory structure, and router sites when force flag is set even if sites are active.

fixes #2192

### DIFF
--- a/internal/cmd/skupper/system/nonkube/system_uninstall_test.go
+++ b/internal/cmd/skupper/system/nonkube/system_uninstall_test.go
@@ -2,14 +2,25 @@ package nonkube
 
 import (
 	"fmt"
-	"github.com/skupperproject/skupper/internal/config"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common/testutils"
+	"github.com/skupperproject/skupper/internal/config"
+	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
+	"github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
+	"github.com/skupperproject/skupper/pkg/nonkube/api"
 	"gotest.tools/v3/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+type CmdSiteStatus struct {
+	siteHandler *fs.SiteHandler
+	Flags       *common.CommandSiteStatusFlags
+	namespace   string
+}
 
 func TestCmdSystemUnInstall_ValidateInput(t *testing.T) {
 	type test struct {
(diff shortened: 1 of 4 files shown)


internal/nonkube/bootstrap/uninstall.go
```diff
@@ -94,9 +97,14 @@ func CheckActiveSites() (bool, error) {
 
 	for _, entry := range entries {
 		if entry.IsDir() {
-			return true, nil
+			activeSites = true
```
fgiorgetti: Instead of just a directory entry under `namespaces`, I believe an extra validation is needed, to ensure
that the given **entry** is a directory and inside of it, there is a `runtime` directory as well.
In example: `namespaces/default/runtime`. If the `runtime` directory exists under an entry, then the
given namespace is active.

internal/cmd/skupper/system/nonkube/system_uninstall_test.go
```diff
@@ -122,16 +134,62 @@ func TestCmdSystemUninstall_Run(t *testing.T) {
...
+						Type:   "Configured",
+						Status: "True",
+					},
+				},
+			},
+		},
+	}
+
+	// add site in runtime directory
+	tmpDir := api.GetDataHome()
```
fgiorgetti, line 181: Data Home must be set to a temporary location first, like:
```
	if os.Getuid() == 0 {
		api.DefaultRootDataHome = t.TempDir()
	} else {
		t.Setenv("XDG_DATA_HOME", t.TempDir())
	}
	tmpDir := api.GetDataHome()
```

fgiorgetti (commented):
Another observation is that the `--force` flag description is not precise.
It currently says: `option to override even with sites present`.

I suggest saying clearly that if `--force` is passed, all existing sites (active or not) will be deleted.

@pwright thoughts ?

pwright:
> Another observation is that the `--force` flag description is not precise. It currently says: `option to override even with sites present`.
> 
> I suggest saying clearly that if `--force` is passed, all existing sites (active or not) will be deleted.
> 
> @pwright thoughts ?

`option to remove all site configuration, including active sites` ?

internal/cmd/skupper/system/nonkube/system_uninstall_test.go
lynnemorrison: done

internal/nonkube/bootstrap/uninstall.go
lynnemorrison: done

c-kruse (approved):

PR #2246, opened 2025-08-28
Adding CLI generated documentation

nluaces:


pwright (commented):
looks good, but what's the intention? run and commit each cli change?
run at release time?

would it make more sense to run and commit into skupper-docs? (and use versioning there for each minor release)

nluaces:
> what's the intention? run and commit each cli change?

Yes, that's the idea, to have this CLI documentation available in the main repository.

> would it make more sense to run and commit into skupper-docs? (and use versioning there for each minor release)

I think this is also a very good idea that we could implement additionally.

fgiorgetti (approved):

fgiorgetti:
@nluaces @pwright I like the idea of having this being generated automatically, maybe on each PR merged or as a
scheduled workflow in the CI. Can we get these markdown files incorporated into the skupper-docs somehow?

nluaces:
> Can we get these markdown files incorporated into the skupper-docs somehow?

should we add the changes per release only in the skupper-docs, like @pwright suggested?

pwright:
I would probably add them directly to skupper-website repo, based on release tag

PR #2254, opened 2025-09-08
Prevents race when a Certificate has multiple owners

### LINKED ISSUE #2253: skupper-site-server certificate continuously updated
**Describe the bug**
Observing the skupper controller logs, observe that the skupper-site-server secret is continuously updated with an incrementing ordinal count.

**How To Reproduce**
- Kind cluster
- Apply skupper-cluster-scope.yaml from 2.1.1 release 
- Create ns foo
- Apply skupper-site.yaml

**Expected behavior**
Would expect steady state to be attained after site creation completes.

**Environment details**
- Skupper Controller 2.1.1 
 - Platform: [e.g kubernetes, kind]

[skupper-controller-logs.txt](https://github.com/user-attachments/files/22173005/skupper-controller-logs.txt)

fgiorgetti:
When a Certificate has multiple owner references, a racing condition impacts the certificate manager,
which keeps updating the ownerReferences, as well as the subject and the list of hosts indefinitely.

This fix adds a special handling for cases, like HA, where a certificte has multiple owner references.

Fixes #2253

### DIFF
--- a/internal/kube/certificates/mgr.go
+++ b/internal/kube/certificates/mgr.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/kube/certificates/mgr.go
```diff
@@ -138,15 +139,34 @@ func (m *CertificateManagerImpl) ensure(namespace string, name string, spec skup
...
 		if !cmp.Equal(spec, current.Spec, compareSpecUnordered...) {
 			// merge hosts as the certificate may be shared by sources each requiring different sets of hosts:
 			hosts := getHostChanges(getPreviousHosts(current, refs), spec.Hosts, key).apply(current.Spec.Hosts)
+			originalHosts := current.Spec.Hosts
+			originalSubject := current.Spec.Subject
 			current.Spec = spec
-			current.Spec.Hosts = hosts
-			changed = true
+			ownerRefsLength := len(current.ObjectMeta.OwnerReferences)
+			if ownerRefsLength > 1 {
```
c-kruse: What would you think about reorganizing such that we first operate on `spec`, then use that cmp.Equal condition to set changed? I think it might read a lot clearer - would at least get around some of the originalHosts juggling.

I think that would be something like this:
 * if ownerRefsLength > 1 { spec.Subject = current.Spec.Subject }
 * spec.Hosts = getHostChanges(...).apply(current.Spec.Hosts)
 * if !cmp.Equal(spec, current.Spec, unordered) { current.Spec = spec; changed = true }

fgiorgetti: @c-kruse let me know if it looks better now.
I had to make an extra change, to ensure that the `internal.skupper.io/hosts-<UID>` annotation
was being added, on updates as well.

fgiorgetti: By the way, there is an extra problem because of that, which happens when HA is disabled, as
it is not removing the annotation and the specific hosts for a removed owner reference.
But I believe it is better to handle that through a separate issue (once this one gets merged).

Issue #2252, opened 2025-09-04
Router deployment stuck with failed Pod


c-kruse:
**Describe the bug**
Skupper router deployments can get stuck. I suspect that this is the chain of events that leads to this:

1. A router pod is running, but enters some failure state: in this case the node it was scheduled on was low on memory.
2. The deployment/replicaset creates a replacement router pod.
3. The replacement pod starts up, but the kube-adaptor becomes trapped in a restart cycle when this condition fails (due to the failed pod that was not deleted): https://github.com/skupperproject/skupper/blob/main/cmd/kube-adaptor/main.go#L84-L85

```
[christian@rh skupper] (main)$ k get pods -n ck-hub -l skupper.io/component=router
NAME                              READY   STATUS                   RESTARTS       AGE
skupper-router-6d46f8944d-nnm7x   1/2     CrashLoopBackOff         10 (69s ago)   60m
skupper-router-6d46f8944d-q4x8p   0/2     ContainerStatusUnknown   3 (60m ago)    104m
```

**How To Reproduce**
TBD

**Expected behavior**
Deployment recovers.

**Environment details**
 - Skupper CLI: [e.g. 1.5.0, 1.4.3]
 - Skupper Operator (if applicable): [e.g. 1.5.0, 1.4.3]
 - Platform: [e.g kubernetes, openshift]

**Additional context**

PR #2260, opened 2025-09-12
Changes kube-adaptor startup conditions

c-kruse:
Removes kube-adaptor startup logic that waits for kube-api to have all
router Pods in phase Running. Adds logic to wait for an active skupper
router management endpoint before continuing to handle configuration
events.

Also includes a qdr Agent connection timeout to protect against
transient connection issues blocking the retry loop.

Fixes https://github.com/skupperproject/skupper/issues/2252

### DIFF
--- a/cmd/kube-adaptor/main.go
+++ b/cmd/kube-adaptor/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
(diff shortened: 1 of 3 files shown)


fgiorgetti (approved):
Works great to me.

Issue #2135, opened 2025-05-20
skupper completion subcommand throws errors into stdout


hash-d:
**Describe the bug**

On failure, the `completion` subcommand prints the error and usage to stdout, instead of stderr.

It also does not return a non-zero exit code.

**How To Reproduce**

Just issue an invalid `completion` command, such as missing the shell specification

```
$ skupper completion 2> /dev/null
Generate the autocompletion script for skupper for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  skupper completion [command]

Available Commands:
  bash        Generate the autocompletion script for bash
  fish        Generate the autocompletion script for fish
  powershell  Generate the autocompletion script for powershell
  zsh         Generate the autocompletion script for zsh

Flags:
  -h, --help   help for completion

Global Flags:
  -c, --context string      Set the kubeconfig context
      --kubeconfig string   Path to the kubeconfig file to use
  -n, --namespace string    Set the namespace
  -p, --platform string     Set the platform type to use [kubernetes, podman, docker, linux]

Use "skupper completion [command] --help" for more information about a command.
$ echo $?
0
```

**Expected behavior**

The error message should go to stderr.  This is specially important for this subcommand, as its output is consumed by the shell.

The exit code should be non-zero, too, in case of failures.

**Environment details**
 - Skupper CLI: 2.0
 - Skupper Operator (if applicable): N/A
 - Platform: N/A

**Additional context**

pwright:
I had another look at this and created https://github.com/pwright/skupper/tree/completion-issue

a) doesn't address Danilo's specific problem (requires lots of code to workaround the mentioned cobra issues)
b) seems to risk duplication of error reporting (idea looks good, but not necessary and risky)

With that in mind, I plan to leave issue open, but labeled wontfix  in the hope cobra changes in future, 
@hash-d remove label if some part of it becomes priority.

PR #2262, opened 2025-09-18
Set stdout so CLI completion feature works correctly

lynnemorrison:
fixes 2135

### DIFF
--- a/internal/cmd/skupper/root/root.go
+++ b/internal/cmd/skupper/root/root.go
@@ -1,6 +1,9 @@
 package root
 
 import (
+	"os"
+	"slices"
+
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/connector"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/debug"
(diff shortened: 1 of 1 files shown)


internal/cmd/skupper/root/root.go
```diff
@@ -38,6 +41,11 @@ func NewSkupperRootCommand() *cobra.Command {
 
 	rootCmd.SetHelpCommand(&cobra.Command{Hidden: true})
 
+	completionCommands := []string{"completion", cobra.ShellCompRequestCmd, cobra.ShellCompNoDescRequestCmd}
+	if len(os.Args) <= 2 || !slices.Contains(completionCommands, os.Args[1]) {
+		rootCmd.SetOut(os.Stderr)
+	}
```
c-kruse, line 47: For my education, could you explain what is going on here? I'm not sure I follow the logic of setting the command output stream to stderr except for completion commands.

lynnemorrison: This change is to fix the issue that Danilo found with completion command.
skupper completion subcommand throws errors into stdout #2135

c-kruse (approved):

internal/cmd/skupper/root/root.go
```diff
@@ -38,6 +41,11 @@ func NewSkupperRootCommand() *cobra.Command {
 
 	rootCmd.SetHelpCommand(&cobra.Command{Hidden: true})
 
+	completionCommands := []string{"completion", cobra.ShellCompRequestCmd, cobra.ShellCompNoDescRequestCmd}
+	if slices.Contains(completionCommands, os.Args[1]) && len(os.Args) <= 2 {
```
hash-d, line 45: This will panic if no arguments are given (there's no os.Args[1], then)

hash-d (changes requested):
The command still does not say what's wrong with the call, or report its error status.

I understand, however, that fixing these might take a fairly big rewrite that may not be worth it.

That said, I agree with Fernando's review, that it would be great to handle also bad arguments (and not just missing arguments).  And the panic scenario needs addressed.

pwright:
closing as no plan to merge

PR #2266, opened 2025-09-19
Fix prefix and workloads; new options

hash-d:
Fixes:

- Creation of TEST_PREFIX failed on some environments (stuck forever)
- Test workloads were being created in duplication (on both east and west sites)

Improvements:

- Allow selection of PYTHON binary
- Allow definition of additional OPTIONS to the `ansible-playbook` call
- Allow user to override value of TEST_PREFIX


granzoto (approved):
LGTM

PR #2272, opened 2025-09-25
TcpListeners should not have the host set in kubernetes

### LINKED ISSUE #2267: Race condition? Creating resources before site breaks Listener
**Describe the bug**

When creating a Skupper Site and its associated resources, if the resources are created immediately before the Site, the Listener ends in a broken state:

Sites, link, listener and connector all show as 'OK', but the communication through the listener fails, and a repeating failure can be seen on the router logs.

**How To Reproduce**

Just proceed as on the description above: create the Connector and Listener resources immediately before creating their sites on a K8S namespace, without waiting for the resources to be in any specific state.

Waiting for the configuration to converge after the fact does not help.

I'll provide a reproducer later.

**Expected behavior**

As V2 is declarative, there is an expectation that the creation of the site and of the resources can be done in any order, and the controller would converge the router configuration to match the K8S resources.

**Environment details**
 - Skupper CLI: N/A
 - Skupper Operator (if applicable): 2.1.1
 - Platform: OpenShift 4.19; GKE

**Additional context**

Tell-tale: you can see that the issue was triggered by checking these two items:

- On the router logs for the namespace where the Listener exists,  you see the following, repeating error lines thousands of times per second:
  
  ```2025-09-25 00:41:43.870905 +0000 FLOW_LOG (info) LOG [89855:273642] BEGIN END parent=89855:0 logSeverity=3 logText=LOG_TCP_ADAPTOR: Listener backend: proactor listener error on backend:8082: proton:io (Cannot assign requested address - listen on backend:8082) sourceFile=/remote-source/skupper-router/app/src/adaptors/adaptor_listener.c sourceLine=172```

  ```2025-09-25 00:52:15.490647 +0000 TCP_ADAPTOR (error) Listener backend: proactor listener error on backend:8082: proton:io (Cannot assign requested address - listen on backend:8082)```

- On config map `skupper-router` in the same namespace, you'll find something like this:

  ```        [
            "tcpListener",
            {
                "name": "backend",
                "host": "backend",
                "port": "8082",
                "address": "backend"
            }
        ]
  ```
  That is, the `host` is being defined to `backend` (whereas on a healthy environment the `tcpListener` should have no `host` defined), and there is no `siteId` field.

If any changes are made to the Listener (such as editing it to change the host to a different value and back, or adding the setting `observer` with value `auto` — which should be a no-op, as that's the default value), the listener gets fixed; the messages stop on the router log and the communication to the backend via the listener starts to work again.

I have executed the same test 50 times each with three different profiles:

- n-w-r-s-l (site_last)
- n-w-s-r-l (link_last)
- n-w-s-l-r (resource_last)

Where:

- n: namespace creation
- w: workload (application) creation
- r: resources (listener, connector) creation
- s: site creation
- l: site linking

Results:

- link_last saw no errors
- resource_last saw 2 errors
- all 50 tests failed for site_last

So, there is strong evidence that the issue is related to the order that the different K8S resources are created.

(Note that I have not confirmed that all the errors on that batch are what's described on this ticket: I have only executed and checked final return code.  However, the manual executions I ran were consistent with the automated results, and on those I have confirmed the issue as described each time)

fgiorgetti:
Fixes #2267


internal/kube/site/extended_bindings.go
```diff
@@ -165,7 +165,9 @@ func (a *ExtendedBindings) updateBridgeConfigForConnector(siteId string, connect
 }
 
 func (a *ExtendedBindings) updateBridgeConfigForListener(siteId string, listener *skupperv2alpha1.Listener, config *qdr.BridgeConfig) {
-	if port, err := a.mapping.GetPortForKey(listener.Name); err == nil {
+	if a.mapping == nil {
+		site.UpdateBridgeConfigForListenerWithHostAndPort(siteId, listener, "", listener.Spec.Port, config)
```
c-kruse: Does this mean that listeners created before site init would try and bind the listener configured port? I suspect that'd probably be a problem - i.e. Listener A{port: 80} + Listener B{port: 80} is valid, but the ports conflict, and also probably are not in the user port range.

fgiorgetti: You're right. I did not notice it was not just setting the host, but it is also persisting the original port.
Let me fix it real quick.

fgiorgetti: @c-kruse could you, please, give it a new try now?

fgiorgetti: By the way, it solved the initial problem. But once site is initialized, the `siteId` field is not being populated.

fgiorgetti: Last commit fixes the missing siteId.

PR #2279, opened 2025-10-01
Remove hosts from Certificate on Owning Resource deletion

### LINKED ISSUE #2257: Disabling HA is not removing skupper-router-2 related host entries from certificate
**Describe the bug**
If HA is disabled on an existing site, the hosts (names and eventual IP addresses) used by
the skupper-router-2 service, are kept in the certificate under `.spec.hosts` as well as through
the `.metadata.annotations`.

**How To Reproduce**
Enable HA, then disable it and notice that the old hosts and IPs are kept.

**Expected behavior**
Only the skupper-router (service) related addresses should be kept.

**Environment details**
 - Skupper: 2.1
 - Platform: kubernetes and openshift

c-kruse:
Fixes the Certificate manager logic for tracking per-owner hosts. This solves an issue where disabling the HA setting on a Site is not reflected on the Certificate, which continues to reamin valid for the HA hosts.

Fixes #2257

### DIFF
--- a/internal/kube/certificates/ownermappings.go
+++ b/internal/kube/certificates/ownermappings.go
@@ -61,9 +61,9 @@ func (m ownerMapping) ApplyMetadata(certificate *skupperv2alpha1.Certificate) bo
 	}
 	changed := false
 	if m.IsControlled {
-		setAnnotation(&certificate.ObjectMeta, "internal.skupper.io/controlled", "true")
+		setAnnotation(&certificate.ObjectMeta, annotationKeySkupperControlled, "true")
 	} else {
-		clearAnnotation(&certificate.ObjectMeta, "internal.skupper.io/controlled")
+		clearAnnotation(&certificate.ObjectMeta, annotationKeySkupperControlled)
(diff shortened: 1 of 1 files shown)


internal/kube/certificates/ownermappings.go
```diff
@@ -0,0 +1,133 @@
...
+// ApplyMetadata Certificate. Returns true when a change was applied
+func (m ownerMapping) ApplyMetadata(certificate *skupperv2alpha1.Certificate) bool {
+	if certificate == nil {
+		return false
+	}
+	changed := false
+	if m.IsControlled {
+		setAnnotation(&certificate.ObjectMeta, "internal.skupper.io/controlled", "true")
+	} else {
+		clearAnnotation(&certificate.ObjectMeta, "internal.skupper.io/controlled")
```
fgiorgetti: Minor: use the new constant: `annotationKeySkupperControlled`.

fgiorgetti (approved):
Awesome job.

Issue #2278, opened 2025-10-01
When HA is disabled it is not reflected in Site Status or in Grants Behavior


c-kruse:
**Describe the bug**
Originally reported by @vsomwanshi https://github.com/skupperproject/skupper/issues/2257#issuecomment-3301780754

After disabling the HA setting on a Site, the Site status.endpoints still contain endpoints for the `skupper-router-2` group, and redeeming an AccessToken to the Site will result in a non operational Link to the removed endpoint.

**How To Reproduce**

* Create an HA Site
* Wait for the Site to be operational and to have .status.endpoints for the skupper-router-2 group.
* Edit the site to disable HA
* Observe the `skupper-router-2` Service is deleted.
* Create an AccessGrant on the subject Site and redeem the corresponding AccessToken on a remote Site.
* Observe two Links - one to the non-operational skupper-router-2 group.

**Expected behavior**
Old endpoints are removed from Site status and Links created through access token redemption accurately reflect the current Site state.

**Environment details**
Skupper 2.0.0+

PR #2282, opened 2025-10-02
Update Endpoints on HA Disable

c-kruse:
Adds a new map to the Site Controller to track SecuredAccess resources configured for a RouterAccess resource + router deployment group combination so that SecuredAccess deletion events can trigger updates to the correct RouterAccess (and eventually Site) Status.Endpoints colleciton. Incidentally fixes the Site Controller logic for deleting SecuredAccess resources on HA disable that had previously relied on a naming coincidence with the default RouterAccess.

Fixes https://github.com/skupperproject/skupper/issues/2278

### DIFF
--- a/internal/kube/site/accessmappings.go
+++ b/internal/kube/site/accessmappings.go
@@ -1,13 +1,13 @@
 package site
 
 type securedAccessMapping struct {
-	RotuerAccessName string
+	RouterAccessName string
 	Group            string
 }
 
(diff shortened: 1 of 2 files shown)


internal/kube/site/accessmappings.go
```diff
@@ -0,0 +1,15 @@
+package site
+
+type securedAccessMapping struct {
+	RotuerAccessName string
```
fgiorgetti: minor: typo

PR #2283, opened 2025-10-08
Clearify error message if user tries to uninstall a nonkube site.

### LINKED ISSUE #2281: Display a friendly message on system uninstall with podman endpoint and controller not installed
If the system-controller and/or the podman socket endpoint are not running, and user runs the
`skupper system uninstall` command, an error message is being displayed, see:

```
failed to uninstall : failed to create container client: container engine is not available on the provided endpoint: "unix:///run/user/1000/podman/podman.sock" (unable to verify version) - Get "http://unix/version": dial unix /run/user/1000/podman/podman.sock: connect: connection refused
```

On such error, skupper could educate user to run `system install` first, as it is unable to communicate with the container engine.

lynnemorrison:
If system-controller or podman socket endpoint is not installed and user issued a "skupper system uninstall" command 
the error message is confusing.   This fix simplifies the error message and provides help on what user should do.

fixes #2281

### DIFF
--- a/internal/nonkube/bootstrap/uninstall.go
+++ b/internal/nonkube/bootstrap/uninstall.go
@@ -44,7 +44,7 @@ func Uninstall(platform string) error {
 
 	cli, err := internalclient.NewCompatClient(endpoint, "")
 	if err != nil {
-		return fmt.Errorf("failed to create container client: %v", err)
+		return fmt.Errorf("failed to create container client: %w", err)
 	}
 
 	container, err := cli.ContainerInspect(containerName)
--- a/internal/cmd/skupper/system/nonkube/system_uninstall.go
+++ b/internal/cmd/skupper/system/nonkube/system_uninstall.go
@@ -3,6 +3,7 @@ package nonkube
 import (
 	"errors"
 	"fmt"
+	"net"
 	"os"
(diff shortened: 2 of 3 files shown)


fgiorgetti (commented):
The message is much clearer now.

But I recommend validating that the error is a `*net.OpError`, before advising
user to run the `skupper system install` command, as if something else happend,
the recommendation can be incorrect.

Thoughts?

internal/nonkube/bootstrap/uninstall.go
```diff
@@ -44,7 +45,11 @@ func Uninstall(platform string) error {
 
 	cli, err := internalclient.NewCompatClient(endpoint, "")
 	if err != nil {
-		return fmt.Errorf("failed to create container client: %v", err)
+		errMsg := fmt.Errorf("failed to create container client: %v", err)
+		netErr := &net.OpError{
```
fgiorgetti: I think it is safer to just wrap the error by switching `%v` by `%w`, with no need to force returning a net.OpError. This way, the validation you're using with `errors.As` still works fine if the actual error is not a net.OpError.

nluaces (approved):

PR #2285, opened 2025-10-09
Update go to 1.24.7

fgiorgetti:


go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/skupperproject/skupper
 
 go 1.24.0
 
-toolchain go1.24.4
+toolchain go1.24.7
```
c-kruse, line 5: IIUC, this is more of a suggestion, and one I think that the container image builds ignores. We may need to update Dockerfile.* - maybe with a parameterized ARG for the builder image tag or similar.

PR #2290, opened 2025-10-17
Allows the Controller to delegate certificate management

fgiorgetti:
This feature enables the Skupper Controller to delegate certificate management to an external controller, allowing the controller to be defined at:

* Controller startup (through: `-certificate-controller` flag or `CERTIFICATE_CONTROLLER` env var)
* Through the `certificate-controller` setting for the following resources:
  * Site
  * SecuredAccess
  * RouterAccess
  * Certificate

### DIFF
--- a/internal/kube/securedaccess/access.go
+++ b/internal/kube/securedaccess/access.go
@@ -96,7 +96,7 @@ func (m *SecuredAccessManager) IsValidAccessType(accessType string) bool {
 	return ok
 }
 
-func (m *SecuredAccessManager) Ensure(namespace string, name string, spec skupperv2alpha1.SecuredAccessSpec, annotations map[string]string, certificateController string, refs []metav1.OwnerReference) error {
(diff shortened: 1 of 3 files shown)


internal/kube/securedaccess/access.go
```diff
@@ -95,7 +96,7 @@ func (m *SecuredAccessManager) IsValidAccessType(accessType string) bool {
 	return ok
 }
 
-func (m *SecuredAccessManager) Ensure(namespace string, name string, spec skupperv2alpha1.SecuredAccessSpec, annotations map[string]string, refs []metav1.OwnerReference) error {
+func (m *SecuredAccessManager) Ensure(namespace string, name string, spec skupperv2alpha1.SecuredAccessSpec, annotations map[string]string, certificateController string, refs []metav1.OwnerReference) error {
```
c-kruse: @fgiorgetti should we accept a separate certificateController arg here or just go off what's in the spec? I'm afraid this flips between what's in certificateController on create (the Site setting) to what's in spec.Settings on update (usually the RouterAccess setting.)

fgiorgetti: Certainly simpler. Thanks for the catch!

Issue #2180, opened 2025-07-10
[v2] Listener conflict with existing service


ajssmith:
**Describe the bug**
When a listener is created, it will modify a pre-existing service with the same name as listener host. 

**How To Reproduce**
Locally xxpose the backend deployment in the hello world example (e.g. east namespace). Then create a listener in the same namespace and observe the service is modified by the controller.

**Expected behavior**
Expectation is that the listener would not modify the pre-existing service. It should indicate the listener is not ready and provide the reason (e.g. detected existing service). 

**Environment details**
Skupper V2.0.1 controller on kubernetes

ajssmith:
A user work-around is to modify the listener host spec to prevent the conflict

PR #2295, opened 2025-10-17
detect existing svc

ajssmith:
Prevent the listener resource from overwriting a pre-existing service. Also, remove log message that would be continuously generated sans configuration change.

### DIFF
--- a/internal/kube/site/site_test.go
+++ b/internal/kube/site/site_test.go
@@ -337,7 +337,7 @@ func TestSite_CheckListener(t *testing.T) {
 			args: args{
 				name: "listener1",
 				listener: &skupperv2alpha1.Listener{
-					ObjectMeta: v1.ObjectMeta{
+					ObjectMeta: metav1.ObjectMeta{
 						Name:      "listener1",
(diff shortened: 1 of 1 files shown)


internal/kube/site/site.go
fgiorgetti, line 579: @ajssmith I believe we need a similar validation here for listeners defined with `exposePodsByName=true`.

internal/kube/site/site.go
```diff
@@ -866,11 +866,10 @@ func (s *Site) CheckConnector(name string, connector *skupperv2alpha1.Connector)
 
 func (s *Site) updateListenerStatus(listener *skupperv2alpha1.Listener, err error) error {
```
nluaces, line 883: Not a blocker: should be this method called something like `updateListenerErrorStatus`, given that it is only used in error cases?

ajssmith: Might lean towards treating in a separate issue. There is a need to look at indicating errors back more generally.

ajssmith:
Fixes #2180

PR #2308, opened 2025-11-04
Adds nginx.command template value to network observer chart

c-kruse:
Adds nginx.command template value to the network observer chart for the network-observer deployment's nginx proxy container. Allows for alternate images and configurations to be used where the command needs to be specified.


c-kruse:
@ajssmith If I remember right, this is important downstream.

fgiorgetti (approved):

Issue #2309, opened 2025-11-06
Console isn't showing data transfers accurately


pwright:
**Describe the bug**
Running iperf, working fine, but console doesn't show 'real' data

**How To Reproduce**
https://github.com/pwright/skupper-example-hello-world/blob/nov-console/skewer.yaml

**Expected behavior**
Console should match iperf reports

**Environment details**
 - Skupper CLI:  2.1.2
 - Skupper Operator (if applicable): na
 - Platform: minikube

**Additional context**

```
[ 13] 156.00-157.00 sec   110 MBytes   923 Mbits/sec   20    786 KBytes       
[SUM] 156.00-157.00 sec   632 MBytes  5.31 Gbits/sec   67             
- - - - - - - - - - - - - - - - - - - - - - - - -
[  5] 157.00-158.00 sec   109 MBytes   911 Mbits/sec   13    662 KBytes       
[  7] 157.00-158.00 sec   154 MBytes  1.30 Gbits/sec   16    560 KBytes       
[  9] 157.00-158.00 sec   105 MBytes   881 Mbits/sec   22    585 KBytes       
[ 11] 157.00-158.00 sec   163 MBytes  1.37 Gbits/sec    7    744 KBytes       
[ 13] 157.00-158.00 sec   106 MBytes   891 Mbits/sec   14    786 KBytes       
[SUM] 157.00-158.00 sec   638 MBytes  5.35 Gbits/sec   72             
- - - - - - - - - - - - - - - - - - - - - - - - -
[  5] 158.00-159.00 sec   110 MBytes   921 Mbits/sec   18    662 KBytes       
[  7] 158.00-159.00 sec   144 MBytes  1.21 Gbits/sec   15    560 KBytes       
[  9] 158.00-159.00 sec   109 MBytes   918 Mbits/sec   21    585 KBytes       
[ 11] 158.00-159.00 sec   145 MBytes  1.22 Gbits/sec    7    744 KBytes       
```

[skupper-dump-east-20251106160113.tar.gz](https://github.com/user-attachments/files/23397516/skupper-dump-east-20251106160113.tar.gz)

[skupper-dump-west-20251106161103.tar.gz](https://github.com/user-attachments/files/23397518/skupper-dump-west-20251106161103.tar.gz)

[Network console (06_11_2025 15：21：58).html](https://github.com/user-attachments/files/23397520/Network.console.06_11_2025.15.21.58.html)

c-kruse:
[...]

PR #2310, opened 2025-11-06
Fixes network observer traffic metrics bug

c-kruse:
Previously traffic metrics (skupper_sent_bytes_total) was only updated when skupper_received_bytes_total was also incremented. This is especially visible in asymetric applications when data flows in one direction like iperf.

Fixes https://github.com/skupperproject/skupper/issues/2309

### DIFF
--- a/cmd/network-observer/internal/collector/connections.go
+++ b/cmd/network-observer/internal/collector/connections.go
@@ -122,11 +122,13 @@ func (c *connectionManager) handleTransportFlow(record vanflow.TransportBiflowRe
 	bs, br := dref(record.Octets), dref(record.OctetsReverse)
 	sentInc := float64(bs - state.BytesSent)
 	receivedInc := float64(br - state.BytesReceived)
-	if receivedInc != 0 {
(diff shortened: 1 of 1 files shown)


c-kruse:
Hi @pwright, that looks to me like a classic example of a confused network observer, and think we've probably fixed just one small slice of traffic data not working here.

From bottom to top:
I see a site-client-10.244.x.x "process" with a ~15s lag between connections an data outbound. This is pretty much exactly what I'd expect when the observer doesn't know who that address (10.244.x.x) is. It sees some traffic from that address, and waits a bit (exactly 15s) then gives up and tosses it under a placeholder.

components/simple-iperf shows only incoming traffic in the server role, and no outgoing bytes as a client. We'd expect that, except the observer did not manage to figure out who that traffic was coming from and made up a placeholder, so it never gets attributed to simple-iperf.

Above this, I kind of lose the thread on what I'm looking at. After the past nearly two years, these pages all still look the same to me.

pwright:
@c-kruse so not related to original issue, thx

PR #2314, opened 2025-11-11
Chart Network Observer incldue ServiceMonitor

c-kruse:
Adds options to enable configuring a ServiceMonitor in conjunction with auth.strategy = "openshift".

Conditionally creates:
- auth.openshift.bearerTokenAuth configuration to allow bearer token delegation by default when openshift auth is selected.
- New ClusterRole and ClusterRoleBindings Resources to grant openshift auth proxy access to subjectaccessreviews for passed bearer tokens when enabled.
- ServiceMonitor configuration
- New ServiceAccount, Role, RoleBinding, and service-account-token Secret for ServiceMonitor to authenticate with auth proxy when enabled. Works with default configuration of auth.openshift.bearerTokenAuth.

### DIFF
--- a/charts/network-observer/templates/_deployment.yaml
+++ b/charts/network-observer/templates/_deployment.yaml
@@ -65,6 +65,9 @@ args:
   - -tls-cert=/etc/certificates/tls.crt
   - -tls-key=/etc/certificates/tls.key
   - -cookie-secret-file=/etc/session-secrets/secret
+  {{ if .Values.auth.openshift.bearerTokenAuth.enabled -}}
+  - -openshift-delegate-urls={{ include "network-observer.bearerTokenAuth" . | fromYaml | toJson }}
+  {{- end }}
 ports:
   - name: https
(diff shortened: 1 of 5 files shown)


c-kruse:
@vsomwanshi I see you have been struggling with instrumenting the network observer with openshift auth and a ServiceMonitor. I have also been struggling to find a nice way to do this. Here is one idea.

Interested in your thoughts on this approach vs. finding a way to bypass the proxy.

*disclaimer - this is very rough and subject to change. The idea is that a values.yaml like this could get it all running (likely with some careful naming and management to avoid conflicts in the required cluster scoped resources.

```
serviceMonitor:
  create: true
  bearerTokenSecret:
    enabled: true

auth:
  strategy: openshift
  openshift:
    bearerTokenAuth:
      createClusterRole: true
      createClusterRoleBinding: true

tls:
  skupperIssued: false
  openshiftIssued: true

route:
  enabled: true

```

Issue #2315, opened 2025-11-12
Instrument Skupper Control Plane Components


c-kruse:
## What

Instrument the Skupper control plane components for Kubernetes, controller and kube-adaptor, with metrics exposing important context about their behavior.

## Why

Operators and maintainers of Skupper need a way to understand the behavior and limitations of the control plane components. Presently the only signal operators have is container logs, which are limited and only answer certain types of questions.

## New Measurements

This proposal suggests two new groups of metrics to be added to the skupper control plane components: kube api utilization and event processing responsiveness.

### Kubernetes API Utilization

Skupper control plane components utilize kubernetes api server resources. Understanding how effectively skupper uses these resources and when api server responsiveness is impacting skupper control plane operations is one key area to understand.

The Kubernetes client for go exposes hooks for such instrumentation: https://pkg.go.dev/k8s.io/client-go/tools/metrics. The subset of metrics involving requests and client imposed rate limiting should be exposed from skupper.

### Event Processing Responsiveness

When a change occurs in the desired or realized state of a skupper resource, the amount of time it takes skupper to perform corrective action is the most important property to watch when scaling skupper. The actionable portion of this measurement can be exposed as work queue delay.

- work queue delay: the delta between the time the skupper process receives an update from the kuberentes api and enters it into its work queue and the time that the worker goroutine pulls that event from the queue to process it.

Secondary measurements that contribute to understanding the contributing factors to a work queue delay

- work queue adds: establishes the rate of new additions to the work queue
- work queue retries: the rate work items are reentered into the queue after a failed attempt at processing the item.
- work duration: how long it takes a worker to process an item from the queue
- work queue depth: how deep is the work queue at present

## Implementation

Expose new TCP ports from the skupper controller and kube-adaptor containers for new http services serving these metrics using prometheus tooling.

### Configuration

Metrics are enabled by default and bind `":9000"`

Override bind address with `SKUPPER_METRICS_ADDRESS=:9000`.
Disable metrics with `SKUPPER_METRICS_DISABLE=true`

Identical for both containers.

PR #2316, opened 2025-11-12
Adds instrumentation for kube client-go

c-kruse:
Exposes a new http server from the skupper controller and kube-adaptor processes for exposing prometheus metrics. Adds instrumentation around kubernetes API http and client rate limiting.

- skupper_kubernetes_client_http_request_duration_seconds: request latency by endpoint.
- skupper_kubernetes_client_http_requests_total: total requests by status code.
- skupper_kubernetes_client_http_retries_total: total request retries by status code.
- skupper_kubernetes_client_rate_limiter_duration_seconds: client side rate limiting latency by endpoint

### DIFF
--- a/internal/kube/metrics/client.go
+++ b/internal/kube/metrics/client.go
@@ -24,13 +24,13 @@ func MustRegisterClientGoMetrics(registry *prometheus.Registry) {
 			Subsystem: "kubernetes_client",
 			Name:      "http_requests_total",
 			Help:      "Total number of kubernetes client requests by status code.",
-		}, []string{"status_code"}),
+		}, []string{"method", "status_code"}),
 		retries: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: "skupper",
 			Subsystem: "kubernetes_client",
 			Name:      "http_retries_total",
 			Help:      "Total number of kubernetes client requests retried by status code.",
-		}, []string{"status_code"}),
+		}, []string{"method", "status_code"}),
 	}
 	rateLimiterMetrics := &clientGoRateLimiterMetrics{
 		latency: prometheus.NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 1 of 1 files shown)


c-kruse:
Issue: https://github.com/skupperproject/skupper/issues/2315

internal/kube/metrics/client.go
```diff
@@ -0,0 +1,77 @@
...
+// k8s.io/client-go/tools/metrics package with the prometheus registry.
+func MustRegisterClientGoMetrics(registry *prometheus.Registry) {
+	httpMetrics := &clientGoHttpMetrics{
+		latency: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: "skupper",
+			Subsystem: "kubernetes_client",
+			Name:      "http_request_duration_seconds",
+			Help:      "Latency of kubernetes client requests in seconds by endpoint.",
+		}, []string{"endpoint"}),
+		results: prometheus.NewCounterVec(prometheus.CounterOpts{
```
fgiorgetti, line 22: It would be nice to have a results counter for "path" and "method" as well to help identifying operations happening during the reconciliation of a given resource ytpe.

c-kruse: That makes sense! I did consider that.

The client-go metrics API doesn't give us all of that context in the [ResultMetric](https://pkg.go.dev/k8s.io/client-go/tools/metrics#ResultMetric) interface, and I think that's intentional. It gives us the [LatencyMetric](https://pkg.go.dev/k8s.io/client-go/tools/metrics#LatencyMetric) (http_request_duration_seconds) that does have method and URL path. I think the best we can do is something along these lines:
- Add method to http_request_duration_seconds (and maybe rate_limiter_duration_seconds.)
- Monitor "sum by (method, endpoint) (rate(http_request_duration_seconds_count[d])) " for request rate by (normalized) url path and method.
- Monitor "sum by (status_code) (rate(http_requests_total{status_code!=200}[d])) " for http error rate in general.

I suspect the intent of the ResultMetric is mostly for clients working with multiple cluster API servers as a lower cardinality indicator of api server trouble. For our single-cluster context that is less interesting, but I included it anyways since the latency metric is absent of the status code field.

internal/kube/metrics/client.go
```diff
@@ -0,0 +1,77 @@
...
+			Subsystem: "kubernetes_client",
+			Name:      "http_request_duration_seconds",
+			Help:      "Latency of kubernetes client requests in seconds by endpoint.",
+		}, []string{"method", "endpoint"}),
+		results: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: "skupper",
+			Subsystem: "kubernetes_client",
+			Name:      "http_requests_total",
+			Help:      "Total number of kubernetes client requests by status code.",
+		}, []string{"status_code"}),
```
fgiorgetti: Minor suggestion just to add the method to this and to the http_retries_total as well.

PR #2327, opened 2025-11-25
Add securityContext to skupper-router deployment

### LINKED ISSUE #2326: Containers missing securityContext
**Describe the bug**
When a site is created, the securityContext for the containers and for the template.spec is not being added.
This could prevent use in restricted clusters.

**Expected behavior**
All containers have the security context applied and seccompProfile on supported Kubernetes versions.

**Environment details**
Skupper 2.1.2

fgiorgetti:
Fixes #2326

### DIFF
--- a/internal/kube/site/resources/apply.go
+++ b/internal/kube/site/resources/apply.go
@@ -31,12 +31,12 @@ type Labelling interface {
 	SetObjectMetadata(namespace string, name string, kind string, meta *metav1.ObjectMeta) bool
 }
 
-func resourceTemplates(site *skupperv2alpha1.Site, group string, size sizing.Sizing, labelling Labelling) []resource.Template {
+func resourceTemplates(site *skupperv2alpha1.Site, group string, size sizing.Sizing, labelling Labelling, disableSecCtx bool) []resource.Template {
 	templates := []resource.Template{
 		{
 			Name:       "deployment",
 			Template:   routerDeploymentTemplate,
-			Parameters: getCoreParams(site, group, size).setLabelsAndAnnotations(labelling, site.Namespace, "skupper-router", "Deployment"),
+			Parameters: getCoreParams(site, group, size, disableSecCtx).setLabelsAndAnnotations(labelling, site.Namespace, "skupper-router", "Deployment"),
 			Resource: schema.GroupVersionResource{
 				Group:    "apps",
 				Version:  "v1",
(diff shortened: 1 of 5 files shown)


internal/kube/site/resources/apply.go
```diff
@@ -162,11 +164,26 @@ func getCoreParams(site *skupperv2alpha1.Site, group string, size sizing.Sizing)
 		Sizing:             size,
 		Labels:             map[string]string{},
 		EnableAntiAffinity: enableAntiAffinity(site),
+		AddPodSecurity:     addPodSecurityContext(clients),
+	}
+}
+
+// addPodSecurityContext Only added if server version is >=1.24
```
c-kruse: I am curious where we came up with this version: Looking back in api docs history I didn't see much evidence that this wouldn't work prior to 1.24 - near as I can tell it would work as far back as 1.19 (when seccomp went GA.)

internal/kube/site/resources/skupper-router-deployment.yaml
```diff
@@ -159,6 +177,12 @@ spec:
       volumes:
       - emptyDir: {}
         name: skupper-router-certs
+{{- if .AddPodSecurity }}
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
```
c-kruse: I suspect seccompProfile type "RuntimeDefault" is going to be suitable for most use cases, but hard coding it would likely end up being a deal breaker in some environments. We definitely need a way to either omit a seccompProfile (and defer the setting to an external controller or mutating admissions webhook) or a way to configure this (enviornment variable/flag/site setting.)

internal/kube/site/resources/apply.go
fgiorgetti: I noticed that as well. I left is as is, because that is the version defined in V1.
Anyway, I believe we will end up removing the `seccompProfile` element from the PR,
so this whole function can go away.

internal/kube/site/resources/skupper-router-deployment.yaml
fgiorgetti: I agree with you. The seccompProfile should be left out of this.

fgiorgetti:
@nluaces @c-kruse @hash-d Do you guys see any issue with keeping just the following entries:

# top level securityContext (seccompProfile left out)
* deployment.spec.template.spec.securityContext.runAsNonRoot (true)

# all containers
* deployment.spec.template.spec.containers[*].securityContext.allowPrivilegeEscalation (false)
* deployment.spec.template.spec.containers[*].securityContext.capabilities (drop all)
* deployment.spec.template.spec.containers[*].securityContext.runAsNonRoot (true)

# all initContainers
* deployment.spec.template.spec.initContainers[*].securityContext.allowPrivilegeEscalation (false)
* deployment.spec.template.spec.initContainers[*].securityContext.capabilities (drop all)
* deployment.spec.template.spec.initContainers[*].securityContext.runAsNonRoot (true)

I believe that even these agreed upon fields, could be added by default as part of the upcoming 2.2 release,
but we could introduce a flag/env-var to disable this behavior, if someone needs that for any reason.

If you guys agree, I will make these changes and request a new round of review.

Thank you.

c-kruse:
@fgiorgetti that approach sounds good to me!

fgiorgetti:
@c-kruse @nluaces @hash-d I have just added a flag to disable the new default behavior of adding a securityContext with the default values we discussed here. Please give it a try and let me know your thoughts.

Issue #2333, opened 2025-12-02
Reload automatically changes in custom resources for docker and podman system platforms


nluaces:
**Is your feature request related to a problem? Please describe.**
Each time that a custom resource is changed, the user needs to run the command `skupper system reload` to make the changes effective. 
This suggested change consist of finding an additional way to reload custom resources while avoiding running the reload command.

**Describe the solution you'd like**
Custom resources' changes (create/update/delete) should be watched by the system-controller.
Additionally, add an environment variable to switch off this feature.

**Additional context**
This change will be applicable only to docker and podman platforms for now.

nluaces:
Implemented by https://github.com/skupperproject/skupper/pull/2356

PR #2334, opened 2025-12-02
Add a watcher for system input files

nluaces:
Implements #2333 

**Changes:**
- The namespace controller has an additional watcher for custom resources created in the designated directories.
- This feature is enabled by default ("auto"). To disable it, export the environment variable `SKUPPER_SYSTEM_RELOAD_TYPE` to "manual"
- CLI change: the command `skupper system install` needs to pass said env. variable to the system-controller pod.

**Additional information:**
- These changes only apply to podman and docker platforms and they can be tested by running:
```
export SKUPPER_SYSTEM_CONTROLLER_IMAGE=quay.io/nluaces/system-controller:latest
```

### DIFF
--- a/internal/nonkube/bootstrap/teardown.go
+++ b/internal/nonkube/bootstrap/teardown.go
@@ -22,7 +22,21 @@ func Teardown(namespace string) error {
 	if err != nil {
 		return err
 	}
-	return RemoveAll(namespace, platform)
+
+	if err := removeRouter(namespace, platform); err != nil {
+		return err
+	}
+
+	if err := removeService(namespace, platform); err != nil {
+		return err
+	}
+
+	if err := removeDefinition(namespace); err != nil {
+		return err
+	}
+
+	fmt.Printf("Namespace \"%s\" has been removed\n", namespace)
+	return nil
 
 }
 
@@ -85,23 +99,4 @@ func removeService(namespace string, platform string) error {
 	}
 
 	return nil
-}
-
-func RemoveAll(namespace string, platform string) error {
-
-	if err := removeRouter(namespace, platform); err != nil {
-		return err
-	}
-
-	if err := removeService(namespace, platform); err != nil {
-		return err
-	}
-
-	if err := removeDefinition(namespace); err != nil {
-		return err
-	}
-
-	fmt.Printf("Namespace \"%s\" has been removed\n", namespace)
-	return nil
-
 }
--- a/internal/nonkube/controller/input_resource_handler.go
+++ b/internal/nonkube/controller/input_resource_handler.go
@@ -23,14 +23,14 @@ type InputResourceHandler struct {
 	inputPath       string
 	Bootstrap       func(config *bootstrap.Config) (*api.SiteState, error)
 	PostExec        func(config *bootstrap.Config, siteState *api.SiteState)
-	TearDown        func(namespace string, platform string) error
+	TearDown        func(namespace string) error
 	ConfigBootstrap bootstrap.Config
 	lock            sync.Mutex
 }
 
 type Bootstrap func(config *bootstrap.Config) (*api.SiteState, error)
 type PostBootstrap func(config *bootstrap.Config, siteState *api.SiteState)
-type TearDown func(namespace string, platform string) error
+type TearDown func(namespace string) error
 
 func NewInputResourceHandler(namespace string, inputPath string, bs Bootstrap, pbs PostBootstrap, td TearDown) *InputResourceHandler {
 
@@ -144,7 +144,7 @@ func (h *InputResourceHandler) processInputFile() error {
 
 func (h *InputResourceHandler) tearDownNamespace() error {
 	h.logger.Info("No site configured, tearing down namespace")
-	err := h.TearDown(h.namespace, string(h.ConfigBootstrap.Platform))
+	err := h.TearDown(h.namespace)
 	if err != nil {
 		return err
 	}
--- a/internal/nonkube/controller/input_resource_handler_test.go
+++ b/internal/nonkube/controller/input_resource_handler_test.go
@@ -148,4 +148,4 @@ func mockBootstrapFailed(config *bootstrap.Config) (*api.SiteState, error) {
 	return nil, fmt.Errorf("failed to bootstrap")
 }
 
-func mockTearDown(namespace string, platform string) error { return nil }
+func mockTearDown(namespace string) error { return nil }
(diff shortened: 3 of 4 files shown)


internal/nonkube/bootstrap/install.go
```diff
@@ -75,6 +76,8 @@ func Install(platform string) error {
 		"CONTAINER_ENDPOINT":  config.containerEndpoint,
 		"SKUPPER_OUTPUT_PATH": config.hostDataHome,
 		"CONTAINER_ENGINE":    config.containerEngine,
+		"SKUPPER_SYSTEM_RELOAD_TYPE": utils.DefaultStr(os.Getenv(types.ENV_SYSTEM_AUTO_RELOAD),
+			types.SystemReloadTypeAuto),
```
fgiorgetti: Should the default here be `SystemReloadTypeManual`, instead?

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,116 @@
...
+	lock            sync.Mutex
+}
+
+func NewInputResourceHandler(namespace string, inputPath string, bStrap func(config *bootstrap.Config) (*api.SiteState, error), postBootStrap func(config *bootstrap.Config, siteState *api.SiteState)) *InputResourceHandler {
+
+	systemReloadType := utils.DefaultStr(os.Getenv(types.ENV_SYSTEM_AUTO_RELOAD),
+		types.SystemReloadTypeManual)
+
+	if systemReloadType == types.SystemReloadTypeManual {
+		slog.Default().Error("Automatic reloading is not configured.")
```
fgiorgetti: Maybe make this a debug level message, not an error.

internal/nonkube/compat/site_state_renderer.go
```diff
@@ -38,10 +38,19 @@ func (s *SiteStateRenderer) Render(loadedSiteState *api.SiteState, reload bool)
...
 	endpoint := os.Getenv("CONTAINER_ENDPOINT")
-	if endpoint == "" {
-		endpoint = fmt.Sprintf("unix://%s/podman/podman.sock", api.GetRuntimeDir())
+
+	// the container endpoint is mapped to the podman socket inside the container
+	if api.IsRunningInContainer() {
+		endpoint = "unix:///var/run/podman.sock"
 		if s.Platform == "docker" {
-			endpoint = "unix:///run/docker.sock"
+			endpoint = "unix:///var/run/docker.sock"
```
fgiorgetti: I believe that the old value is still appropriate.

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,148 @@
...
+	namespace       string
+	inputPath       string
+	Bootstrap       func(config *bootstrap.Config) (*api.SiteState, error)
+	PostExec        func(config *bootstrap.Config, siteState *api.SiteState)
+	TearDown        func(namespace string, platform string) error
+	ConfigBootstrap bootstrap.Config
+	lock            sync.Mutex
+}
+
+func NewInputResourceHandler(namespace string, inputPath string, bStrap func(config *bootstrap.Config) (*api.SiteState, error), postBootStrap func(config *bootstrap.Config, siteState *api.SiteState), tearDown func(namespace string, platform string) error) *InputResourceHandler {
```
fgiorgetti: Minor suggestion, maybe `bStrap, postBootStrap and tearDown` could be part of an interface or just a type (just for readability)?

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	handler.ConfigBootstrap = bootstrap.Config{
+		Namespace: namespace,
+		InputPath: inputPath,
+		Platform:  platform,
+		Binary:    binary,
+	}
+
+	handler.logger = slog.Default().With("component", "input.resource.handler", "namespace", namespace)
```
fgiorgetti: Maybe initialize and use this logger right after the InputResourceHandler creation.
If a `linux` or an invalid platform is used, the logs won't show for which namespace it was set.

internal/nonkube/controller/input_resource_handler.go
nluaces: done!

internal/nonkube/controller/input_resource_handler.go
nluaces: done!

internal/nonkube/compat/site_state_renderer.go
nluaces: I have double checked this using docker as platform and leaving `unix:///run/docker.sock`:

```
2026/02/09 16:15:26 ERROR Failed to bootstrap: failed to render site state: failed to create container client: Container engine is not available on provided endpoint - stat /run/docker.sock: no such file or directory component=input.resource.handler namespace=default
```
I think the problem is similar to podman platform, regarding where the container endpoint is mapped to the socket.

fgiorgetti: I believe that `/var/run` should work, as it is supposed to be just a lihk to `/run`, but I am wondering when did it change. Which version of docker do you have, Noe?
Anyway, if we want to change this, then there are several other locations in the repo that must be updated as well. If something has changed recently, then better changing it everywhere.

nluaces: `Docker version 29.2.1, build a5c7197`

nluaces: > Anyway, if we want to change this, then there are several other locations in the repo that must be updated as well. If something has changed recently, then better changing it everywhere.

I agree. Should be changed everywhere as part of this pull request? I can also do it as part of another issue.

fgiorgetti: You can do it here, at least all references will be using /var/run/docker.sock.

nluaces: @fgiorgetti let me know if these commits work for you:
867d6ac204b7f34a5120cca22426d8c9cb5b3a38
a90b119fe6afc3e493d2d2fae9f90b9be4f07793

I haven't done these changes in the implementations of install/uninstall commands, as they are not meant to be executed inside the container (same goes for bootstrap.sh and system_controller.sh).

internal/nonkube/bootstrap/teardown.go
```diff
@@ -102,3 +86,22 @@ func removeService(namespace string, platform string) error {
 
 	return nil
 }
+
+func RemoveAll(namespace string, platform string) error {
```
fgiorgetti: Maybe remove the platform as a parameter and move the platform loading logic into this function,
as the platform is not variable.

nluaces: I removed this function and used TearDown instead!

nluaces:
Closing as this implementation was included in https://github.com/skupperproject/skupper/pull/2356

PR #2352, opened 2026-01-13
Enhanced Labeling Configuration

### LINKED ISSUE #2109: Enhanced Labeling Configuration
In skupper 2.0, a special ConfigMap with the label `skupper.io/label-template` can be used to configure the skupper controller for kubernetes to apply labels to specific resources that it creates. When a label template ConfigMap has a data field for "name" or "kind", the skupper controller will only apply the ConfigMap labels to resources matching those properties.

In the case of more dynamic resources, like v2alpha1.Listeners and the Service(s) they create, it would be nice to be able to apply user-provided labels without a per-resource ConfigMap.

Example: A user has a Listener with exposePodsByName set, and needs to to target the set of Services created by this Listener. The existing scheme cannot accommodate this.

**proposed solution**

A two part plan:
1. Add an additional label-template ConfigMap data field "labelSelector". When set, the labels from the ConfigMap will be applied only to resources matching that labelSelector. Invalid labelSelectors will be logged and ignored.
2. Add skupper provided labels to identify resources created by the skupper controller. Continuing the above example, we would add an `internal.skupper.io/listener: <listener-name>` label to Listener services.

After implementing these, a user could create a ConfigMap with labels `skupper.io/label-template` and `com.acme.foo/custom-label`, and a data field `labelSelector: "internal.skupper.io/listener = my-listener"`.  Services skupper creates for the Listener "my-listener" would get the com.acme.foo/custom-label label.

AryanP123:
add labelSelector support; set listener Service label to name; unit tests

Fixes #2109

### DIFF
--- a/internal/kube/site/labels/registry.go
+++ b/internal/kube/site/labels/registry.go
@@ -5,6 +5,8 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/tools/cache"
 )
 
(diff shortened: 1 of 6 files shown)


internal/kube/site/labels/registry.go
```diff
@@ -115,6 +116,26 @@ func (r *Registry) filter(name string, kind string, labels map[string]string, an
 		if !matchKey(cm, "kind", kind) {
 			continue
 		}
+		if cm.Data != nil {
+			if selector, ok := cm.Data["labelSelector"]; ok && selector != "" {
+				if target == nil {
+					continue
+				}
+				req, err := k8slabels.Parse(selector)
```
c-kruse: @AryanP123 I'd really rather see this cached in the Registry - probably ideally up-front on Update, but on first access is probably okay also.

It is possible that Registry.filter is going to get called a lot, so logging "Ignoring label-template" each time could get noisy, and rerunning k8slabels.Parse could end up churning lots of memory+cpu cycles in the long run.

c-kruse:
@AryanP123 that's right!

fgiorgetti (approved):

PR #2356, opened 2026-01-18
Enable dynamic changes in the router for non-kube environments 

nluaces:
Implements #2337 
Branch based on https://github.com/skupperproject/skupper/pull/2334, the proper commits that implement this issue are bb8ab35d66b2159fe163799f16d67a73b3106733 and the following.

```mermaid
flowchart TD
   id1([new listener resource is created])-->|triggers|InputResourceHandler-->Q{site exists?}
   Q -->|Yes| A[Refresh Router Config]
   Q -->|No| B[Bootstrap]
```
```mermaid
flowchart LR
  SystemAdaptorHandler-->|has a callback|RouterStatusHandler
  SystemAdaptorHandler-->|spawns|processRouterConfig-->id2([reconciliates the router config file with the router])
```
The InputResourceHandler was modified to not bootstrap in case a site is already created in the namespace.

EDIT: A new flag called `--reload-type` is available in the `skupper system install` command. It has priority over the env. variable.

### DIFF
--- a/internal/nonkube/bootstrap/teardown.go
+++ b/internal/nonkube/bootstrap/teardown.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/skupperproject/skupper/api/types"
-	"github.com/skupperproject/skupper/internal/config"
 	internalclient "github.com/skupperproject/skupper/internal/nonkube/client/compat"
 	"github.com/skupperproject/skupper/internal/nonkube/client/fs"
 	"github.com/skupperproject/skupper/internal/nonkube/common"
@@ -23,14 +21,6 @@ func Teardown(namespace string) error {
 	configuredPlatform, err := platformLoader.Load(namespace)
 	if err != nil {
 		return err
-	}
-
-	currentPlatform := config.GetPlatform()
-	if currentPlatform.IsKubernetes() {
-		currentPlatform = types.PlatformPodman
-	}
-	if string(currentPlatform) != configuredPlatform {
-		return fmt.Errorf("existing namespace uses %q platform and it cannot change to %q", configuredPlatform, string(currentPlatform))
 	}
 
 	if err := removeRouter(namespace, configuredPlatform); err != nil {
--- a/internal/cmd/skupper/system/nonkube/system_stop.go
+++ b/internal/cmd/skupper/system/nonkube/system_stop.go
@@ -4,8 +4,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/skupperproject/skupper/api/types"
 	"github.com/skupperproject/skupper/internal/config"
 	"github.com/skupperproject/skupper/internal/nonkube/bootstrap"
+	"github.com/skupperproject/skupper/internal/nonkube/common"
 	"github.com/skupperproject/skupper/internal/utils/validator"
 	"github.com/spf13/cobra"
 )
@@ -43,6 +45,20 @@ func (cmd *CmdSystemStop) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("namespace is not valid: %s", err))
 		}
+	}
+
+	platformLoader := &common.NamespacePlatformLoader{}
+	configuredPlatform, err := platformLoader.Load(cmd.Namespace)
+	if err != nil {
+		return err
+	}
+
+	currentPlatform := config.GetPlatform()
+	if currentPlatform.IsKubernetes() {
+		currentPlatform = types.PlatformPodman
+	}
+	if string(currentPlatform) != configuredPlatform {
+		validationErrors = append(validationErrors, fmt.Errorf("existing namespace uses %q platform and it cannot change to %q", configuredPlatform, string(currentPlatform)))
 	}
 
 	return errors.Join(validationErrors...)
--- a/internal/cmd/skupper/system/nonkube/system_stop_test.go
+++ b/internal/cmd/skupper/system/nonkube/system_stop_test.go
@@ -2,10 +2,15 @@ package nonkube
 
 import (
 	"fmt"
+	"os"
+	"path"
 	"testing"
 
+	"github.com/skupperproject/skupper/api/types"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common"
 	"github.com/skupperproject/skupper/internal/cmd/skupper/common/testutils"
+	"github.com/skupperproject/skupper/internal/config"
+	"github.com/skupperproject/skupper/pkg/nonkube/api"
 	"gotest.tools/v3/assert"
 )
 
@@ -16,23 +21,62 @@ func TestCmdSystemTearDown_ValidateInput(t *testing.T) {
 		args            []string
 		envSystemReload string
 		expectedError   string
+		configPlatform  string // platform.yaml content
+		currentPlatform string // SKUPPER_PLATFORM env var value
 	}
 
 	testTable := []test{
 		{
-			name:          "arg-not-accepted",
-			args:          []string{"namespace"},
-			expectedError: "this command does not accept arguments",
+			name:            "arg-not-accepted",
+			args:            []string{"namespace"},
+			configPlatform:  "podman",
+			currentPlatform: "podman",
+			expectedError:   "this command does not accept arguments",
 		},
 		{
-			name:          "invalid-namespace",
-			namespace:     "Invalid",
-			expectedError: "namespace is not valid: value does not match this regular expression: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$",
+			name:            "invalid-namespace",
+			namespace:       "Invalid",
+			configPlatform:  "docker",
+			currentPlatform: "docker",
+			expectedError:   "namespace is not valid: value does not match this regular expression: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$",
+		},
+		{
+			name:            "platform-mismatch-podman-vs-docker",
+			namespace:       "test-ns",
+			configPlatform:  "podman",
+			currentPlatform: "docker",
+			expectedError:   `existing namespace uses "podman" platform and it cannot change to "docker"`,
 		},
 	}
 
 	for _, test := range testTable {
 		t.Run(test.name, func(t *testing.T) {
+
+			// Create a temporary directory to simulate the namespace output path
+			tmpDir := t.TempDir()
+
+			namespace := test.namespace
+			if namespace == "" {
+				namespace = "default"
+			}
+
+			internalDir := path.Join(tmpDir, "skupper", "namespaces", namespace, string(api.InternalBasePath))
+			err := os.MkdirAll(internalDir, 0755)
+			assert.NilError(t, err)
+
+			//create platform config file
+			platformYaml := fmt.Sprintf("platform: %s\n", test.configPlatform)
+			err = os.WriteFile(path.Join(internalDir, "platform.yaml"), []byte(platformYaml), 0644)
+			assert.NilError(t, err)
+
+			t.Setenv("XDG_DATA_HOME", tmpDir)
+
+			//set the current platform
+			config.ClearPlatform()
+			t.Setenv(types.ENV_PLATFORM, test.currentPlatform)
+			t.Cleanup(func() {
+				config.ClearPlatform()
+			})
 
 			command := &CmdSystemStop{}
 			command.Namespace = test.namespace


internal/nonkube/common/site_state_renderer_common.go
```diff
@@ -81,3 +82,22 @@ func CreateRouterAccess(siteState *api.SiteState) error {
...
+		name := fmt.Sprintf("skupper-local")
+		if config == nil {
+			return fmt.Errorf("skupper-local RouterAccess cannot be created without a router config")
+		}
+
+		if config.Listeners == nil {
+			return fmt.Errorf("skupper-local RouterAccess cannot be updated without listeners")
+		}
+
+		port := config.Listeners["skupper-local-normal"].Port
```
fgiorgetti: Safer to check if the "skupper-local-normal" listener actually exists before reading its port.

internal/nonkube/common/site_state_renderer_common.go
```diff
@@ -81,3 +82,22 @@ func CreateRouterAccess(siteState *api.SiteState) error {
 	}
 	return nil
 }
+
+func UpdateRouterAccess(siteState *api.SiteState, config *qdr.RouterConfig) error {
```
fgiorgetti: Maybe rename to `RecoverRouterAccess`, wdyt?

internal/nonkube/compat/site_state_renderer.go
```diff
@@ -38,10 +38,19 @@ func (s *SiteStateRenderer) Render(loadedSiteState *api.SiteState, reload bool)
...
 	endpoint := os.Getenv("CONTAINER_ENDPOINT")
-	if endpoint == "" {
-		endpoint = fmt.Sprintf("unix://%s/podman/podman.sock", api.GetRuntimeDir())
+
+	// the container endpoint is mapped to the podman socket inside the container
+	if api.IsRunningInContainer() {
+		endpoint = "unix:///var/run/podman.sock"
 		if s.Platform == "docker" {
-			endpoint = "unix:///run/docker.sock"
+			endpoint = "unix:///var/run/docker.sock"
```
fgiorgetti: I believe the original one is still correct.

internal/nonkube/compat/site_state_renderer.go
```diff
@@ -38,10 +38,19 @@ func (s *SiteStateRenderer) Render(loadedSiteState *api.SiteState, reload bool)
 	}
 	s.loadedSiteState = loadedSiteState
 	endpoint := os.Getenv("CONTAINER_ENDPOINT")
-	if endpoint == "" {
-		endpoint = fmt.Sprintf("unix://%s/podman/podman.sock", api.GetRuntimeDir())
+
+	// the container endpoint is mapped to the podman socket inside the container
+	if api.IsRunningInContainer() {
+		endpoint = "unix:///var/run/podman.sock"
```
fgiorgetti: If running in a container, the endpoint should be provided through:
```
os.Getenv("CONTAINER_ENDPOINT")
```
Then, if the endpoint is not set, you can use the defaults you have below.

Maybe we could have a function under `pkg/nonkube/api/environment.go` that takes the platform as an argument and returns the correct container endpoint, eliminating the dups we have (I should have done that earlier).

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,143 @@
...
+
+	_, err := os.Stat(api.GetInternalOutputPath(h.namespace, api.RuntimeSiteStatePath))
+	if err == nil {
+		//a site has already been created, no need to bootstrap
+		siteState, err := h.siteStateLoader.Load()
+		if !siteState.IsBundle() {
+			if err != nil {
+				return fmt.Errorf("Failed to load site: %s", err)
+			}
+			h.siteStateRenderer.Refresh(siteState)
```
fgiorgetti: Error is not being handled.

internal/nonkube/controller/namespace_controller.go
```diff
@@ -39,8 +45,16 @@ func (w *NamespaceController) Start() {
 		routerConfigHandler.AddCallback(routerStateHandler)
 		collectorLifecycleHandler := NewCollectorLifecycleHandler(w.ns)
 		routerStateHandler.SetCallback(collectorLifecycleHandler)
+		systemAdaptorHandler := NewSystemAdaptorHandler(w.ns)
+		routerStateHandler.SetCallback(systemAdaptorHandler)
```
fgiorgetti: This one is overwriting the collectorLifecycleHandler.
I believe we need change the RouterStateHandler to handle multiple callbacks.

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (h *InputResourceHandler) OnBasePathAdded(basePath string) {}
+
+func (h *InputResourceHandler) processInputFile() error {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	_, err := os.Stat(api.GetInternalOutputPath(h.namespace, api.RuntimeSiteStatePath))
+	if err == nil {
```
fgiorgetti: I believe an extra validation is needed here.
If the namespace has already been initialized, the controller needs to check if the
router container is using the correct image. If not, then we need to "reload" the site.

This will be needed in cases when the system controller is updated.

By the way, we still need to handle system controller updates through the CLI.
I am not sure if it could be part of the `system install` command, through a flag,
or if it should be a separate command like `reinstall`.

Anyway, I believe we need to think about it, as it will be needed as existing sites
cannot be reloaded when using automated reloads.

internal/qdr/amqp_mgmt.go
fgiorgetti, line 1: A general comment here. The amqp Agent originally prints its log messages to its own kube-adaptor container.

Now that it is running as part of the system-controller container, the messages being logged are not qualified for the namespace, so it is hard to tell for which namespace the operations are being executed against. In example:

```
2026/02/06 19:08:46 CREATE io.skupper.router.tcpListener backend map[address:backend host:0.0.0.0 name:backend port:8080]
```

Maybe rebasing this PR on top of the main branch, which already contains the work done by @JPadovano1483 to use slog instead could help (but we'd still need to add the namespace attribute).

internal/nonkube/compat/site_state_renderer.go
nluaces: https://github.com/skupperproject/skupper/pull/2334#discussion_r2783844313

internal/nonkube/controller/namespace_controller.go
nluaces: Thank you! I've missed this one!

internal/nonkube/controller/input_resource_handler.go
nluaces: > If the namespace has already been initialized, the controller needs to check if the
> router container is using the correct image.

I need some clarification, what is considered the correct image? the one specified in the env variable? Or simply check that there is a router running for that namespace?

internal/nonkube/compat/site_state_renderer.go
nluaces: When `skupper system install` runs, it creates a volume bind between with `CONTAINER_ENDPOINT` env variable and a volume destination that starts with `/var/run/` (I believe that the bootstrap script works like that as well): https://github.com/skupperproject/skupper/blob/ce10f83e0e55e4ff916781e7b1cdb9ab9e6404d1/internal/nonkube/bootstrap/install.go#L80-L104

If running inside the container we use the env variable instead of the mapped endpoint, I think it is not going to work as expected.

nluaces:
> * The site name in the router config is changing (we need to check if this will cause issues to the console or vanflow metrics)

I checked that the name of the site was not changing when the controller just refresh (for example, when a listener has been created); do you have any examples for me to understand this better?

internal/qdr/amqp_mgmt.go
nluaces: I have rebased this PR; after merging this, I'm planning to do some refactoring like using `sync_router_ops` for both kube-adaptor and system-adaptor, in order to not duplicate code: https://github.com/nluaces/skupper/blob/d4bc626eb701ba734864ebecf9ab3e61be0b58ee/internal/qdr/sync_router_ops.go

I can change the amqp Agent functions to add the namespace as part of this refactoring. What do you think?

internal/nonkube/controller/input_resource_handler.go
fgiorgetti: Maybe we can handle updates separately from this initial PR, so it will be simpler to evaluate.
Let's just raise an enhancement issue for now, so we can think more about it.
Or we could do something similar to what the kube controller is doing, which consists of comparing
the version defined at the controller with the version defined to a given site/namespace. If controller
has a newer version, then we could use the same router image we use when a new site is created.
But again, better to keep it separate. WDYT?

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,181 @@
...
+	return strings.HasSuffix(name, ".json") || strings.HasSuffix(name, ".yaml") || strings.HasSuffix(name, ".yml")
+}
+
+func (h *InputResourceHandler) OnBasePathAdded(basePath string) {}
+
+func (h *InputResourceHandler) processInputFile() error {
+	_, err := os.Stat(api.GetInternalOutputPath(h.namespace, api.RuntimeSiteStatePath))
+	if err == nil {
+		//a site has already been created, no need to bootstrap
+		siteState, err := h.siteStateLoader.Load()
```
fgiorgetti: I'd recommend handling error first here, but looking at the Load method, I see that
it is returning a siteState (not nil) and an error at a unique place, so better to always
return siteState as nil on error conditions there (at the Load method).

internal/nonkube/compat/site_state_renderer.go
fgiorgetti: Actually, you brought up a good point. For auto-reloads, the system-controller container must be aware of all endpoints on the host machine, beforehands. So when the system-controller container is created, it needs to know the PODMAN_CONTAINER_ENDPOINT as well as the DOCKER_CONTAINER_ENDPOINT.

This way if the system controller is running with podman, but a docker site is created, the system controller already knows what is the endpoint to map, from the host machine into the target router container.

fgiorgetti: Or... we could simply argue that if the system controller is using podman, it will only initialize podman sites.

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,181 @@
...
+
+	siteStateLoader := &common2.FileSystemSiteStateLoader{
+		Path:   h.ConfigBootstrap.InputPath,
+		Bundle: h.ConfigBootstrap.IsBundle,
+	}
+	siteState, err := siteStateLoader.Load()
+
+	//If there is no site configured, the namespace needs to be removed
+	if err != nil || siteState == nil || siteState.Site == nil {
+		err = h.tearDownNamespace()
```
fgiorgetti, line 131: Here I believe we should not teardown the namespace if an error occurred.

For example, if you have a site named "mysite" running on the default namespace and you accidentally create another site named "mysite2", site state loader will fail, as 2 sites have been defined, which is acceptable.

But then, if you remove the mysite2, to fix the situation, it will teardown this running site, which is not desirable, leaving it in a bad state.

internal/nonkube/compat/site_state_renderer.go
nluaces: If a system controller can operate with different system platforms it would be really nice, I would prefer though to have this change implemented in a different pull request, what do you think?

internal/nonkube/controller/input_resource_handler.go
nluaces: I agree with you on keeping this change separately, we need to define as well the CLI improvement for this.

internal/qdr/amqp_mgmt.go
nluaces: https://github.com/skupperproject/skupper/issues/2380

internal/nonkube/controller/input_resource_handler.go
nluaces: fixed!

internal/nonkube/controller/input_resource_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	inputSitesMatchRuntimeSites := utils.StringSlicesEqual(inputSiteNames, runtimeSiteNames)
+
+	if !inputSitesMatchRuntimeSites {
+		needsBootstrap = true
+	} else {
+		siteState, err = h.siteStateLoader.Load()
+		if err != nil || siteState == nil {
+			needsBootstrap = true
```
fgiorgetti: If an error occurred, should it initialize the site?

internal/nonkube/controller/namespace_controller.go
```diff
@@ -38,9 +44,17 @@ func (w *NamespaceController) Start() {
 		routerStateHandler := NewRouterStateHandler(w.ns)
 		routerConfigHandler.AddCallback(routerStateHandler)
 		collectorLifecycleHandler := NewCollectorLifecycleHandler(w.ns)
-		routerStateHandler.SetCallback(collectorLifecycleHandler)
+		routerStateHandler.AddCallback(collectorLifecycleHandler)
+		inputResourceHandler := NewInputResourceHandler(w.ns, w.pathProvider.GetNamespace(), bootstrap.Bootstrap, bootstrap.PostBootstrap, bootstrap.Teardown)
+		systemAdaptorHandler := NewSystemAdaptorHandler(w.ns)
+		routerStateHandler.AddCallback(systemAdaptorHandler)
```
fgiorgetti: The `systemAdaptorHandler` should not be added as a callback when reload type is manual, as the `NewSystemAdaptorHandler` returns nil in such cases.

internal/nonkube/compat/site_state_renderer.go
```diff
@@ -38,12 +38,11 @@ func (s *SiteStateRenderer) Render(loadedSiteState *api.SiteState, reload bool)
...
 	s.loadedSiteState = loadedSiteState
 	endpoint := os.Getenv("CONTAINER_ENDPOINT")
-	if endpoint == "" {
-		endpoint = fmt.Sprintf("unix://%s/podman/podman.sock", api.GetRuntimeDir())
-		if s.Platform == "docker" {
-			endpoint = "unix:///run/docker.sock"
-		}
+
+	if api.IsRunningInContainer() || endpoint == "" {
+		endpoint = internalclient.GetDefaultContainerEndpoint()
```
fgiorgetti, line 43: The `GetDefaultContainerEndpoint()` is ignoring the `--platform` when I run `system start --platform docker`, so it returns a podman endpoint.

This behavior makes sense for existing sites, in cases of `reload`, `stop` and `uninstall`.

But for new sites (`system start`) using manual reload type, it should consider the `--platform`.

Thoughts?

internal/nonkube/client/compat/client.go
```diff
@@ -157,10 +158,20 @@ func NewCompatClient(endpoint, basePath string) (*CompatClient, error) {
 }
 
 func GetDefaultContainerEndpoint() string {
-	if os.Getenv(types.ENV_PLATFORM) == "docker" {
-		return "unix:///run/docker.sock"
+	platform := os.Getenv(types.ENV_PLATFORM)
```
fgiorgetti: It could also use the `CONTAINER_ENGINE` variable if still empty after checking the `SKUPPER_PLATFORM`.

internal/nonkube/bootstrap/teardown.go
fgiorgetti, line 34: Although not related to this PR, the removeDefinition does not seem to work for me.

The reason seems to be because it is running from a container when reload type = auto, therefore it should use `GetDefaultOutputPath` instead of `GetHostNamespaceHome`.

internal/nonkube/controller/input_resource_handler.go
nluaces: if there is no site definition, `h.siteStateLoader.Load()` returns a nil value and an error: 

https://github.com/skupperproject/skupper/blob/9ac367b3b24c86dc8bd4d88afbf36e7021612e54/internal/nonkube/common/site_state_loader.go#L52-L53

If that is the case, it needs to be bootstrapped, right?

fgiorgetti: My understanding is that the `h.siteStateLoader` is looking at the `input/resources` not at the `runtime/resources`. Is it right?

internal/nonkube/controller/input_resource_handler.go
nluaces: #2390

internal/nonkube/controller/namespace_controller.go
nluaces: nice catch! thank you!

internal/nonkube/compat/site_state_renderer.go
nluaces: I think this change has more implications; you could run 
`skupper system install --platform podman`
that creates the system-controller as a podman container and
`skupper system start --platform docker`
that creates the router as a docker container, and I don't think we want that scenario. 

In my opinion, if the system install command installs the controller in a selected platform, system start must not be able to choose a different platform. Do you agree on this?

internal/nonkube/bootstrap/teardown.go
nluaces: thank you!

internal/nonkube/controller/input_resource_handler.go
nluaces: That is correct. I will remove that part. Thanks!

internal/nonkube/compat/site_state_renderer.go
fgiorgetti: > In my opinion, if the system install command installs the controller in a selected platform, system start must not be able to choose a different platform. Do you agree on this?

Absolutely!

The issue I mentioned was related to the CLI itself, even when we don't have the system controller installed (meaning, everything is still manual). When the `--platform docker` is passed to a traditional bootstrap, it was not working.

Let me go through now and see if it is still happening.

fgiorgetti: It is still happening, check it out:

```
$ skupper --platform docker site create mysite --enable-link-access
File written to /home/fgiorget/.local/share/skupper/namespaces/default/input/resources/Site-mysite.yaml
File written to /home/fgiorget/.local/share/skupper/namespaces/default/input/resources/RouterAccess-router-access-mysite.yaml

$ skupper --platform docker system start
Sources will be consumed from namespace "default"
Site "mysite" has been created on namespace "default"
Platform: docker
Static links have been defined at: /home/fgiorget/.local/share/skupper/namespaces/default/runtime/links
Definition is available at: /home/fgiorget/.local/share/skupper/namespaces/default/input/resources

$ podman ps
CONTAINER ID  IMAGE                                COMMAND               CREATED        STATUS        PORTS                     NAMES
7ab9417de3e3  quay.io/skupper/skupper-router:main  /home/skrouterd/b...  6 seconds ago  Up 7 seconds  5671-5672/tcp, 55672/tcp  default-skupper-router
```

internal/nonkube/controller/input_resource_handler.go
fgiorgetti: I believe it still needs to teardown only when `err == nil && len(sites) ==0`

fgiorgetti: Maybe adding a return to the previous if (line:125) would help then here you just proceed with the tearDownNamespace when len(sites) == 0?

internal/nonkube/compat/site_state_renderer.go
nluaces: thanks! fixed now

internal/nonkube/client/compat/client.go
```diff
@@ -158,9 +159,9 @@ func NewCompatClient(endpoint, basePath string) (*CompatClient, error) {
 }
 
 func GetDefaultContainerEndpoint() string {
-	platform := os.Getenv(types.ENV_PLATFORM)
+	platform := config.GetPlatform()
```
fgiorgetti, line 168: Just before doing this, I'd recommend checking if the `CONTAINER_ENDPOINT` environment variable is set, if so, just return its value.

internal/nonkube/bootstrap/install.go
```diff
@@ -71,10 +72,16 @@ func Install(platform string) error {
...
+	if reloadType == "" {
+		reloadType = utils.DefaultStr(os.Getenv(types.ENV_SYSTEM_AUTO_RELOAD),
+			types.SystemReloadTypeManual)
+	}
+
 	env := map[string]string{
-		"CONTAINER_ENDPOINT":  config.containerEndpoint,
-		"SKUPPER_OUTPUT_PATH": config.hostDataHome,
-		"CONTAINER_ENGINE":    config.containerEngine,
+		"CONTAINER_ENDPOINT":         config.containerEndpoint,
```
fgiorgetti, line 81: The `CONTAINER_ENDPOINT` variable is incorrectly set on the skupper-controller container.
It should be pointing to:
* For Podman: /var/run/podman.sock (but it points to /run/user/1000/podman/podman.sock)
* For Docker: /var/run/docker.sock (but it points to /run/docker.sock)

internal/nonkube/bootstrap/install.go
```diff
@@ -202,14 +209,12 @@ func configEnvVariables(platform string) (*ControllerConfig, error) {
 		return nil, fmt.Errorf("Failed to create directory: %v", err)
 	}
 
-	containerEndpoint := os.Getenv("CONTAINER_ENDPOINT")
```
fgiorgetti, line 205: I believe you can keep this one. If the CONTAINER_ENDPOINT is set, use it, otherwise what you have looks correct.

internal/nonkube/bootstrap/teardown.go
```diff
@@ -18,15 +20,24 @@ type LocalData struct {
...
+
+	currentPlatform := config.GetPlatform()
+	if currentPlatform.IsKubernetes() {
+		currentPlatform = types.PlatformPodman
+	}
+	if string(currentPlatform) != configuredPlatform {
+		return fmt.Errorf("existing namespace uses %q platform and it cannot change to %q", configuredPlatform, string(currentPlatform))
+	}
+
+	if err := removeRouter(namespace, configuredPlatform); err != nil {
```
fgiorgetti, line 26: I believe that this is misplaced. When the system-controller is running with reload-type auto, the validation above will always fail, if the system-controller is running with docker.

If this gets moved to `internal/cmd/skupper/system/nonkube/system_stop.go (ValidateInput)`, I believe it should produce the desired effect.

internal/nonkube/bootstrap/install.go
nluaces: done

internal/nonkube/bootstrap/teardown.go
nluaces: done

PR #2369, opened 2026-01-29
Adding configurability to Prometheus helm chart

### LINKED ISSUE #2229: chart/network-observer: Control Prometheus configuration
Advanced users may want control over prometheus configuration for a number of reasons
- Setting storage parameters like retention, or backing storage
- Excluding certain metrics from collection (such as internal metrics or traffic metrics when not permissible.)
- Configuring a remote write

Expose values in the helm chart to allow users to overwrite:
- The prometheus container arguments ( e.g. for setting --storage.tsdb.retention.size )
- The prometheus data storage volume details (e.g. inserting a  volume backed by persistent storage)
- The prometheus configuration file contents

**Additional context**

JPadovano1483:
Exposes the prometheus container arguments and the prometheus configuration in the values.yaml file

Adds the ability to configure persistence for the prometheus volume and to create and mount new persistent volumes

Fixes #2229


charts/network-observer/values.yaml
```diff
@@ -25,6 +25,49 @@ prometheus:
...
+    # - --storage.tsdb.retention.size=10GB
+
+  # Persistence configuration for Prometheus data
+  persistence:
+    enabled: false
+    storageClass: ""
+    accessModes: ["ReadWriteOnce"]
+    size: 8Gi
+
+  # extra persistent volumes for Prometheus
```
c-kruse, line 40: extraVolumes doesn't seem like a bad idea, but I'm not sure I follow the reasoning for adding them. Did you have a use case in mind for these? Could there be some types of configurations that use an additional volume or something?

charts/network-observer/templates/deployment.yaml
```diff
@@ -86,13 +89,27 @@ spec:
...
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
       volumes:
       - configMap:
           defaultMode: 420
           name: {{ include "network-observer.fullname" . }}-prometheus
         name: prometheus-config
+      {{- if .Values.prometheus.persistence.enabled }}
+      - name: prometheus-storage-volume
+        persistentVolumeClaim:
```
c-kruse, line 106: Attaching this persistent volume in the deployment makes restarts consistently unreliable because we are using the default RollingUpdate deployment strategy type. At least with my local storage provider, that means that the replacement prometheus container will fail to acquire a filesystem lock - which is about the best case scenario.

The lightest lift would be to conditionally set the Deployment strategy type to `Recreate` conditionally when persistence is enabled.

```
[christian@rh skupper] (james-prom-config)$ k logs -l app.kubernetes.io/name=network-observer -c prometheus --previous
time=2026-02-02T18:06:11.900Z level=INFO source=main.go:1046 msg="Notify discovery manager stopped"
time=2026-02-02T18:06:11.900Z level=INFO source=manager.go:169 msg="Starting rule manager..." component="rule manager"
time=2026-02-02T18:06:11.900Z level=INFO source=main.go:1032 msg="Scrape discovery manager stopped"
time=2026-02-02T18:06:11.900Z level=INFO source=main.go:1079 msg="Scrape manager stopped"
time=2026-02-02T18:06:11.900Z level=INFO source=notifier.go:702 msg="Stopping notification manager..." component=notifier
time=2026-02-02T18:06:11.900Z level=INFO source=notifier.go:409 msg="Draining any remaining notifications..." component=notifier
time=2026-02-02T18:06:11.900Z level=INFO source=notifier.go:415 msg="Remaining notifications drained" component=notifier
time=2026-02-02T18:06:11.900Z level=INFO source=notifier.go:345 msg="Notification manager stopped" component=notifier
time=2026-02-02T18:06:11.900Z level=INFO source=main.go:1357 msg="Notifier manager stopped"
time=2026-02-02T18:06:11.901Z level=ERROR source=main.go:1367 msg="Fatal error" err="opening storage failed: lock DB directory: resource temporarily unavailable"
```


```
[christian@rh skupper] (james-prom-config)$ k describe pod -l app.kubernetes.io/name=network-observer
Name:             mynetworkobs-network-observer-5ffcd6ccb6-lhc9q
Namespace:        default
...
Containers:
  prometheus:
    ...
    State:          Running
      Started:      Mon, 02 Feb 2026 10:06:35 -0800
    Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Mon, 02 Feb 2026 10:06:11 -0800
      Finished:     Mon, 02 Feb 2026 10:06:11 -0800
    Ready:          True
    Restart Count:  3
...
Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Normal   Scheduled  25m                default-scheduler  Successfully assigned default/mynetworkobs-network-observer-5ffcd6ccb6-lhc9q to skupper-dev-control-plane
  Normal   Pulling    25m                kubelet            Pulling image "quay.io/skupper/network-observer:v2-dev"
  Normal   Pulled     25m                kubelet            Successfully pulled image "quay.io/skupper/network-observer:v2-dev" in 462ms (462ms including waiting). Image size: 19288889 bytes.
  Normal   Created    25m                kubelet            Container created
  Normal   Started    25m                kubelet            Container started
  Normal   Pulled     25m                kubelet            Container image "mirror.gcr.io/nginxinc/nginx-unprivileged:1.27.3-alpine" already present on machine and can be accessed by the pod
  Normal   Created    25m                kubelet            Container created
  Normal   Started    25m                kubelet            Container started
  Warning  BackOff    25m (x3 over 25m)  kubelet            Back-off restarting failed container prometheus in pod mynetworkobs-network-observer-5ffcd6ccb6-lhc9q_default(1b0c6961-ea12-47ec-990e-a15ac664bc77)
  Normal   Pulled     24m (x4 over 25m)  kubelet            Container image "quay.io/prometheus/prometheus:v3.1.0" already present on machine and can be accessed by the pod
  Normal   Created    24m (x4 over 25m)  kubelet            Container created
  Normal   Started    24m (x4 over 25m)  kubelet            Container started
```

JPadovano1483: Hmm, I ended up leaving this as a deployment because I wanted to leave as much of the existing configuration alone as I could while still solving the problem, but I think stateful sets handle this issue by default, so it may be a cleaner solution and help to avoid extra downtime from a Recreate deployment strategy. Any thoughts on which would be the best way forward with this?

charts/network-observer/values.yaml
JPadovano1483: I added these because I thought being able to mount extra PVs was part of the requirements for this issue, but that may have been a misunderstanding on my part. I think there are potentially some configurations that would use additional volumes, but I'm not sure whether or not they are things we are looking to add. I think one possible use for adding additional volumes could be for supporting credentials or certificates for things like configured remote writes. If this is outside the scope of what were looking for though, this functionality could get removed.

charts/network-observer/templates/deployment.yaml
c-kruse: @JPadovano1483 For now I agree: let's stick with the deployment and just change the strategy.

charts/network-observer/values.yaml
c-kruse: That makes sense, and those use cases (certificates, credentials) sound like something we may need to add someday. I would not really expect to see someone implement adding credentials to a pod with a PV, though. I'd expect to see something like a Secret or a CSI volume mounted instead.

What you've got is actually really close to solving for this if you wanted to adapt it (drop the extra PVC resource definitions and just drop the extraVolumes directly into the deployment spec instead of being rigid about it being a presistentVolumeClaim.

EDIT: So I'd call this one up to you: drop it entirely and call it a day or call it a discovered requirement and allow flexible definitions of extraVolumes.

JPadovano1483: Great, that all makes sense. Thanks!

Issue #2301, opened 2025-10-22
Add MultiKeyListener Support


c-kruse:
## Summary

Add support for  a new generic multi-routing-key Listener type and an initial priority strategy.

## Motivation

Presently the Listener configuration has exactly one routing key. Service traffic to that Listener is routed by the skupper router network to one of the reachable Connector locations with a matching routing key using a link-cost biased least-connected load balancing strategy. For many use cases this approach is insufficient, and can greatly increase deployment complexity (e.g. deploying a proxy in front of multiple listener services.)

By introducing a multi-routing-key Listener type, users are enabled to specify Connectors in a granular or hierarchical fashion that suits their use case without losing the ability to expose those Connectors from a single listener address.

## Proposal

Add a new resource: **skupper.io/v2alpha1 MultiKeyListener**

MultiKeyListenerSpec will keep identical semantics to the ListenerSpec for the following fields: `host`, `port`, `tlsCredentials`. MultiKeyListenerSpec will omit the `exposePodsByName` field, which will remain applicable only for Listener types. Instead of a `routingKey` field, MultiKeyListenerSpec will contain a `strategy` object of type `MultiKeyListenerStrategy` where multiple routingKeys and any additional routing parameters are configured.

MultiKeyListenerStrategy will have fields for each strategy type and will enforce exactly-one semantics. Initially there will be a single option: Priority, but should be ready for extension with additional strategies.

- `priority PriorityStrategySpec`

PriorityStrategySpec will contain an array of routingKey strings in preference order.

## Example Use Case
A service is deployed in many AZs. At a particular Site, the user prefers all traffic be routed exclusively to the nearest available geographic instances. The user configures overlapping Connector definitions for routingKeys "backend-all" and "backend-{{ AZ }}" then configures a MultiKeyListener with PriorityFailover in preference order.

```
apiVersion: skupper.io/v2alpha1
kind: MultiKeyListener
metadata:
  name: ex01
spec:
  host: example-01
  port: 8000
  strategy:
    priority:
      routingKeys:
      - backend-us-east-1
      - backend-us-east-2
      - backend-all
```

PR #2372, opened 2026-02-04
Add multikeylistener

c-kruse:
Implements  https://github.com/skupperproject/skupper/issues/2301

Extends skupper with a new MultiKeyListener CRD. Incompatible with skupper-router builds without listenerAddress support.

### DIFF
--- a/internal/kube/site/extended_bindings.go
+++ b/internal/kube/site/extended_bindings.go
@@ -194,7 +194,7 @@ func (a *ExtendedBindings) updateBridgeConfigForMultiKeyListener(siteId string, 
 	if a.mapping == nil {
 		a.mapping = qdr.RecoverPortMapping(nil)
 	}
-	port, err := a.mapping.GetPortForKey(mkl.Name)
+	port, err := a.mapping.GetPortForKey(multiKeyListenerPortName(mkl.Name))
 	if err != nil {
 		bindings_logger.Error("Could not allocate port for multikeylistener",
 			slog.String("namespace", mkl.Namespace),
@@ -204,8 +204,12 @@ func (a *ExtendedBindings) updateBridgeConfigForMultiKeyListener(siteId string, 
 	site.UpdateBridgeConfigForMultiKeyListenerWithHostAndPort(siteId, mkl, "", port, config)
 }
 
+func multiKeyListenerPortName(name string) string {
+	return "multiaddress-" + name
+}
+
 func (a *ExtendedBindings) multiKeyListenerUpdated(mkl *skupperv2alpha1.MultiKeyListener) {
-	allocatedRouterPort, err := a.mapping.GetPortForKey(mkl.Name)
+	allocatedRouterPort, err := a.mapping.GetPortForKey(multiKeyListenerPortName(mkl.Name))
 	if err != nil {
 		bindings_logger.Error("Unable to get port for multikeylistener",
 			slog.String("namespace", mkl.Namespace),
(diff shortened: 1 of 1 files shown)


internal/qdr/qdr.go
```diff
@@ -553,15 +580,39 @@ type Address struct {
...
+	Observer             string `json:"observer,omitempty"`
+	VerifyHostname       *bool  `json:"verifyHostname,omitempty"`
+	ProcessID            string `json:"processId,omitempty"`
+	MultiAddressStrategy string `json:"multiAddressStrategy,omitempty"`
+	AuthenticatePeer     bool   `json:"authenticatePeer,omitempty"`
+}
+
+type ListenerAddress struct {
+	Name     string `json:"name,omitempty"`
+	Address  string `json:"address,omitempty"`
```
gabordozsa: I'm just curious why all fields except `Value` can be omitted.` Address` and `Listener `are both required, too, in the router management schema.

internal/kube/site/extended_bindings.go
```diff
@@ -174,13 +180,90 @@ func (a *ExtendedBindings) updateBridgeConfigForListener(siteId string, listener
 	if a.mapping == nil {
 		a.mapping = qdr.RecoverPortMapping(nil)
 	}
-	if port, err := a.mapping.GetPortForKey(listener.Name); err == nil {
-		site.UpdateBridgeConfigForListenerWithHostAndPort(siteId, listener, "", port, config)
-	} else {
+	port, err := a.mapping.GetPortForKey(listener.Name)
```
fgiorgetti, line 183: Minor issue and most unlikely to happen, but if you have a MultiKeyListener named `backend` and for any reason you define a Listener named `multiaddress-backend`, the returned port will be the same returned for the `backend` MKL.
Adding a prefix to listener port keys, like `listener-` would require an update to all existing k8s service port names.
Not a big deal, but something to keep in mind.

c-kruse: Hi! Yeah, that is true. I added the prefix for the very likely collision between a listener and mkl with the same name.

I briefly explored the idea of a more ideal solution but never quite nailed it down: maybe something like just using the listener/mkl name for port name then resolving conflicts by appending a short hash of the resource uid or something.

config/crd/bases/skupper_multikeylistener_crd.yaml
```diff
@@ -0,0 +1,230 @@
...
+    plural: multikeylisteners
+    singular: multikeylistener
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - description: The status of the multikeylistener
+      jsonPath: .status.status
+      name: Status
+      type: string
+    - description: Any human reandable message relevant to the multikeylistener
```
pwright, line 22: typo
```suggestion
    - description: Any human-readable message relevant to the multikeylistener
```

config/crd/bases/skupper_multikeylistener_crd.yaml
```diff
@@ -0,0 +1,230 @@
...
+                  properties:
+                    lastTransitionTime:
+                      description: |-
+                        lastTransitionTime is the last time the condition transitioned from one status to another.
+                        This should be when the underlying condition changed.  If that is not known, then using the time when the API field changed is acceptable.
+                      format: date-time
+                      type: string
+                    message:
+                      description: |-
+                        message is a human readable message indicating details about the transition.
```
pwright, line 143:
```suggestion
                        message is a human-readable message indicating details about the transition.
```

config/crd/bases/skupper_multikeylistener_crd.yaml
```diff
@@ -0,0 +1,230 @@
...
+                        maxItems: 256
+                        minItems: 1
+                        type: array
+                        x-kubernetes-list-type: set
+                    required:
+                    - routingKeys
+                    type: object
+                type: object
+                x-kubernetes-validations:
+                - message: exactly one of the fields in [priority] must be
```
pwright, line 110: Dunno if this is where you specify the strategy?

Exactly one routing strategy (e.g., priority) must be set

pwright (commented):
just some comments, small typo

PR #2374, opened 2026-02-09
added additional network observer /metrics endpoint

### LINKED ISSUE #2194: chart/network-observer: openshift auth with external ServiceMonitor/PodMonitor
**Describe the bug**
Running the skupper network-observer with the openshift auth proxy, it is difficult to configure external metrics collection.

```
# values.yaml
auth:
  strategy: openshift
tls:
  skupperIssued: false
  openshiftIssued: true
route:
  enabled: true
```

The network-observer pod exposes just port 443 - which is protected by the oauth proxy complicating collection. Internally to the pod, the observer exposes an HTTP API on port 8080 and binds to localhost only (not externally accessible.)

Image

**How To Reproduce**

**Expected behavior**
There should be a clear way to mitigate this.

The most obvious workaround is to change the deployment such that the network-observer container binds to 0.0.0.0:8080 (instead of the default 127.0.0.1:8080.)

```
# values-workaround.yaml
extraArgs:
  - -listen=:8080

auth:
  strategy: openshift
tls:
  skupperIssued: false
  openshiftIssued: true
route:
  enabled: true
```

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
spec:
  podMetricsEndpoints:
  - port: api
  selector:
    matchLabels:
      app.kubernetes.io/name: network-observer
```

Image

**Environment details**
 - Skupper Operator 2.1
 - Platform: Openshift

JPadovano1483:
added a metrics endpoint to the network-observer container to allow external metrics collection

Fixes #2194

### DIFF
--- a/charts/network-observer/templates/service.yaml
+++ b/charts/network-observer/templates/service.yaml
@@ -25,7 +25,7 @@ metadata:
   name: {{ include "network-observer.fullname" . }}-metrics
   labels:
     {{ include "network-observer.labels" . | nindent 4 }}
-    app: metrics
+    app.kubernetes.io/component: metrics
 spec:
   type: ClusterIP
   ports:
--- a/cmd/network-observer/main.go
+++ b/cmd/network-observer/main.go
@@ -252,7 +252,7 @@ func main() {
 
 	flags.StringVar(&cfg.VanflowLoggingProfile, "vanflow-logging-profile", "silent", "Controls low level vanflow record logging. Options are silent, minimal, moderate and all")
 
-	flags.StringVar(&cfg.MetricsListenAddress, "listen-metrics", ":9000", "The address that the Metrics Server will listen on.")
+	flags.StringVar(&cfg.MetricsListenAddress, "listen-metrics", "", "The address that the Metrics Server will listen on.")
 
 	flags.Parse(os.Args[1:])
 	if *isVersion {


cmd/network-observer/main.go
```diff
@@ -219,6 +252,8 @@ func main() {
 
 	flags.StringVar(&cfg.VanflowLoggingProfile, "vanflow-logging-profile", "silent", "Controls low level vanflow record logging. Options are silent, minimal, moderate and all")
 
+	flags.StringVar(&cfg.MetricsListenAddress, "listen-metrics", ":9000", "The address that the Metrics Server will listen on.")
```
c-kruse: Suggest defaulting this to "" so that it is off by default.

charts/network-observer/templates/service.yaml
```diff
@@ -17,3 +18,20 @@ spec:
...
   selector:
     {{- include "network-observer.selectorLabels" . | nindent 4 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "network-observer.fullname" . }}-metrics
+  labels:
+    {{ include "network-observer.labels" . | nindent 4 }}
+    app: metrics
```
c-kruse: `app` does not seem quite right to me. Maybe `app.kubernetes.io/component` instead?

c-kruse (approved):
Very nice!

Deployed to openshift, added a ServiceMonitor, and it worked great.
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: network-observer-metrics
spec:
  endpoints:
  - interval: 30s
    port: metrics
  selector:
    matchLabels:
      app.kubernetes.io/component: metrics
      app.kubernetes.io/part-of: skupper-network-observer
```

PR #2375, opened 2026-02-10
Normalize link cost to 1 to stop connector churn.

### LINKED ISSUE #2373: Missing Link Cost validation and defaulting logic
**Describe the bug**
The skupper Link resource's `spec.cost` field is an optional integer that is today mapped directly into a skupper-router connector configuration entity. In the router, a connector's cost must be an integer greater than 0. This gap leaves us open to two problems.

* When cost is omitted or set to `0` the skupper-router defaults it to `1`. The kube-adaptor detects drift between desired and actual connector configuration and recreates the connector - churning connections.
* When cost is explicitly set to a negative number there are two failure modes:
  * The management agent of a RUNNING skupper-router defaults cost to `1` and the same kube-adaptor behavior above churns connector configuration.
  * The router fails to start loading the configuration from disk with stricter validations. `Invalid cost (-1) specified. Minimum value for cost is 1 and maximum value is 2147483647`

**How To Reproduce**
- Generate a link or token with cost unset. This is the default behavior using the skupper ansible token module.
- Apply this link and watch the router and kube-adaptor logs for evidence of this pattern
- Update the link cost to a negative number
- Watch the router pod logs for continued evidence of connector config churn
- Restart the router deployment and see a CrashBackOff Loop.

**Environment**
Kube

Versions: mainly v2-dev - an unrelased change makes Cost=0 unstable.

**Expected behavior**
TBD

* A Link with no cost set, or cost=0 resolves to a stable router connector with cost=1
* A link with cost set less than 0 is put in an error state, is removed from router configuration and the CONFIGURED status condition reflects that it is not configured.

AryanP123:
Fixes #2373

### DIFF
--- a/internal/qdr/qdr.go
+++ b/internal/qdr/qdr.go
@@ -1102,10 +1102,16 @@ func (desired Connector) IsLinkConnector() bool {
 }
 
 func (desired Connector) Equivalent(actual Connector) bool {
+	normaliseCost := func(c int32) int32 {
+		if c < 1 {
+			return 1
+		}
+		return c
+	}
 	return desired.Name == actual.Name &&
 		desired.Host == actual.Host &&
 		desired.Port == actual.Port &&
-		desired.Cost == actual.Cost &&
+		normaliseCost(desired.Cost) == normaliseCost(actual.Cost) &&
 		desired.SslProfile == actual.SslProfile
 }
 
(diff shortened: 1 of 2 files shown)


c-kruse (approved):

internal/qdr/qdr.go
```diff
@@ -1104,10 +1102,16 @@ func (desired Connector) IsLinkConnector() bool {
 }
 
 func (desired Connector) Equivalent(actual Connector) bool {
+	normaliseCost := func(c int32) int32 {
+		if c == 0 {
```
fgiorgetti: What you have here works just fine, but maybe use the same logic you have below on link.go? 
If `c < 1` instead of `c == 0`.

Issue #2362, opened 2026-01-21
Proposal: Helm-First Resource Definitions


c-kruse:
This proposal makes the Skupper Helm chart the source of truth for deployment resources (Deployment, RBAC, ServiceAccount, etc.), replacing the previous source of truth embedded in the `skupper-deployment-generator.sh` script, operator bundle generator script, and config/rbac kustomize directory.

## Motivation

- **Clarity**: The source of truth for our resources is presently fractured: deployments and rbac resources are embedded in a shell script, and are duplicated for the operator bundle build.
- **Maintainability**: Injecting templates for helm into a plain YAML resources is difficult.
- **Flexibility**: Users have requested additional deployment options (tolerations, affinity, resource limits) and will eventually need guidance with alternate controller configurations (access types, grants server config, etc.)

Expands on https://github.com/skupperproject/skupper/issues/2351

### Architecture
```
├── charts
│   ├── skupper
│   │   └── Chart.yaml // Source of truth for deployment resources
│   └── skupper-crds
│       └── Chart.yaml // Generated crds-only chart (optional)
└── config
    ├── crd
    │   └── kustomization.yaml // Source of truth for CRD definitions
    └── samples
        └── kustomization.yaml // Source of truth for examples for operator bundle
```

* Add a source-controlled skupper chart
* Remove crds from skupper chart
* Refactor the deployment-generator script to use `helm template` and `kustomize` to produce install manifests
* Refactor the operator-bundle-generator script to use `helm template`, `kustomize` and `operator-sdk` to generate bundles.
* Extend skupper chart with community requested template values: tolerations, affinity, resources, etc.
* Source of truth for CRDs and samples remains in config tree.
* Add a crds-only manifest - generated with `kustomize ./config/crd`
* Add a skupper-crds Chart with templates generated from crds.

### Artifacts

Add:
- `skupper-crds.yaml` For cluster admins to install CRDs independently from a manifest file.
- `oci://quay.io/skupper/helm/skupper-crds` CRD only helm chart functionally identical to skupper-crds.yaml. For cluster admins to install CRDs independently but with helm involved.

Revised:
- `oci://quay.io/skupper/helm/skupper` skupper helm chart - backwards compatible with exception for the removal of CRDs.

### Kustomize-First

An alternative approach would be to make kustomize the canonical source by moving deployment specs into `./config` a la kubebuilder and operator framework. The Helm chart would either continue to be generated, or resource definitions would be duplicated.

| Pros | Cons |
[...]

Issue #2306, opened 2025-11-03
Toleration support in Controller helm chart and Site resource for router pod


jlo-norlysenergytrading:
Summary
In clusters where nodes are tainted, Skupper’s router pods created from a Site resource remain in Pending because the generated Deployment does not include tolerations. Similarly, the Helm chart for the controller does not expose tolerations. This makes dynamic deployment via CI/CD difficult in environments where taints are standard.
Proposal

Add tolerations to:

The Helm chart for the controller.
The Site CR spec so the controller can apply them to the router Deployment.

Why
Without this, Skupper cannot run on clusters where all nodes have NoSchedule taints without manual patching or admission webhooks.
Example
```
YAMLspec:  
  tolerations:
      - key: "dedicated"
         operator: "Equal"     
         value: "networking"
         effect: "NoSchedule"

```
Questions for maintainers

Is this something you’d like to support in Skupper?
Would you prefer a minimal tolerations field or a full podTemplate approach for future flexibility?

Contribution offer
This would be my first contribution to any open source project. I’m happy to give this a go if you think it’s a good fit and can point me to any development guidelines

c-kruse:
Hi @jlo-norlysenergytrading. Thank you for the issue and for the interest in contributing!

Sounds like you could already be ahead of me here, but I'll state that there is a workaround for users finding this Issue. You should be able to patch the deployments after the fact: we intentionally used server-side apply in the skupper controller so that we would not conflict with external controllers and manually interventions like this (at least for the skupper router Deployments.)

I'll split my thoughts into two areas here: the controller deployment (helm/manifest yaml/kustomize) and the skupper-router deployment (Site CR).

## Controller Deployment

I agree that this would be a nice thing to have. In an ideal world we'd support a Helm chart that exposes some of the deployment configurations like this along with several of our controller-specific configurations that are mostly left up to the user to discover and configure today.

You'll probably notice looking at our skupper chart that we have not fully embraced Helm. This mainly comes down to a lack of resources on our end - we wanted to avoid dual maintenance between deployment resource templates (helm) and a kustomize-style layout we use for generating the static manifest yamls and the OLM operator.

For now, I'd welcome one (or both) of these approaches:
* A new community maintained Helm chart outside the skupper repository to evaluate interest.
* Better documentation on installing our manifests with custom overlays using Kustomize.

## Router Deployment (Site CR)

This is something we are broadly struggling with, and are open to proposals.

[...]

PR #2376, opened 2026-02-11
Use helm chart as source of install resources

### LINKED ISSUE #2351: Manage Helm chart as code instead of generating it
Today I wanted to make a contribution to your Helm chart. We need to set resources (CPU/memory) on the skupper controller deployment, but the Helm chart currently does not allow for that. Apparently the current Helm chart is generated using multiple `.sh` scripts, in which yamls are defined and later adjusted with some `sed` commands that allow the image name to be overwritten.

As much as I love `bash` and `sed`, this setup makes it very cumbersome to propose changes that go further than simple search and replace actions. If this remains as is, I will find a way, but I'd rather not go there.

The Helm chart should be available as code in GitHub, making it easier to contribute; especially for people not familiar with the current scripting.

N/A. A word on whether this will be considered/implemented in the short term would be appreciated, if necessary I will still make the `resources` contribution in the current setup, as we need that to stabilise our deployment.

**Additional context**
- https://github.com/skupperproject/skupper/blob/main/scripts/skupper-deployment-generator.sh
- https://github.com/skupperproject/skupper/blob/main/scripts/skupper-helm-chart-generator.sh

c-kruse:
Adds the skupper helm chart to version control to be the primary source of Skupper controller installation resources. Replaces the skupper-deployment-generator scripts as the source of truth.

Changes:
- Breaking: Drops CRDs from skupper chart.
- Adds new skupper-crds.yaml manifest release artifact
- Adds skupper-crds helm chart
- Updates skupper chart README with CRD installation instructions
- Extends the skupper chart with templated values for user requested overrides: tolerations, affinity, resources, labels+annotations.
- Update charts/README.md with quick start guide
- Update kind-dev-cluster script to use helm install

Part of https://github.com/skupperproject/skupper/issues/2362 and https://github.com/skupperproject/skupper/issues/2306
Closes https://github.com/skupperproject/skupper/issues/2351


.circleci/config.yml
JPadovano1483, line 400: I think this line can get removed since the makefile command and the associated script were removed.

PR #2381, opened 2026-02-12
Prefix tcpListener/tcpConnector names to avoid same-site collision with exposePodsByName. Added test in controller_test.go.

### LINKED ISSUE #2058: Listener and Connector in same site incompatible with exposePodsByName
**Describe the bug**
When a Listener and Connector are added to the same site with exposePodsByName the control plane attempts to create tcpListener and tcpConnector records with the same name, violating the unique constraint in router configuration.

**How To Reproduce**

**Expected behavior**

**Environment details**
 - Skupper CLI: [e.g. 1.5.0, 1.4.3]
 - Platform: [e.g kubernetes, openshift]

**Additional context**

AryanP123:
Fixes #2058

### DIFF
--- a/internal/kube/controller/controller_test.go
+++ b/internal/kube/controller/controller_test.go
@@ -258,7 +258,7 @@ func TestGeneral(t *testing.T) {
 				resource.DeploymentResource(): f.routerDeployment("skupper-router", "test"),
 			},
 			expectedRouterConfig: []*RouterConfig{
-				f.routerConfig("skupper-router", "test").tcpListener("mylistener", "1024", "", ""),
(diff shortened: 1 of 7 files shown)


c-kruse (commented):
Nice! This was a lot cleaner than I had initially thought.

I know it was me who suggested a `resource-type.resource-name` scheme, but I'd like to walk that back. We should definitely pick something that cannot be a in a k8s resource name. I'd lean towards `/`.

Also, the catch here that we should call out is that this does end up bubbling back to the user in `site.status.network.services.listeners`. Maybe not a deal breaker, but something we should be careful to consider and maybe document.

Thoughts on that @fgiorgetti?

PR #2386, opened 2026-02-23
Create must-gather image on new releases

granzoto:
Creates a must-gather image when a new release is created

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,6 +166,10 @@ workflows:
             - generate-network-observer-operator-bundle
           context:
             - skupper-org
+      - generate-must-gather:
+          <<: *run_for_numeric_tags
+          requires:
+            - publish-oci-images
   build:
     jobs:
       - build-all:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -166,6 +166,10 @@ workflows:
             - generate-network-observer-operator-bundle
           context:
             - skupper-org
+      - generate-must-gather:
+          <<: *run_for_numeric_tags
+          requires:
+            - publish-oci-images
```
nluaces, line 172: oh, I see now: this step is going to be executed in parallel with all these steps:
            - generate-manifest
            - generate-skupper-setup
            - generate-operator-bundle
            - generate-network-observer-operator-bundle

So the draft release can be created even in the remote case that this step fails:

```
      - publish-github-release-artifacts:
          <<: *run_for_numeric_tags
          requires:
            - publish-oci-images
            - generate-manifest
            - generate-skupper-setup
            - generate-operator-bundle
            - generate-network-observer-operator-bundle
          context:
            - skupper-org
```

PR #2387, opened 2026-02-25
Recreate listener Service when deleted

### LINKED ISSUE #2383: Listener service should be re-created if deleted
**Describe the bug**
The listener resource creates a service in the namespace when it is defined in. If the service is deleted (e.g. by a user) the service should be restored.

**How To Reproduce**
Hello-world example, delete the frontend kubernetes services in the listener namespace.

**Expected behavior**
If a dependent listener service is deleted, it should be restored.

**Environment details**
detected using v2.1.3 controller

AryanP123:
Fixes #2383

### DIFF
--- a/internal/kube/controller/controller_test.go
+++ b/internal/kube/controller/controller_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/skupperproject/skupper/internal/kube/resource"
 	"github.com/skupperproject/skupper/internal/network"
 	"github.com/skupperproject/skupper/internal/qdr"
+	"github.com/skupperproject/skupper/internal/utils"
 	"github.com/skupperproject/skupper/internal/version"
 	skupperv2alpha1 "github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 )
@@ -625,6 +626,18 @@ func TestUpdate(t *testing.T) {
 				negativeServiceCheck("mysvc", "test"),
 			},
 		}, {
+			name: "listener service recreated when deleted",
+			skupperObjects: []runtime.Object{
+				f.site("mysite", "test", "", false, false),
+				f.listener("mylistener", "test", "mysvc", 8080),
+			},
+			functions: []WaitFunction{
+				isListenerStatusConditionTrue("mylistener", "test", skupperv2alpha1.CONDITION_TYPE_CONFIGURED),
+				serviceCheck("mysvc", "test").check,
+				deleteListenerService("mysvc", "test"),
+				waitForService("mysvc", "test"),
+			},
+		}, {
 			name: "exposePodsByName handles pod delete",
 			k8sObjects: []runtime.Object{
 				f.skupperNetworkStatus("test", f.networkStatusInfo("mysite", "test", map[string]string{"mysvc": "mysvc", "mysvc.mypod-1": "mysvc@mypod-1", "mysvc.mypod-2": "mysvc@mypod-2"}, map[string]string{"mysvc": "mysvc", "mysvc.mypod-1": "10.1.1.10", "mysvc.mypod-2": "10.1.1.11"}).info()),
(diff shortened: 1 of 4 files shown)


internal/kube/site/site.go
```diff
@@ -936,6 +936,12 @@ func (s *Site) CheckListenerService(svc *corev1.Service) error {
 	return nil
 }
 
+func (s *Site) EnsureListenerService(listener *skupperv2alpha1.Listener) {
+	if listener != nil && s.site != nil {
+		s.bindings.ListenerUpdated(listener)
```
ajssmith: Calling ListenerUpdated results in query to api server (via Expose) which would be good to avoid if possible.

Suggest looking at updating CheckListenerService to handle the delete case (e.g. svc == nil) and possibly use isHostExposed to distinguish deletion due to listener deletion versus user deletion. 

The question will be whether there is enough context available to re-create the user deleted service (E.g. could isHostExposed return ports, etc. If not, maybe storing what is needed in a site scoped cm that tracks the service lifecycle might be an alternative.

fgiorgetti: The `HandleDeletedListenerService` seems to be doing what is needed.
Is `EnsureListenerService` really needed?

AryanP123: Ah, you're right. Didn't notice the redundancy, will fix. Thanks!

internal/kube/controller/controller_test.go
```diff
@@ -839,6 +813,41 @@ func TestUpdate(t *testing.T) {
...
+		ctrl.eventProcessor.TestProcess()
+	}
+	for !isListenerStatusConditionTrue("mylistener", "test", skupperv2alpha1.CONDITION_TYPE_CONFIGURED)(t, clients) {
+		ctrl.eventProcessor.TestProcess()
+	}
+	for !serviceCheck("mysvc", "test").check(t, clients) {
+		ctrl.eventProcessor.TestProcess()
+	}
+	deleteListenerService("mysvc", "test")(t, clients)
+	for !waitForService("mysvc", "test")(t, clients) {
```
fgiorgetti: Better to use a retry mechanism here (i.e.: `internal/utils/retry.go`).

internal/kube/site/extended_bindings.go
```diff
@@ -103,30 +103,41 @@ func (a *ExtendedBindings) ListenerUpdated(listener *skupperv2alpha1.Listener) {
...
+		Port:       listener.Spec.Port,
+		TargetPort: allocatedRouterPort,
+		Protocol:   listener.Protocol(),
+	}
+	exposed := a.exposed.Expose(listener.Spec.Host, port)
+	if exposed == nil {
+		if existing, ok := a.exposed[listener.Spec.Host]; ok {
+			exposed = existing
 		}
-		if exposed := a.exposed.Expose(listener.Spec.Host, port); exposed != nil {
```
fgiorgetti, line 119: The original code was only going to call context.Expose if a change is needed,
now it seems to be calling it either way. I see that the `context.Expose` has mechanisms
to detect if something has changed or not, but once it is called it tries to retrieve the latest
service from the kube api.
Did you find a reason to change this (sorry if it is obvious but I am not seeing it)?

AryanP123: So the only effect of the change is to recreate the Service earlier in the case where a Listener event is processed before the Service delete event, mainly added it as a defensive path during implementation. It does not add a path that’s required for correctness, the delete path (HandleDeletedListenerService) already guarantees recreation. Looking at it again, I do suppose this change here is unnecessary, so I can revert this. Thanks for pointing it out.

fgiorgetti (approved):
Works fine.
Just a suggestion to consider indexing the leadListeners map by `host/port` instead, to make it simpler. But no objection to the current approach.

PR #2389, opened 2026-02-26
detect listener config conflicts

### LINKED ISSUE #2379: Skupper allows the misconfiguration of listeners that have colliding host/port combinations.
**Describe the bug**
If I create two listeners in the same namespace with two different routing keys but the same host/port combination, this is an error condition but no indication is provided that I've created a misconfiguration.

**How To Reproduce**
In a skupper-site namespace, create two listeners:

```
kind: Listener
metadata:
  name: listener-1
spec:
  routingKey: key-1
  host: listener
  port: 8080
---
kind: Listener
metadata:
  name: listener-2
spec:
  routingKey: key-2
  host: listener
  port: 8080
```

Create two connectors to match:

```
kind: Connector
metadata:
  name: connector-1
spec:
  routingKey: key-1
  host: domain.com
  port: 8080
---
kind: Connector
metadata:
  name: connector-2
spec:
  routingKey: key-2
  host: domain.com
  port: 8080
```
Now look at the listener status:

```
$ kubectl get listeners
NAME         ROUTING KEY   PORT   HOST       STATUS   HAS MATCHING CONNECTOR   MESSAGE
listener-1   key-1         8080   listener   Ready    true                     OK
listener-2   key-2         8080   listener   Ready    true                     OK
```
Yet there is only one service:

```
$ kubectl get services
NAME                   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
listener               ClusterIP   172.21.214.84    <none>        8080/TCP   15m
skupper-router-local   ClusterIP   172.21.131.101   <none>        5671/TCP   18m
```
That service references the listener port of only one of the listeners.  The other is left hanging and will never be used.

**Expected behavior**
In this case, I expect at least one of the listeners to have a status that indicates a host/port collision, rather than giving the impression that it is "ready" and working normally.

**Environment details**
 - Skupper CLI: 2.1.3
 - Skupper Operator (if applicable): 2.1.3
 - Platform: Openshift

ajssmith:
fixes #2379

### DIFF
--- a/internal/kube/site/site.go
+++ b/internal/kube/site/site.go
@@ -27,6 +27,7 @@ import (
 	"github.com/skupperproject/skupper/internal/qdr"
 	"github.com/skupperproject/skupper/internal/site"
 	"github.com/skupperproject/skupper/internal/version"
+	"github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 	skupperv2alpha1 "github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 )
 
@@ -62,6 +63,7 @@ type Site struct {
 	labelling     Labelling
 	profiles      *secrets.ProfilesWatcher
 	disableSecCtx bool
+	leadListeners map[string]string
 }
 
 func NewSite(namespace string, eventProcessor *watchers.EventProcessor, certs certificates.CertificateManager, access SecuredAccessFactory, sizes *sizing.Registry, labelling Labelling, disableSecCtx bool) *Site {
@@ -82,6 +84,7 @@ func NewSite(namespace string, eventProcessor *watchers.EventProcessor, certs ce
 		),
 		labelling:     labelling,
 		disableSecCtx: disableSecCtx,
+		leadListeners: map[string]string{},
 	}
 	site.profiles = secrets.NewProfilesWatcher(
 		sslSecretsWatcher(namespace, eventProcessor),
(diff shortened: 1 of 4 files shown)


internal/kube/site/site.go
```diff
@@ -114,6 +117,14 @@ func (s *Site) verifySiteSpec(site *skupperv2alpha1.Site) error {
 	return nil
 }
 
+func (s *Site) CheckLeadListeners(listener *skupperv2alpha1.Listener) bool {
```
AryanP123, line 120: Should we only add a listener to leadListeners if that host/port isn’t already claimed? Right now both colliding listeners get added in the first pass, so in the second pass neither is marked as colliding  in CheckListener.

fgiorgetti: Today when a Listener is created, the resulting Kubernetes Service is labeled with `internal.skupper.io/listener = <listener-name>`. Actually this is already inefficient today because a Kubernetes service can result of one or more listeners (with same hosts, but different ports).

But the label above has the benefit of determining the source(s) of a service.

Maybe we could, instead, enhance this existing label and store a CSV like: `listener1:port1,listener2:port2`.
This would help preserving the same state we had before the controller is restarted.

fgiorgetti (approved):
Works fine.
Just a suggestion to consider indexing the leadListeners map by host/port instead, to make it simpler. But no objection to the current approach.

nluaces (approved):

PR #2409, opened 2026-03-18
Configurable ingressClassName and ingress access type

### LINKED ISSUE #2397: Add support for specifying IngressClass/IngressController in Skupper-managed Ingress resources
In clusters with more than one IngressController (e.g., a private and a public controller), Skupper v2 always defaults to using the cluster's default IngressController and provides no way to specify which IngressClass or IngressController should be used. 

The relevant code seems to be in `internal > kube > securedaccess > ingress.go > func toIngress()` which has no logic to specify an `ingressClassName` field in the ingress `spec` stanza.

Expose a configuration option — such as a Custom Resource (CR) field, CLI flag, or environment variable — that allows users to explicitly set the desired ingressClassName for Skupper-managed Ingress resources. 

This may also require a new AccessType of "ingress" (in addition to the existing "ingress-nginx"). This was possible in Skupper v1 and seems to  have been removed in v2.

Manually editing the generated Ingress resource's spec field to set ingressClassName to the desired class appears like it will be overwritten on every reconciliation cycle by Skupper, making it an unviable workaround.

**Additional context**

Skupper CLI version: 2.1
Skupper Operator version: 2.1
Platform: OpenShift
Example cluster setup with multiple IngressControllers:

```
% oc get ingressclasses --all-namespaces
NAME CONTROLLER PARAMETERS AGE
openshift-default openshift.io/ingress-to-route IngressController.operator.openshift.io/default 115d
openshift-public-ingress-controller openshift.io/ingress-to-route IngressController.operator.openshift.io/public-ingress-controller 115d
In this scenario, Skupper selects openshift-default (mapped to the private controller) instead of the intended public IngressController, blocking external connectivity.
```

AryanP123:
Fixes #2397

### DIFF
--- a/internal/kube/site/site.go
+++ b/internal/kube/site/site.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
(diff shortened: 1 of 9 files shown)


internal/kube/site/site.go
```diff
@@ -277,6 +278,47 @@ func (s *Site) groups() []string {
 	}
 }
 
+// routerAccessSettingsMerge copies existing RouterAccess settings, then applies
+// site.spec.settings.ingressClassName when that key is present (non-empty sets,
```
fgiorgetti, line 282: If the `ingressClassName` is provided through a RouterAccess and through the Site, the one provided through the Site resource is overwriting this one. Is this the desired behavior?

fgiorgetti (commented):
One minor comment, but it seems to work.
Not sure, though, how it is effectively going to work, since we do not set controller specific settings for ssl-passthrough other than nginx.

internal/kube/site/site.go
AryanP123: My thought process was that we still want users to configure it on the Site (with linkAccess), so we copy `Site.spec.settings.ingressClassName` into the controller-managed RouterAccess. Without that, the Site field would have no effect on the Ingress. When the key is absent on the Site, we don’t change whatever is already on RouterAccess. Let me know your thoughts.

nluaces (approved):
Looks good to me! I would wait for @fgiorgetti review though.

internal/kube/site/site.go
fgiorgetti: It is acceptable, but better to raise a skupper-docs issue to explain it properly through the docs.

fgiorgetti (approved):
I'd recommend merging this after 2.2.0 is out.
WDYT @AryanP123 and @nluaces ?

PR #2414, opened 2026-03-26
Raise default REST client QPS/burst when unset

AryanP123:
Fixes #2412

### DIFF
--- a/internal/kube/client/client.go
+++ b/internal/kube/client/client.go
@@ -3,6 +3,9 @@
 package client
 
 import (
+	"os"
+	"strconv"
+
 	openshiftapps "github.com/openshift/client-go/apps/clientset/versioned"
 	openshiftroute "github.com/openshift/client-go/route/clientset/versioned"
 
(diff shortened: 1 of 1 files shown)


AryanP123:
Context:

With many Listeners/Connectors, readiness lagged because rest.Config had QPS/Burst = 0, so client-go used implicit ~5 QPS / 10 burst and the controller spent a long time client-throttled.

Set QPS=100, Burst=200 when both are still unset and RateLimiter is nil

Not dynamic. Values apply when clients are built at process start. change env --> restart controller.

Testing e2e:
service-scale with high service_count (e.g. 300). compare time / retries on Wait for all Connectors/Listeners Ready vs baseline.

internal/kube/client/client.go
```diff
@@ -20,6 +23,43 @@ import (
...
+	}
+	if cfg.QPS <= 0 && cfg.Burst <= 0 {
+		cfg.QPS = 100
+		cfg.Burst = 200
+		return
+	}
+	if cfg.QPS > 0 && cfg.Burst <= 0 {
+		b := int(cfg.QPS * 2)
+		if b < 100 {
+			b = 100
```
fgiorgetti: Not sure if 100 is a good number if, for example, QPS==1, maybe just use what you have on line 49?

internal/kube/client/client.go
```diff
@@ -20,6 +23,43 @@ import (
...
+		}
+		cfg.Burst = b
+	}
+	if cfg.Burst > 0 && cfg.QPS <= 0 {
+		cfg.QPS = float32(cfg.Burst) / 4
+		if cfg.QPS < 25 {
```
fgiorgetti: Same thing here. If Burst is set to 4, is it appropriate to set this to 25?

fgiorgetti (commented):
Overall results are really good with the suggested defaults.
Total e2e test time dropped from 16m to 6m locally.

internal/kube/client/client.go
```diff
@@ -20,6 +23,43 @@ import (
...
+			cfg.Burst = n
+		}
+	}
+	if cfg.QPS <= 0 && cfg.Burst <= 0 {
+		cfg.QPS = 100
+		cfg.Burst = 200
+		return
+	}
+	if cfg.QPS > 0 && cfg.Burst <= 0 {
+		b := int(cfg.QPS * 2)
```
nluaces, line 49: what is the reasoning behind this operation and the other one in line 56?

AryanP123: For when only QPS is set, client-go still needs a burst for the token bucket, so we set it to twice the qps. And then when only burst is set, the same idea but in reverse.

nluaces: Got it, thanks! But why 2x specifically rather than some other multiplier?

AryanP123: @nluaces No strong reason, mainly consistency. The defaults when neither is set are 100 / 200, and client-go’s are 5 / 10, both being 2:1. Chose to reuse that same ratio when only one value is set so we don’t invent a second policy. Happy to use another multiplier if we change the defaults to match. Let me know what you think.

nluaces (approved):

c-kruse (approved):

fgiorgetti (approved):

PR #2418, opened 2026-04-08
Clean up api/types for v2 controller naming and router RBAC

AryanP123:
align controller constants with v2, centralize router Role rules in types.TransportPolicyRule, drop unused v1 service-interface JSON conversion and its test.


AryanP123:
Not modified yet:

- `ControllerClusterRoleName` and `ControllerExtendedClusterRoleName` now match v2’s single cluster role, keep both or deprecate one?
- `type ServiceInterfaceList []ServiceInterface` appears unused here
- `RouterSpec` / `DeploymentSpec` /  `AssemblySpec` still needed?

api/types/types.go
```diff
@@ -74,17 +73,53 @@ var TransportPrometheusAnnotations = map[string]string{
...
-	ControllerDeploymentName             string = "skupper-service-controller"
-	ControllerComponentName              string = "service-controller"
-	ControllerContainerName              string = "service-controller"
-	ControllerServiceAccountName         string = "skupper-service-controller"
-	ControllerRoleBindingName            string = "skupper-service-controller"
-	ControllerClusterRoleBindingNsFormat string = "skupper-service-controller-%s"
-	ControllerRoleName                   string = "skupper-service-controller"
-	ControllerClusterRoleName            string = "skupper-service-controller-basic"
-	ControllerExtendedClusterRoleName    string = "skupper-service-controller-extended"
+	ControllerDeploymentName             string = "skupper-controller"
```
fgiorgetti: Are any of these constants being used at all?

api/types/types.go
fgiorgetti: I believe that the `Headless` is also not used anywhere.

api/types/types.go
```diff
@@ -713,17 +687,6 @@ func (s *Headless) HasMemoryLimit() bool {
...
-	Size          int                `json:"size"`
-	TargetPort    int                `json:"targetPort,omitempty"`
-	Affinity      map[string]string  `json:"affinity,omitempty"`
-	AntiAffinity  map[string]string  `json:"antiAffinity,omitempty"`
-	NodeSelector  map[string]string  `json:"nodeSelector,omitempty"`
-	CpuRequest    *resource.Quantity `json:"cpuRequest,omitempty"`
-	MemoryRequest *resource.Quantity `json:"memoryRequest,omitempty"`
-}
-
 type ByServiceInterfaceAddress []ServiceInterface
```
fgiorgetti: This one either.

internal/kube/site/site.go
```diff
@@ -391,51 +391,16 @@ func (s *Site) checkRoleBinding(ctxt context.Context) error {
...
-		},
-	}
 	desired := &rbacv1.Role{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "rbac.authorization.k8s.io/v1",
 			Kind:       "Role",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:            "skupper-router",
+			Name:            types.TransportRoleName,
```
fgiorgetti: Actually I believe that the idea is not to put stuff into the api/types/types.go anymore.

api/types/types.go
fgiorgetti: This can be removed as well.
Most of this file, I'd say.

AryanP123: it’s the type of ServiceInterface.Headless (api/types/types.go 531), checked in RequiresIngressPortAllocations (573–574), and read in GetRouterConfigForHeadlessProxy (internal/qdr/qdr.go 1310, 1337+)

Correct me if I'm wrong

AryanP123: Used on the public client surface in api/types/client.go and used from internal/qdr. If both should be cleaned up let me know

fgiorgetti: You're right. It is referred to by `GetRouterConfigForHeadlessProxy`, which is not being used as well.
So we could also get rid of `GetRouterConfigForHeadlessProxy`.
The `ServiceInterface` type, was a representation of a Skupper service in V1, no longer used in v2.

nluaces (approved):

PR #2420, opened 2026-04-16
Allow users to provide certificates through secrets

### LINKED ISSUE #2128: Allow user to provide their own certificates to system sites, using just secrets
The existing mechanism that allows users to provide their own custom certificates when
dealing with system sites, is through placing the actual certificates into specific directories
before starting the respective system namespace.

The existing mechanism can be easily improved to allow users to provide theiro own certificates
by simply placing secrets under `input/resources`.

This is a nice enhancement, especially, when combined with the new `system apply` CLI command,
as users could easily provide their own secrets, without the need to deal with directories and files.

And this is not just a benefit for the CLI users, but to people using the `skupper.v2` Ansible collection,
as it also allows Secrets to be provided through the `skupper.v2.resource` module.

fgiorgetti:
... for system sites

Fixes #2128

### DIFF
--- a/internal/nonkube/common/fs_config_renderer.go
+++ b/internal/nonkube/common/fs_config_renderer.go
@@ -379,9 +379,15 @@ func (c *FileSystemConfigurationRenderer) createTlsCertificates(siteState *api.S
 		if certificate.Spec.Signing == false {
 			continue
 		}
-		secret, err := certs.GenerateSecret(name, certificate.Spec.Subject, nil, 0, nil)
-		if err != nil {
-			return err
+		var secret *corev1.Secret
+		var ok bool
+		if secret, ok = siteState.Secrets[name]; !ok {
+			secret, err = certs.GenerateSecret(name, certificate.Spec.Subject, nil, 0, nil)
+			if err != nil {
+				return err
+			}
+		} else {
+			fmt.Printf("-> User provided CA found (secret): %s\n", name)
 		}
 
 		ignoreExisting := true
@@ -390,7 +396,7 @@ func (c *FileSystemConfigurationRenderer) createTlsCertificates(siteState *api.S
 			// override with user provided CA
 			ignoreExisting = false
 			secret = userCaSecret
-			fmt.Printf("-> User provided CA found: %s\n", name)
+			fmt.Printf("-> User provided CA found (file): %s\n", name)
 		}
 		caPath := path.Join(outputPath, string(api.IssuersPath), name)
 		err = writeSecretFilesIgnore(caPath, secret, ignoreExisting)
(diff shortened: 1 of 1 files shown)


internal/nonkube/common/fs_config_renderer.go
```diff
@@ -379,9 +379,15 @@ func (c *FileSystemConfigurationRenderer) createTlsCertificates(siteState *api.S
...
+		if secret, ok = siteState.Secrets[name]; !ok {
+			secret, err = certs.GenerateSecret(name, certificate.Spec.Subject, nil, 0, nil)
+			if err != nil {
+				return err
+			}
+		} else {
+			fmt.Printf("-> User provided CA found (secret): %s\n", name)
 		}
 
 		ignoreExisting := true
```
pwright, line 393: should this be false, for updating secret?

internal/nonkube/common/fs_config_renderer.go
```diff
@@ -411,19 +418,27 @@ func (c *FileSystemConfigurationRenderer) createTlsCertificates(siteState *api.S
...
-			secret, err = certs.GenerateSecret(name, certificate.Spec.Subject, certificate.Spec.Hosts, 0, caSecret)
-			if err != nil {
-				return err
+			if secret, ok = siteState.Secrets[name]; !ok {
+				secret, err = certs.GenerateSecret(name, certificate.Spec.Subject, certificate.Spec.Hosts, 0, caSecret)
+				if err != nil {
+					return err
+				}
+			} else {
+				fmt.Printf("-> User provided %s certificate found (secret): %s\n", purpose, name)
```
pwright, line 427: should this be logged, something like:
logger.Debug("user provided certificate found (secret)",  
    slog.String("purpose", purpose),  
    slog.String("name", name),  
)

pwright (commented):
I'm still puzzling over this, but want to ask about the user story... 
Supposing I'm already populating `input/issuers/` and want to switch to using this method? is that path supported? or will this become the sole method supported?

internal/nonkube/common/fs_config_renderer.go
nluaces: If I'm not mistaken, if a user provides a CA secret the variable gets updated to false: https://github.com/fgiorgetti/skupper/blob/100b5367b99ade7b14249386e0241cf4b29f7c77/internal/nonkube/common/fs_config_renderer.go#L397

nluaces (approved):
Tested with a the reload type set to auto in the controller. It works after running `skupper system reload`, and it should be that way because with new certificates and issuers the router must be restarted. 
On the other hand, this change needs to be documented (how to create the structure of certs and how to make them effective) somewhere.

internal/nonkube/common/fs_config_renderer.go
fgiorgetti: @pwright both methods are supported, to keep it compatible.
We are just introducing a new way for users to provide their certs.

internal/nonkube/common/fs_config_renderer.go
fgiorgetti: This is a user facing message logged as part of the `system start` command, so we want it to be visible by default (no need to log timing info I believe).

PR #2424, opened 2026-04-28
Extend site state validator to ensure that multi-key listener strateg…

gabordozsa:
…y is immutable (#2411)

Fixes #2411

### DIFF
--- a/internal/nonkube/bundle/site_state_renderer.go
+++ b/internal/nonkube/bundle/site_state_renderer.go
@@ -34,7 +34,8 @@ type SiteStateRenderer struct {
 func (s *SiteStateRenderer) Render(loadedSiteState *api.SiteState, reload bool) error {
 	var err error
 	var validator api.SiteStateValidator = &common.SiteStateValidator{}
-	err = validator.Validate(loadedSiteState)
+	currentSiteState, _ := common.LoadCurrentSiteState(loadedSiteState.Site.Namespace)
+	err = validator.Validate(loadedSiteState, currentSiteState)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


nluaces:
By using the reload type set to `auto`:
`skupper system install --reload-type=auto`
I can modify the strategy from the input resources, and it will be modified in the runtime resources as well.
I think the issue is that this validation is not done in the `Refresh` method that updates the site changes without restarting the router:

https://github.com/gabordozsa/skupper/blob/a12a254e5017d5f99ada2a9befca160c1ce54712/internal/nonkube/compat/site_state_renderer.go#L382

gabordozsa:
> By using the reload type set to `auto`: `skupper system install --reload-type=auto` I can modify the strategy from the input resources, and it will be modified in the runtime resources as well. I think the issue is that this validation is not done in the `Refresh` method that updates the site changes without restarting the router:
> 
> 

@nluaces  I added the validation to the` Refresh() `method, too. 

It works for me now: `runtime` will not pick up the strategy change from the `input` dir.  However, I did not see any error message in the system-controller log. Is that okay?

nluaces:
@gabordozsa I have tried your last changes and I think it works as expected:

``` 
2026/04/30 15:22:28 INFO Resource has been created: /output/namespaces/west/input/resources/multikey.yaml component=input.resource.handler namespace=west
2026/04/30 15:22:28 ERROR Failed to refresh site state: invalid multikeylistener: my-backend-multikey - strategy cannot be changed component=input.resource.handler namespace=west
```
I've got this from the system-controller log.

nluaces (approved):

fgiorgetti (approved):

PR #2429, opened 2026-05-01
add system sites e2e tests

nluaces:
e2e tests that implement the Hello World example in system sites (with manual reload).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,7 +351,7 @@ jobs:
           name: Running E2E Tests
           command: |
             set -x
-            export SKUPPER_SYSTEM_CONTROLLER_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
+            export SKUPPER_SYSTEM_CONTROLLER_IMAGE="quay.io/skupper/system-controller:<< parameters.image_tag >>"
             export SKUPPER_CLI_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
             cd tests
             cp ~/kc.west ~/.kube/config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,6 +351,7 @@ jobs:
           name: Running E2E Tests
           command: |
             set -x
+            export SKUPPER_SYSTEM_CONTROLLER_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
             export SKUPPER_CLI_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
             cd tests
             cp ~/kc.west ~/.kube/config
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,6 +132,12 @@ workflows:
       - build-oci-images:
           <<: *run_for_numeric_tags
           image_tag: << pipeline.git.tag >>
+      - e2e:
+          <<: *run_for_numeric_tags
+          image_tag: << pipeline.git.tag >>
+          requires:
+            - build-oci-images
+            - build-all
       - publish-oci-images:
           <<: *run_for_numeric_tags
           <<: *v2_release_requires
@@ -195,6 +201,7 @@ workflows:
           <<: *run_for_all_branches
       - e2e:
           <<: *run_for_all_branches
+          image_tag: v2-dev
           requires:
             - build-oci-images
             - build-all
@@ -268,6 +275,10 @@ jobs:
           destination: test-artifacts
   e2e:
     executor: local_cluster_test_executor
+    parameters:
+      image_tag:
+        type: string
+        default: "v2-dev"
     steps:
       - checkout
       - kubectl-install
@@ -335,11 +346,12 @@ jobs:
             # the connector in east
             kubectl wait listener/backend \
               --timeout=60s \
-              '--for=jsonpath={.status.hasMatchingConnector}=true' 
+              '--for=jsonpath={.status.hasMatchingConnector}=true'
       - run:
           name: Running E2E Tests
           command: |
             set -x
+            export SKUPPER_CLI_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
             cd tests
             cp ~/kc.west ~/.kube/config
             make ci-tests
--- a/tests/e2e/scenarios/system-sites/cleanup-tasks.yml
+++ b/tests/e2e/scenarios/system-sites/cleanup-tasks.yml
@@ -1,0 +1,34 @@
+---
+- name: "[ {{ test_identifier }} ] - Cleanup frontend container"
+  ansible.builtin.command:
+    cmd: "{{ platform }} rm -f frontend"
+  failed_when: false
+  when:
+    - not skip_teardown | default(false) | bool
+    - "'west' in inventory_hostname"
+
+- name: "[ {{ test_identifier }} ] - Cleanup backend container"
+  ansible.builtin.command:
+    cmd: "{{ platform }} rm -f backend"
+  failed_when: false
+  when:
+    - not skip_teardown | default(false) | bool
+    - "'east' in inventory_hostname"
+
+- name: "[ {{ test_identifier }} ] - Stop namespaces"
+  skupper.v2.system:
+    action: stop
+    namespace: "{{ namespace_name }}"
+    platform: podman
+  failed_when: false
+  when: not skip_teardown | default(false) | bool
+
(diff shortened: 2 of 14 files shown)


tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+    - name: "[ {{ test_identifier }} ] - Pre-test cleanup"
+      ansible.builtin.include_tasks: cleanup-tasks.yml
+
+    - name: "[ {{ test_identifier }} ] - Main test block"
+      block:
+        - name: "[ {{ test_identifier }} ] - Install system controller"
+          skupper.v2.controller:
+            action: install
+            platform: "{{ platform }}"
+          failed_when: false
```
fgiorgetti: I believe that `failed_when` should be removed here.
If the system-controller is already installed, the module will return OK, not changed.

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+          register: system_install
+
+        # West Site - Setup and get token
+        - name: "[ {{ test_identifier }} ] - West Site Setup"
+          block:
+            - name: "[ {{ test_identifier }} ] - Deploy frontend service"
+              ansible.builtin.shell:
+                cmd: "{{ platform }} run --name frontend -d --rm -p 127.0.0.1:7070:8080 quay.io/skupper/hello-world-frontend --backend http://host.containers.internal:8080"
+              changed_when: false
+              when: "'west' in inventory_hostname"
```
fgiorgetti: Is the `when` clause needed in these 2 tasks? Considering it is already defined for the `block`?

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+                cmd: "{{ platform }} run --name backend -d --rm -p 127.0.0.1:9090:8080 quay.io/skupper/hello-world-backend"
+              changed_when: false
+              register: backend_deploy
+
+            - name: "[ {{ test_identifier }} ] - Apply east site resources"
+              skupper.v2.resource:
+                path: "{{ playbook_dir }}/resources/east/"
+                platform: "{{ platform }}"
+                namespace: "{{ namespace_name }}"
+              when: "'east' in inventory_hostname"
```
fgiorgetti: Same comment for the `when` clause here, maybe not needed.

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+            - name: "[ {{ test_identifier }} ] - Apply west site resources"
+              skupper.v2.resource:
+                path: "{{ playbook_dir }}/resources/west/"
+                platform: "{{ platform }}"
+                namespace: "{{ namespace_name }}"
+              when: "'west' in inventory_hostname"
+
+            - name: "[ {{ test_identifier }} ] - Start west namespace"
+              skupper.v2.system:
+                action: start
```
fgiorgetti, line 45: The `image` option should be passed here (CLI image), using the one built by the CI.

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+  tasks:
+    # Pre-test cleanup
+    - name: "[ {{ test_identifier }} ] - Pre-test cleanup"
+      ansible.builtin.include_tasks: cleanup-tasks.yml
+
+    - name: "[ {{ test_identifier }} ] - Main test block"
+      block:
+        - name: "[ {{ test_identifier }} ] - Install system controller"
+          skupper.v2.controller:
+            action: install
```
fgiorgetti, line 22: The `image` option should be passed here, using the one built through the CI.

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -0,0 +1,98 @@
...
+
+            - name: "[ {{ test_identifier }} ] - Apply token to east site"
+              skupper.v2.resource:
+                def: "{{ west.links['0.0.0.0'] }}"
+                namespace: "{{ namespace_name }}"
+                platform: "{{ platform }}"
+  
+            - name: "[ {{ test_identifier }} ] - Start east namespace"
+              skupper.v2.system:
+                action: start
```
fgiorgetti, line 77: Same thing here. Pass the `image` option to the system module.

tests/e2e/scenarios/system-sites/cleanup-tasks.yml
```diff
@@ -0,0 +1,34 @@
...
+  ansible.builtin.command:
+    cmd: "{{ platform }} rm -f backend"
+  failed_when: false
+  when:
+    - not skip_teardown | default(false) | bool
+    - "'east' in inventory_hostname"
+
+- name: "[ {{ test_identifier }} ] - Stop namespaces"
+  skupper.v2.system:
+    action: stop
```
fgiorgetti, line 20: Pass the `image` option here (using the CLI image built through the CI).

fgiorgetti (commented):
It is working fine. I have added a couple of comments related to the `image` option for some of the module invocations, but I believe we should raise a separate issue to handle that, as it seems to affect all e2e tests.

tests/e2e/scenarios/system-sites/cleanup-tasks.yml
```diff
@@ -20,13 +20,15 @@
...
     platform: "{{ platform }}"
+    image: "{{ skupper_cli_image }}"
   failed_when: false
   when: not skip_teardown | default(false) | bool
 
 - name: "[ {{ test_identifier }} ] - Uninstall system controller"
   skupper.v2.controller:
     action: uninstall
     platform: "{{ platform }}"
+    image: "{{ skupper_cli_image }}"
```
fgiorgetti: Here we need the system controller image instead.

tests/e2e/scenarios/system-sites/test.yml
```diff
@@ -21,7 +21,7 @@
           skupper.v2.controller:
             action: install
             platform: "{{ platform }}"
-          failed_when: false
+            image: "{{ skupper_cli_image }}"
```
fgiorgetti: System controller image here as well

.circleci/config.yml
```diff
@@ -335,11 +346,12 @@ jobs:
             # the connector in east
             kubectl wait listener/backend \
               --timeout=60s \
-              '--for=jsonpath={.status.hasMatchingConnector}=true' 
+              '--for=jsonpath={.status.hasMatchingConnector}=true'
       - run:
           name: Running E2E Tests
           command: |
             set -x
+            export SKUPPER_CLI_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
```
fgiorgetti, line 355: We also need to pass the SKUPPER_SYSTEM_CONTROLLER_IMAGE variable.

.circleci/config.yml
```diff
@@ -335,11 +346,13 @@ jobs:
             # the connector in east
             kubectl wait listener/backend \
               --timeout=60s \
-              '--for=jsonpath={.status.hasMatchingConnector}=true' 
+              '--for=jsonpath={.status.hasMatchingConnector}=true'
       - run:
           name: Running E2E Tests
           command: |
             set -x
+            export SKUPPER_SYSTEM_CONTROLLER_IMAGE="quay.io/skupper/cli:<< parameters.image_tag >>"
```
fgiorgetti: Env var is correct, but the image is wrong.

nluaces: sorry!

c-kruse (commented):
Hi! Sorry for the late entry to the review.

I think this is going to pull the latest v2-dev images (and break on a new release tag.)

One thing we had to take some special care to make work with the kube e2e tests was using the local build images instead of pulling them from quay. As I understand it, there isn't a way to compel skupper system sites to not `--pull always` like there is with the kube control plane.

Open to other ideas, but I think that leaves us with the choice of developing a system site image pull policy mechanism (larger than just this PR) or adding a local registry  to our CI jobs so that we can push the images for the e2e tests to pull.

nluaces:
@c-kruse agree with your input. I suggest to do it in a separate pull request though. We could merge this one after resolving the issue that you have mentioned.

nluaces:
need to include: e2e tests with auto reload.

nluaces:
this PR should include #2491

Issue #2380, opened 2026-02-12
Add namespace attribute in amqp  agent logs and reduce duplicated code


nluaces:
**Is your feature request related to a problem? Please describe.**
The amqp Agent originally prints its log messages to its own kube-adaptor container.

Now that can run as part of the system-controller container, the messages being logged are not qualified for the namespace, so it is hard to tell for which namespace the operations are being executed against. 

**Describe the solution you'd like**
Add the namespace attribute in slog.

**Additional context**
It is also necessary to do some refactoring by using sync_router_ops for both kube-adaptor and system-adaptor, in order to not duplicate code: https://github.com/nluaces/skupper/blob/d4bc626eb701ba734864ebecf9ab3e61be0b58ee/internal/qdr/sync_router_ops.go

PR #2435, opened 2026-05-06
Reduce duplicated code and improve system adaptor logs

nluaces:
Solves #2380

### DIFF
--- a/internal/kube/adaptor/config_sync.go
+++ b/internal/kube/adaptor/config_sync.go
@@ -100,7 +100,7 @@ func (c *ConfigSync) configEvent(key string, configmap *corev1.ConfigMap) error 
 	if err := c.syncSslProfileCredentialsToDisk(desired.SslProfiles); err != nil {
 		return err
 	}
-	if err := c.syncSslProfilesToRouter(desired.SslProfiles); err != nil {
+	if err := qdr.SyncSslProfilesToRouter(c.agentPool, desired.SslProfiles); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internal/kube/adaptor/config_sync.go
```diff
@@ -99,143 +99,20 @@ func (c *ConfigSync) configEvent(key string, configmap *corev1.ConfigMap) error
...
-	var synced bool
-
-	synced, err = syncBridgeConfig(agent, desired)
-
-	c.agentPool.Put(agent)
-	if err != nil {
-		return fmt.Errorf("Error while syncing bridge config : %s", err)
-	}
-	if !synced {
-		return fmt.Errorf("Bridge config is not synchronised yet")
```
fgiorgetti, line 154: I don't think it is a problem, but just one difference in behavior is that on the qdr.SyncBridgeConfig, it is just logging an error when synced is false, while here it is returning an error.
Considering that it is called periodically, I believe we should be fine.

fgiorgetti (approved):

nluaces:
hi @ajssmith, 
in my last commit I have included `syncProxyProfilesToRouter` with the rest of the sync operations used by kube-adaptor and system-adaptor. That way is ready to use for #2439 in the system-adaptor.

internal/kube/adaptor/config_sync.go
nluaces: I changed it, because the log was showing something like `ERROR: Bridge config is not synchronised yet` often; and I think that something like that could puzzle the user. Also, if this message is frequent it should not be considered an error, right?

ajssmith (approved):
lgtm, verified proxy config, thanks

PR #2446, opened 2026-05-14
add http proxy vanflow record to Link records

### LINKED ISSUE #2361: Add Observability support for HTTP Proxy tunnel
Such as update to vanflow record to indicate associated proxy profile used by connection to traverse HTTP Proxy.

JPadovano1483:
added a new attribute on the LinkRecord to pick up the new HTTP_PROXY record from the router that indicates if a proxy is being used on the link

Fixes #2361

### DIFF
--- a/cmd/network-observer/spec/openapi.yaml
+++ b/cmd/network-observer/spec/openapi.yaml
@@ -1126,6 +1126,10 @@ components:
               type: string
               nullable: true
               description: When connected, the name of the destitation (peer) site.
+            httpProxy:
+              type: string
+              nullable: true
+              description: When connected, whether this link uses an HTTP proxy.
     RouterAccessRecord:
       allOf:
         - $ref: '#/components/schemas/baseRecord'
--- a/cmd/network-observer/spec/openapi.yaml
+++ b/cmd/network-observer/spec/openapi.yaml
@@ -785,6 +785,12 @@ components:
         - inter-router
         - edge
         - unknown
+    proxyProtocolType:
+      type: string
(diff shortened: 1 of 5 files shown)


fgiorgetti (approved):
LGTM @c-kruse, but I'd like to get your confirmation as well.

cmd/network-observer/spec/openapi.yaml
```diff
@@ -1126,6 +1126,10 @@ components:
               type: string
               nullable: true
               description: When connected, the name of the destitation (peer) site.
+            httpProxy:
+              type: string
+              nullable: true
+              description: When connected, whether this link uses an HTTP proxy.
```
c-kruse: I am iffy on the raw pass-through string here, would prefer a boolean "httpProxy" to "yes/no" or a "proxyProtocol" enum with "NONE,HTTP" were we to ever entertain another option like SOCKS5 or something (I don't remember whether or not we've decided that makes any sense.)

Also the description could use some work: e.g. "true when the link connection is established through an HTTP proxy"

c-kruse (changes requested):
Plumbing all looks good, just a couple notes on what we want the API spec to be.

cmd/network-observer/spec/openapi.yaml
JPadovano1483: Thanks. I'll switch it to a boolean for now. I can bring it up in our meeting on Tuesday to see if there was a decision or thoughts about using other protocols and if it makes sense to switch to an enum.

c-kruse (approved):
Looks great! Thanks!

PR #2449, opened 2026-05-14
Skip router ConfigMap entries when TLS credential secrets are missing

### LINKED ISSUE #2433: Kube-adaptor is not syncing resources when a secret is missing
**Describe the bug**
If you have a resource that depends on an externally provided tls credential (`generateTlsCredentials: false`), like a `RouterAccess`, and the respective secret is not available, kube-adaptor is not proceeding with the syncronization logic for the other entities, like bridges and links.

**How To Reproduce**
1. Create a RouterAccess resource with `.spec.generateTlsCredentials: false`.
2. Create any other resource like, a listener: `skupper listener create backend 8080`.
3. List the active tcpListeners, using: `kubectl exec deploy/skupper-router -- skmanage query --type tcpListener`

You will see an empty list of tcpListeners as they have been defined to the ConfigMap, but not through the Management API.

**Expected behavior**
If a resource, like `RouterAccess`, `Listener`, `Connector` refers to a TLS Credential that does not exist, they should simply be skipped from the router configuration at all (including the configmap itself), so that the configuration of other entities are not impacted.

In case the router pod restarts, the `config-init` container will prevent it from starting, with the following error:

```
Waiting for Secrets to be created for SslProfiles
```

**Environment details**
- Skupper 2.1.3

AryanP123:
Fixes #2433

### DIFF
--- a/internal/kube/secrets/manager.go
+++ b/internal/kube/secrets/manager.go
@@ -241,6 +241,21 @@ func (w *ProfilesWatcher) keyfunc(name string) string {
 	return w.namespace + "/" + name
 }
 
+// TlsCredentialSecretPresent reports whether a TLS secret for the credential is in the
+// ProfilesWatcher informer cache (same source kube-adaptor uses for ssl profiles).
+func (w *ProfilesWatcher) TlsCredentialSecretPresent(credentialName string) bool {
+	if credentialName == "" || w.Cache == nil {
+		return false
+	}
+	for _, secretName := range profileSecrets(credentialName) {
+		secret, err := w.Cache.Get(w.keyfunc(secretName))
+		if err == nil && secret != nil && secret.Type == corev1.SecretTypeTLS {
+			return true
+		}
+	}
+	return false
+}
+
 func (w *ProfilesWatcher) checkPriorValidity(secret *corev1.Secret) uint64 {
 	result := w.pvProvider.TLSPriorValidRevisions()
 	if secret.ObjectMeta.Annotations == nil {
--- a/internal/kube/site/extended_bindings.go
+++ b/internal/kube/site/extended_bindings.go
@@ -406,6 +406,9 @@ func (b *ExtendedBindings) Apply(config *qdr.RouterConfig) bool {
 		}
 	}
 	for _, ptl := range b.perTargetListeners {
+		if ptl.definition.Spec.TlsCredentials != "" && !b.bindings.TlsCredentialIncluded(ptl.definition.Spec.TlsCredentials) {
+			continue
+		}
 		if ptl.updateBridgeConfig(b.bindings.SiteId, &desired) {
 			updated = true
 		}
@@ -425,6 +428,9 @@ func (b *ExtendedBindings) Apply(config *qdr.RouterConfig) bool {
 func (b *ExtendedBindings) AddSslProfiles(config *qdr.RouterConfig, definitions map[string]*skupperv2alpha1.AttachedConnector) bool {
 	profiles := map[string]qdr.SslProfile{}
 	for _, c := range definitions {
+		if c.Spec.TlsCredentials != "" && !b.bindings.TlsCredentialIncluded(c.Spec.TlsCredentials) {
+			continue
+		}
 		if c.Spec.TlsCredentials != "" {
 			if !c.Spec.UseClientCert {
 				//if only ca is used, need to qualify the profile to ensure that it does not collide with
@@ -451,6 +457,7 @@ func (b *ExtendedBindings) AddSslProfiles(config *qdr.RouterConfig, definitions 
 
 func (b *ExtendedBindings) SetSite(site *Site) {
 	b.bindings.SetSiteId(site.site.GetSiteId())
+	b.bindings.SetTlsSecretAllowed(site.tlsCredentialSecretPresent)
 	b.site = site
 }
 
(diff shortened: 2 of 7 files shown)


internal/kube/site/extended_bindings.go
```diff
@@ -425,6 +428,9 @@ func (b *ExtendedBindings) Apply(config *qdr.RouterConfig) bool {
 func (b *ExtendedBindings) AddSslProfiles(config *qdr.RouterConfig, definitions map[string]*skupperv2alpha1.AttachedConnector) bool {
 	profiles := map[string]qdr.SslProfile{}
 	for _, c := range definitions {
+		if c.Spec.TlsCredentials != "" && !b.bindings.TlsCredentialIncluded(c.Spec.TlsCredentials) {
```
fgiorgetti: 1. Minor (optional), but we could move the 2nd condition here under the if statement below.
2. Maybe log a message saying  attached connector is being skipped?

internal/site/routeraccess.go
```diff
@@ -60,9 +60,25 @@ func (m RouterAccessMap) findInterRouterRole() (*skupperv2alpha1.RouterAccessRol
...
+	source := m
+	if tlsSecretAllowed != nil {
+		source = make(RouterAccessMap, len(m))
+		for k, ra := range m {
+			if ra.Spec.TlsCredentials != "" && !tlsSecretAllowed(ra.Spec.TlsCredentials) {
+				continue
+			}
+			source[k] = ra
+		}
+	}
```
nluaces: if this function is skipping RouterAccess entries with tlsCredentials configured but not included in the profile watcher cache; I would suggest some naming changes: 
- `tlsSecretAllowed` --> `isCredentialInCache` or `credentialExists`
- `DesiredConfigAllowingTlsSecrets` --> `DesiredConfigWithAvailableCredentials`

internal/site/bindings.go
```diff
@@ -29,6 +29,7 @@ type Bindings struct {
 	listeners         map[string]*skupperv2alpha1.Listener
 	multiKeyListeners map[string]*skupperv2alpha1.MultiKeyListener
 	handler           BindingEventHandler
+	tlsSecretAllowed  func(secretName string) bool
```
nluaces: maybe it could be renamed to `isTlsSecretAvailable` or `isTlsSecretPresent` ?

internal/site/bindings.go
```diff
@@ -53,6 +54,22 @@ func (b *Bindings) SetSiteId(siteId string) {
 	b.SiteId = siteId
 }
 
+func (b *Bindings) SetTlsSecretAllowed(f func(string) bool) {
```
nluaces: maybe it could be renamed  to `SetIsTlsSecretPresent` or to `SetSecretChecker` for better readability, what do you think?

fgiorgetti (commented):
It worked well for me.

I have been able to create links and accesses resources without providing a secret and that did not prevent other resources from being reconciled and configured.

Restarting the skupper-router deployment worked fine without blocking the config-init container.

One comment, though, was the Listeners and Connectors that refer to tlsCredentials that do not exist, are processed without warning or errors on the CR status. It would be nice to get a similar behavior to what we have with Links and RouterAccesses.

internal/kube/secrets/manager.go
```diff
@@ -241,6 +241,21 @@ func (w *ProfilesWatcher) keyfunc(name string) string {
...
 
+// TlsCredentialSecretPresent reports whether a TLS secret for the credential is in the
+// ProfilesWatcher informer cache (same source kube-adaptor uses for ssl profiles).
+func (w *ProfilesWatcher) TlsCredentialSecretPresent(credentialName string) bool {
+	if credentialName == "" || w.Cache == nil {
+		return false
+	}
+	for _, secretName := range profileSecrets(credentialName) {
+		secret, err := w.Cache.Get(w.keyfunc(secretName))
+		if err == nil && secret != nil && secret.Type == corev1.SecretTypeTLS {
```
c-kruse, line 252: Have we committed anywhere else that the Secret must have type `kubernetes.io/tls`? IIRC I've used opaque before just out of kubectl convenience and it works out fine.

AryanP123: Was matching existing code. ProfilesWatcher and kube-adaptor secrets.Sync already only handle kubernetes.io/tls secrets, and doc/tls/README.md documents that layout. It states Skupper expects TLS credentials in standard kubernetes.io/tls format with tls.crt, tls.key, and optionally ca.crt. I kept the same check here so we don’t treat a secret as “ready” if the adaptor won’t sync it. Let me know if you prefer to change it.

fgiorgetti:
It is working for me. Now we get appropriate status on resources that are not ready, pending TLS secret availability.
@c-kruse @nluaces I'd like to know thoughts as well before approving it.

nluaces (approved):

c-kruse (approved):
I think we've got a few more things to work through before this gets released, but I'm happy to get it merged now and running in development before we work things out.

Quickly:
* It appears that Listener and Connector (and AttachedConnector I think) may never get configured after a missing tls secret is added without a spec change.
* I'm suspicious about the logging here, I suspect it will log (and re-log) a lot of redundant information that users should be getting from the resource status instead.
* We're missing MultiKeyListener logic - could be this work predates that resource.
* I'm vaguely concerned about the assumptions we make (outside of this changeset) on Site readiness in relation to RouterAccess readiness and configuration, and think we should play with it for a bit longer before resolving to change anything.

fgiorgetti (commented):
@AryanP123 Could you rebase your work on top of main? As @c-kruse pointed, this is missing MKL, Proxy and a few other changes already on main.

fgiorgetti (approved):
Looks cleaner and seems to work great now!
@c-kruse @nluaces if you wanna give it another try.
I tried a local scenario with listeners, connectors, links, routeraccess, attached connectors and mkls.
A missing secret did not seem to block the configuration of other resources now.

PR #2451, opened 2026-05-18
add common labels and annotations for network-observer resources

### LINKED ISSUE #2438: Support additional network observer configuration options
For the  Skupper Network Observer Helm chart:

Custom labels and annotations for all resources (Routes, etc).  These custom labels and annotations should include the labels and annotations in the `spec.template.metadata` part of the Deployment.

(Research) The hostname to use for the Skupper console (in the Route).

Specific CPU and memory requests and limit for each container in the skupper-network-observer deployment (with defaults).

JPadovano1483:
- adds a common set of labels and annotations to be applied to all resources generated by the network-observer helm chart
- adds configurability to the route hostname
- adds resource configurability per container

fixes #2438

### DIFF
--- a/charts/network-observer/values.yaml
+++ b/charts/network-observer/values.yaml
@@ -184,9 +184,11 @@ route:
   labels: {}
   annotations: {}
 
+# Deprecated: use containerResources instead.
+resources: {}
 
 # Per-container resources for the network-observer Deployment.
-resources:
+containerResources:
   networkObserver: {}
   #   requests:
   #     cpu: 100m
--- a/charts/network-observer/values.yaml
+++ b/charts/network-observer/values.yaml
@@ -141,6 +141,10 @@ auth:
       create: true
       nameOverride: ""
 
+# This is for setting Kubernetes Annotations to all resources.
+commonAnnotations: {}
+# This is for setting Kubernetes Labels to all resources.
+commonLabels: {}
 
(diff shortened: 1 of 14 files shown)


charts/network-observer/values.yaml
```diff
@@ -175,14 +179,35 @@ ingress:
 # observer
 route:
   enabled: false
+  host: ""
+  subdomain: ""
+  labels: {}
+  annotations: {}
+
 
-resources: {}
```
c-kruse, line 179: I think we should go a different way with changing `resources`. I think we should keep resources as-is for backwards compatibility and general helm/kube conventions (folks make assumptions about the schema of a "resources" field.)

Instead I think we should go with something like this:
```
# deprecated, use containerResources instead
resources: {}

# per container resource configuration
containerResources:
  networkObserver: {}
  proxy: {}
  prometheus: {}
```

And then the template could be something along these lines - where it uses the containerResources with a fallback of resources.

```
{{- $resources := .Values.resources }}
{{- with .Values.containerResources.networkObserver }}
{{- $resources = . }}
{{- end }}
{{- with $resources }}
resources:
  {{- toYaml . | nindent 12 }}
{{- end }}

```

c-kruse (changes requested):
Annotations and labels looks good to me! One suggestion with the resources.

c-kruse (approved):
Thank you!

PR #2456, opened 2026-05-20
Add new process that handles event duplication when a watched file is created or updated

### LINKED ISSUE #2432: Auto-reload: updated resources don't get processed
**Describe the bug**
When updating an existing system site resource, OnUpdate event is triggered but nothing happens

**How To Reproduce**
- Create a resource and wait for it to be processed (see if it is reflected in the runtime directory).
- Watch the logs to see that the following events are triggered: `OnCreate`, `OnUpdate`
- Update the resource 
- Watch the logs to see that the `OnUpdate` event is triggered but nothing else happens.

**Expected behavior**
- The resource gets updated accordingly.

**Environment details**
 - Skupper CLI: main
 - Platform: podman/docker

**Additional information**
if the resource gets updated using an editor, it works as expected, because the OnCreate event is triggered.

nluaces:
Fixes #2432 

### Why was this approach chosen?
From `fsnotify` package[1]: 

>A single "write action" initiated by the user may show up as one or multiple                      writes, depending on when the system syncs things to disk. For example when compiling a large Go program you may get hundreds of Write events, and you may want to wait until you've stopped receiving them (see the dedup example in cmd/fsnotify).[2]

Following the dedup example, this change adds a new process called `EventDeduplicator`, that handles deduplication of _OnCreate_ and _OnUpdate_ events; collecting them in a time window and processing only the final event.

```mermaid
flowchart TD
   id1([new listener resource is created])-->|triggers|InputResourceHandler-->Q{site exists?}
   Q -->|Yes| A[Refresh Router Config]
   Q -->|No| B[Bootstrap]
   InputResourceHandler-->|sends event|EventDeduplicator
   EventDeduplicator-->|sends final event|InputResourceHandler
```

This new process gets stopped during the `teardown` (a site is deleted) or when the `NamespaceController` gets stopped as well.

[1]https://pkg.go.dev/github.com/fsnotify/fsnotify#section-readme
[2]https://github.com/fsnotify/fsnotify/blob/v1.10.1/cmd/fsnotify/dedup.go#L17

### DIFF
--- a/internal/nonkube/controller/event_deduplicator_test.go
+++ b/internal/nonkube/controller/event_deduplicator_test.go
@@ -1,0 +1,114 @@
+package controller
+
+import (
+	"log/slog"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestEventDeduplicator(t *testing.T) {
+	t.Run("deduplicates multiple events for same file", func(t *testing.T) {
+		var processedCount int
+		var mu sync.Mutex
+		var processedFiles []string
+
+		handler := func(filename string) {
+			mu.Lock()
+			defer mu.Unlock()
+			processedCount++
+			processedFiles = append(processedFiles, filename)
+		}
+
+		logger := slog.Default()
+		stopCh := make(chan struct{})
+		deduplicator := NewEventDeduplicator(stopCh, handler, logger)
+		defer deduplicator.Stop()
+
+		deduplicator.QueueEvent("test.yaml")
+		deduplicator.QueueEvent("test.yaml")
+		deduplicator.QueueEvent("test.yaml")
+
+		time.Sleep(200 * time.Millisecond)
+
+		mu.Lock()
+		defer mu.Unlock()
+
+		if processedCount != 1 {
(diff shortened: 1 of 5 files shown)


internal/nonkube/controller/event_deduplicator_test.go
```diff
@@ -0,0 +1,114 @@
...
+		logger := slog.Default()
+		stopCh := make(chan struct{})
+		deduplicator := NewEventDeduplicator(stopCh, handler, logger)
+		defer deduplicator.Stop()
+
+		deduplicator.QueueEvent("test.yaml")
+		deduplicator.QueueEvent("test.yaml")
+		deduplicator.QueueEvent("test.yaml")
+
+		time.Sleep(200 * time.Millisecond)
```
c-kruse, line 33: Suggestion: Now that we're on go 1.25 we should look at running this in a testing/synctest bubble. We already have a lot of tests that sleep for timing based things like this and they tend to be one of the more common sources of flaky tests.

fgiorgetti (approved):
Great improvemente, Noe!

PR #2463, opened 2026-05-26
Revert "Fixes HA router deployment selector overlap (#2363)"

### LINKED ISSUE #2440: Controller is unable to upgrade the router-deployment in 2.2.0
**Describe the bug**
The skupper-router deployment created for each running site older than 2.2.0,
has the `.spec.selector` field set as:

```yaml
  selector:
    matchLabels:
      skupper.io/component: router
```

In 2.2.0, the controller tries to sync it with the desired state, and it attempts to update the `spec.selector` to:

```yaml
  selector:
    matchLabels:
      skupper.io/component: router
      skupper.io/group: {{ .Group }}
```

Which causes a failure, as the `.spec.selector` field is immutable. Leaving the skupper-router deployment outdated.

**How To Reproduce**

- Install an older verions of the skupper-controller (older than 2.2.0)
- Create a site
- Upgrade the skupper-controller to 2.2.0 and observe its logs

You should observe a message like:

```
time=2026-05-12T14:57:34.382Z level=ERROR msg="Error configuring site" component=kube.site.site namespace=tross-colo-bone name=co-located id=90ca3fd2-bcc5-4623-a4ef-f4ad4270f403 error="Deployment.apps \"skupper-router\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"skupper.io/component\":\"router\", \"skupper.io/group\":\"skupper-router\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable"
```

This seems to be caused by the following change:
https://github.com/skupperproject/skupper/pull/2363/changes

**Expected behavior**

We need to decide whether to rollback this change and provide a better way to handle such change,
or workaround it somehow, possibly forcing a delete/create of the skupper-router deployment?

**Environment details**
- Skupper Controller: Upgrade from any 2.x to latest 2.2.0
- Platform: Kubernetes / OpenShift

**Additional context**

https://github.com/skupperproject/skupper/issues/2299
https://github.com/skupperproject/skupper/pull/2363/changes

c-kruse:
This reverts commit fcb7f45fb8d18d1c6c8e8e5b4671d19a07938f6f.

Resolves https://github.com/skupperproject/skupper/issues/2440

### DIFF
--- a/internal/kube/controller/controller_test.go
+++ b/internal/kube/controller/controller_test.go
@@ -1293,7 +1293,7 @@ func (f *factory) routerDeployment(name string, namespace string) *unstructured.
 	content := map[string]interface{}{
 		"spec": map[string]interface{}{
 			"selector": map[string]interface{}{
-				"matchLabels": f.routerSelectorWithGroup(name),
+				"matchLabels": f.routerSelector(false),
 			},
 			"template": map[string]interface{}{
 				"spec": map[string]interface{}{
@@ -1418,7 +1418,7 @@ func (r *ExpectedResources) deployment(name string, namespace string) *unstructu
 	content := map[string]interface{}{
 		"spec": map[string]interface{}{
 			"selector": map[string]interface{}{
-				"matchLabels": f.routerSelectorWithGroup(name),
+				"matchLabels": f.routerSelector(false),
 			},
 			"template": map[string]interface{}{
 				"spec": map[string]interface{}{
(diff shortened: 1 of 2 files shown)


nluaces (approved):

c-kruse:
Second thoughts on this one: May be that the best way is through. An upgrade from 2.2.0 to 2.2.1 migh be broken same as 2.1 to 2.2.0 :facepalm:

fgiorgetti:
> Second thoughts on this one: May be that the best way is through. An upgrade from 2.2.0 to 2.2.1 migh be broken same as 2.1 to 2.2.0 🤦

True stuff. If someone created a new site using 2.2.0, reverting the selector won't help.
What about detecting selectors during sites recovery (controller init) and recreating the skupper-router deployment if selectors diff?

PR #2465, opened 2026-05-27
Fix controller panic when MultiKeyListener has no Site

### LINKED ISSUE #2441: Controller crashes if a MultiKeyListener is created without a Site
**Describe the bug**
If a MultiKeyListener resource is created on a namespace that does not have a Site defined,
the controller is crashing and not recovering.

**How To Reproduce**
Create a MultiKeyListener resource on a namespace that does not have a Site.

**Expected behavior**
Controller ignores the MultiKeyListener resource as there is no Site on the namespace.

**Environment details**
Skupper 2.2.0

AryanP123:
Fixes #2441


fgiorgetti (approved):
LGTM! Thank you.

PR #2474, opened 2026-06-03
Retry site flow controller start after acquiring leader lease

### LINKED ISSUE #2466: kube-adaptor does not retry site flow controller startup after acquiring leader lease
When `kube-adaptor` becomes leader (`OnStartedLeading`), it starts status sync and then calls `startFlowController`. If `startFlowController` fails (transient Kubernetes API / etcd error while fetching the transport Deployment), the error is logged once and not retried while the pod remains leader.

There is no backoff/retry loop. leadership is not lost, so `OnStartedLeading` does not run again.

Observed in a Summit/OpenShift report (transient etcd timeout during leader lease renewal): after re-acquiring `skupper-site-leader`, logs showed `COLLECTOR: Failed to start controller for emitting site events: failed to get transport deployment: ...` with no subsequent restart of the controller.

Suggested fix
Retry `startFlowController` (with backoff) while still holding leadership and until `leaderCtx` is cancelled, instead of logging once and giving up.

AryanP123:
Fixes #2466

### DIFF
--- a/internal/kube/adaptor/collector.go
+++ b/internal/kube/adaptor/collector.go
@@ -83,7 +83,7 @@ func siteCollector(ctx context.Context, cli *internalclient.KubeClient) {
 }
 
 func startFlowController(ctx context.Context, cli *internalclient.KubeClient) error {
-	deployment, err := cli.Kube.AppsV1().Deployments(cli.Namespace).Get(context.TODO(), deploymentName(), metav1.GetOptions{})
+	deployment, err := cli.Kube.AppsV1().Deployments(cli.Namespace).Get(ctx, deploymentName(), metav1.GetOptions{})
 	if err != nil {
 		return fmt.Errorf("failed to get transport deployment: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


c-kruse (approved):
Looks good to me! Don't have a great way to validate how it behaves it in an error mode: i suspect this was identified in the first place during some transient error.

fgiorgetti (approved):
I couldn't reproduce the original problem, but the retry mechanism seems to be working as expected.

internal/kube/adaptor/collector.go
```diff
@@ -119,6 +119,43 @@ func startFlowController(ctx context.Context, cli *internalclient.KubeClient) er
 	return nil
 }
 
+func retryStartFlowController(ctx context.Context, cli *internalclient.KubeClient) {
```
nluaces: Not a blocker, but if this function can start the flow controller without retrying, how about calling it `ensureStartFlowController`?

nluaces (approved):

PR #2484, opened 2026-06-09
Reject token issuance for edge sites and update tests (#2482)

### LINKED ISSUE #2482: Reject access token issuance from edge sites
Edge sites cannot accept incoming links from remote sites (per the Site CRD), but `skupper token issue` does not check for edge mode today. If an edge site has link access enabled, token issuance can succeed and produce a token that should not work.

There is an unimplemented TODO in `internal/kube/grants/tokens.go` (`setValidHostsFromSite`) to reject this case. 

Also add a unit test for this.

evanwang9x:
Fixes (#2482)

### DIFF
--- a/internal/cmd/skupper/token/kube/token_issue.go
+++ b/internal/cmd/skupper/token/kube/token_issue.go
@@ -101,6 +101,12 @@ func (cmd *CmdTokenIssue) ValidateInput(args []string) error {
 		if !ok {
 			validationErrors = append(validationErrors, fmt.Errorf("there is no active skupper site in this namespace"))
 		} else {
+			for _, site := range siteList.Items {
+				if site.IsReady() && site.Spec.Edge {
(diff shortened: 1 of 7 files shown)


AryanP123:
Nice work. The tokens.go fix looks good. The issue also mentions that skupper token issue can still succeed on edge sites, can we address that path too so issuance is rejected end-to-end, not only at redeem time? Add an edge-site check in token_issue.go ValidateInput so the CLI fails early, not only at redeem time

fgiorgetti:
@evanwang9x I agree with @AryanP123. It works, but it would be nice to improve on the handling of the AccessGrant by the controller as well, to prevent it from changing to a Ready state.

AryanP123:
LGTM. @fgiorgetti wdyt?

fgiorgetti (approved):
Looks good to me now! Thank you @evanwang9x !

PR #2494, opened 2026-06-10
Collect pod resource usage and cgroup metrics in must-gather

evanwang9x:
Must gather script can now collect memory and CPU usage via container exec and also captures cgroup limits, requests, and throttling stats.
Outputs the info to "must-gather/namespaces/"namespace"/pods/"pod-name"/"container-name"/"container-name"/resources/". Creates files to track what each does

1. memory.info: Tracks current memory usage and shows the memory limit.
2. cpu.stat: Provides CPU stats on total usage time, user/syxstem breakdown, and nr_throttled which shows performance bottlenecks. 
Screenshot 2026-06-10 at 2 41 54 PM

3. cpu.max: maximum amount of CPU processing power allotted.
4. last.terminated: Provides an error message on why the previous container crashed if applicable. Helpful for identifying errors like Out of memory.
5. memory.pressure / cpu.pressure: Tracks how much time the pod spends waiting for CPU or memory to become available. This helps identify if the pod is slowing down because the node is too busy.

To run the script:

1. podman build -t skupper-gather-test -f Dockerfile.must-gather .
2. mkdir ./must-gather
3. podman run --rm --net=host -v $HOME/.kube/config:/root/.kube/config:Z -v $(pwd)/must-gather:/must-gather:Z -e KUBECONFIG=/root/.kube/config skupper-gather-test

## Summary by CodeRabbit

* **New Features**
  * Added detailed per-pod/per-container resource collection for Skupper workloads, including memory and CPU usage.
  * Includes memory/CPU pressure (PSI) indicators and reports “not available” when a signal can’t be retrieved.
  * Records each container’s prior termination state to aid troubleshooting.
  * Expanded data collection across both Skupper Site and AttachedConnector namespaces, plus the Skupper controller’s namespace.

### DIFF
--- a/scripts/skupper-gather.sh
+++ b/scripts/skupper-gather.sh
@@ -34,8 +34,14 @@ fi
 echo "===memory.info==="
 curr=$(cat "$d/memory.current" 2>/dev/null)
 max=$(cat "$d/memory.max" 2>/dev/null)
-curr_kb=$((${curr:-0} / 1024))
-curr_mb=$((${curr:-0} / 1024 / 1024))
+
+if [ -z "$curr" ]; then
+  usage_str="unknown (memory.current missing - cgroup v1?)"
+else
+  curr_kb=$((curr / 1024))
+  curr_mb=$((curr / 1024 / 1024))
+  usage_str="$curr bytes ($curr_kb KiB / $curr_mb MiB)"
+fi
 
 if [ -z "$max" ]; then
   limit_str="unknown (memory.max missing - cgroup v1?)"
@@ -47,7 +53,7 @@ else
   limit_str="$max bytes ($max_kb KiB / $max_mb MiB)"
 fi
 
-echo "Usage: $curr bytes ($curr_kb KiB / $curr_mb MiB) - Limit: $limit_str"
+echo "Usage: $usage_str - Limit: $limit_str"
 
 echo "===memory.pressure==="
 if [ -f "$d/memory.pressure" ]; then
@@ -79,8 +85,10 @@ while read -r key val; do
 done < "$d/cpu.stat"
 
 echo "===cpu.max==="
-read -r quota period < "$d/cpu.max"
-if [ "$quota" = "max" ]; then
+read -r quota period < "$d/cpu.max" 2>/dev/null
+if [ -z "$quota" ] || [ -z "$period" ]; then
+  echo "CPU Limit: unknown (cpu.max missing - cgroup v1?)"
+elif [ "$quota" = "max" ]; then
   echo "CPU Limit: uncapped (no quota set)"
 else
   cores=$(awk "BEGIN {printf \"%.2f\", $quota/$period}")
--- a/scripts/skupper-gather.sh
+++ b/scripts/skupper-gather.sh
@@ -32,12 +32,14 @@ else
 fi
 
 echo "===memory.info==="
-curr=$(cat "$d/memory.current")
-max=$(cat "$d/memory.max")
-curr_kb=$((curr / 1024))
-curr_mb=$((curr / 1024 / 1024))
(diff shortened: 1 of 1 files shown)


scripts/skupper-gather.sh
```diff
@@ -28,6 +124,7 @@ function getSkupperRouterSkstatInNamespace() {
     for flag in "${flags[@]}"; do
       oc exec -n "${sitens}" "${routerName}" -c router -- skstat -"${flag}" > "${logPath}/skstat.${flag}" 2>&1
     done
+    getPodResources "${sitens}" "${routerName}" "router"
```
ajssmith: Suggest that resources be retrieved for all skupper containers not just the router 

e.g. getSkupperResourcesInNamespace iterates over label app.kubernetes.io/part-of: skupper

that then calls to getPodResources and iterates across all containers in the pod...

fgiorgetti: I agree with Andy, this is the only observation I have as well.
But the data being retrieved from the router container looks good to me.
Just add all the other possible containers into the list and we should be good to merge it.

scripts/skupper-gather.sh
```diff
@@ -16,6 +16,102 @@ get_log_collection_args() {
...
+      limit_str="max (uncapped)"
+    else
+      max_kb=$((max / 1024))
+      max_mb=$((max / 1024 / 1024))
+      limit_str="$max bytes ($max_kb KiB / $max_mb MiB)"
+    fi
+
+    echo "Usage: $curr bytes ($curr_kb KiB / $curr_mb MiB) - Limit: $limit_str" > /tmp/memory.info
+  ' 2>/dev/null
+  oc exec -n "${ns}" "${podName}" -c "${container}" -- cat /tmp/memory.info > "${resourcePath}/memory.info" 2>&1
```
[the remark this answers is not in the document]
fgiorgetti: This suggestion also makes sense.

scripts/skupper-gather.sh
```diff
@@ -16,6 +16,102 @@ get_log_collection_args() {
...
+    key=\$(echo \$line | awk '{print \$1}')
+    desc=\${cpu_descriptions[\$key]}
+    if [[ \$key == *_usec ]] && [ \$val -ge 1000000 ]; then
+      sec=\$(awk \"BEGIN {printf \\\"%.2f\\\", \$val/1000000}\")
+      echo \"\$line (\$sec s) | \$desc\"
+    else
+      echo \"\$line | \$desc\"
+    fi
+  done
+" > "${resourcePath}/cpu.stat" 2>&1
```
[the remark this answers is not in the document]
fgiorgetti: @evanwang9x maybe you could try to use `oc debug` using an image that provides bash.

fgiorgetti (commented):
One last comment, if not too much to ask.
Maybe we could also try to collect resource utilization for the skupper-controller pod itself.
Other than that, it seems to work fine for all sites I have on my CRC cluster.

scripts/skupper-gather.sh
```diff
@@ -95,6 +239,8 @@ function main() {
   controllerNamespace=$(oc get pods --all-namespaces -l app.kubernetes.io/name=skupper-controller -o jsonpath="{.items[0].metadata.namespace}")
   # this gets also logs for all pods in that namespace
   inspect "ns/$controllerNamespace"
+  # collect resource utilization for the skupper-controller pod itself
+  getSkupperResourcesInNamespace "${controllerNamespace}" "app.kubernetes.io/name=skupper-controller"
```
fgiorgetti: The controller can be installed to multiple namespaces (considering namespace scoped installations), so better to consider that eventually we might find multiple namespaces, instead of a single.

ajssmith (approved):
lgtm, thanks

PR #2496, opened 2026-06-12
Refactored controller_test.go to use helper functions

### LINKED ISSUE #2495: Have a common resource "factory" for tests, etc.
There are many instances of unit (mock) and potentially integration tests that define skupper resources. Currently the controller test has a private resource "factory". Suggestion is to re-factor and create an internal utils factory that can be used across unit tests and future integration tests, etc.

evanwang9x:
Plan is to later use these helper functions in the other test suites

Closes #2495

### DIFF
--- a/internal/kube/controller/controller_test.go
+++ b/internal/kube/controller/controller_test.go
@@ -30,6 +30,7 @@ import (
 	fakedynamic "k8s.io/client-go/dynamic/fake"
 	k8stesting "k8s.io/client-go/testing"
 
+	"github.com/skupperproject/skupper/internal/fixtures"
 	internalclient "github.com/skupperproject/skupper/internal/kube/client"
 	fakeclient "github.com/skupperproject/skupper/internal/kube/client/fake"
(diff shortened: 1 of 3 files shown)


internal/kube/controller/controller_test.go
```diff
@@ -1044,182 +1044,82 @@ func (*factory) addUID(site *skupperv2alpha1.Site, uid string) *skupperv2alpha1.
...
-			Name:      name,
-			Namespace: namespace,
-		},
-		Spec: skupperv2alpha1.SiteSpec{
-			HA:         ha,
-			LinkAccess: linkAccess,
-			Edge:       edge,
-		},
-	}
+	s := utils.NewSite(name, namespace)
```
c-kruse: Let's put these in their own `fixture` package instead of dumping them in utils. `fixture.Site()` reads cleaner than `utils.NewSite()` IMO. That, and it is poor style to mix the code for test helpers in with code compiled into our released executables.

c-kruse (approved):

nluaces (approved):

PR #2503, opened 2026-06-19
Fixes Controller Port Mapping Recovery

### LINKED ISSUE #2406: Bug: recovered() in port_mapping.go always discards recovered TCP listener ports, causing port drift on controller restart
## Summary

`recovered()` in `internal/qdr/port_mapping.go` contains a self-defeating logic bug that causes all TCP listener port recovery to silently fail on every controller restart. This leads to the controller and router re-allocating ports from scratch and potentially diverging, resulting in stale `EndpointSlice` ports and `Connection refused` (HTTP 503) errors for services routed through Skupper.

## Affected Version

- Skupper v2.1.3 (current latest), bug present in `main`

## Root Cause

`recovered()` correctly writes the recovered port into the map, then immediately enters an `if` block that is **always true** and undoes everything:

```go
func (p *PortMapping) recovered(key string, portstr string) {
    p.pool.InUse(port)
    p.mappings[key] = port              // ✅ writes key → port

    if existing, ok := p.mappings[key]; ok {  // ⚠️ ALWAYS true — just wrote it
        p.pool.Release(existing)              // ❌ releases port back to free pool
        delete(p.mappings, key)              // ❌ deletes the mapping we just created
    }
    // net result: empty map, port back in free pool
}
```

The `if` block is copy-pasted from `ReleasePortForKey` and should not be here. `RecoverPortMapping()` always returns an empty mapping — all previously allocated ports are treated as free.

## Failure Chain

1. Controller restarts, calls `RecoverPortMapping(config)` to recover prior port assignments.
2. Bug causes recovery to be a no-op — `mappings` map is empty.
3. Controller and router both re-allocate ports from scratch independently.
4. Due to ordering differences, they can arrive at **different port numbers**.
5. Controller updates the `EndpointSlice` with its chosen port. Router listens on a different port. Traffic hits `Connection refused`.

## Why a Second Restart Fixes It

The first restart updates the router ConfigMap to the newly chosen ports. On the second restart both controller and router read the same updated ConfigMap and converge — despite the bug still being present.

## Fix

```go
func (p *PortMapping) recovered(key string, portstr string) {
    port, err := strconv.Atoi(portstr)
    if err != nil {
        p.logger.Error("Failed to convert port to int", slog.String("port", portstr), slog.Any("error", err))
        return
    }
    p.pool.InUse(port)
    p.mappings[key] = port
    // removed the if/Release/delete block — it undoes the recovery
}
```

## Historical Context

Regression of v1 issue [#671](https://github.com/skupperproject/skupper/issues/671) / [PR #673](https://github.com/skupperproject/skupper/pull/673) ("Ensure we collect all recovered ports", v0.8.6, Feb 2022). The v2 rewrite reintroduced a different form of the same failure.

## Workaround

Run `kubectl rollout restart deployment/skupper-controller` **twice**. The first restart stabilises the ConfigMap; the second converges controller and router.

c-kruse:
Fixes flaw in port mapping recovery that prevented recovery from functioning. Also handles recent changes to tcpListener naming conventions for config entity name prefixing and multikeylisteners.

Closes https://github.com/skupperproject/skupper/issues/2406

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for port mapping recovery to validate that port mappings are correctly preserved across multiple configured listener endpoints.

* **Refactor**
  * Enhanced port mapping key derivation logic during recovery to ensure more reliable and deterministic listener endpoint associations.

### DIFF
--- a/internal/kube/site/bindings_test.go
+++ b/internal/kube/site/bindings_test.go
@@ -1,6 +1,8 @@
 package site
 
 import (
+	"log/slog"
+	"strconv"
 	"testing"
 
 	"github.com/skupperproject/skupper/internal/qdr"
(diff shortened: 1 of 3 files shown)


nluaces (approved):
Tested in local: 
- created listeners and checked their ports
- deleted one of them and restarted controller
- recreated the missing listener
- checked that the remaining listeners have the same ports

fgiorgetti (approved):

PR #2509, opened 2026-06-24
Fixing the lexical error Hello-world test

ramdrvcs:
While I was running the E2E tests, I had the follwing error for hello-world module. 

```
fatal: [west]: FAILED! => {"attempts": 30, "changed": true, "msg": "Task failed: Action failed: Unknown error.", "rc": 1, "return_code": 1, 
"stderr": "cannot parse the data: `lexical error: invalid character inside string.\n
{\"user\":{\"uid\":0,\"gid\":0,\"addi\n                     (right here) ------^\n`\n",
"stderr_lines": ["cannot parse the data: `lexical error: invalid character inside string.", " 
{\"user\":{\"uid\":0,\"gid\":0,\"addi", "                     (right here) ------^", "`"], 
"stdout": "", "stdout_lines": []}

```

## Description

Fixes a critical runtime failure in the `run_curl` role where the task fails with an `Unknown error` (`lexical error: invalid character inside string`). 

The failure occurs because ansible task definition contain end of line character in sh location. As of now it has been noticed in s390x. 

### Key Changes

* **Single Network Call (`curl -s -w "\n%{http_code}" "{{ run_curl_address }}"`)**: Reduces traffic by making exactly one request to the backend. The API response text is captured alongside the 3-digit HTTP status code, which is appended to a brand-new line at the very bottom.
* **Local Code Extraction (`tail -n 1`)**: Pipes the combined response string and isolates only the final line to extract the HTTP status code.
* **Local Body Preservation (`sed "$ d"`)**: Trims the appended status code line from the stream, safely preserving the original API message body (including any native trailing text or newlines).
* **Defensive String Evaluation (`[ "$code" = "200" ]`)**: Switches the conditional check to a robust string comparison. This protects the shell executor from crashing if network drops cause the status code to return empty.
* **Clean YAML Syntax**: Formatted using the folded block scalar (`>`) to preserve intended script indentation and readability without breaking the Kubernetes command processor array.

## Testing Triaged

- [x] Verified that payloads containing nested JSON quotes no longer trigger a `lexical error`.
- [x] Confirmed that the loop correctly runs exactly one HTTP request per retry iteration.
- [x] Confirmed the task successfully completed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end curl output handling by capturing both the response body and HTTP status code from a single curl execution.
  * Updated the role’s output formatting to reliably extract the status code and present results consistently as `BODY: <body>` on success and on failure.

### DIFF
--- a/tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
+++ b/tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
@@ -32,17 +32,18 @@
     namespace: "{{ curl_namespace }}"
     kubeconfig: "{{ kubeconfig }}"
     pod: curl
-    command: >-
+    command: >
       sh -c '
-        response=$(curl -s {{ run_curl_address }});
-        code=$(curl -s -w "%{http_code}" -o /dev/null {{ run_curl_address }});
-        if [ $code -eq 200 ]; then
-          echo "BODY: $response";
-          exit 0;
-        else
-          echo "BODY: $response";
-          exit 1;
-        fi
+      response=$(curl -s -w "\n%{http_code}"  "{{ run_curl_address }}");
+      code=$(echo "$response" | tail -n 1);
+      body=$(echo "$response" | sed "$ d");
+      if [ "$code" = "200" ]; then
+        echo "BODY: $body";
+        exit 0;
+      else
+        echo "BODY: $body";
+        exit 1;
+      fi
       '
   register: curl_result
   until: curl_result is success


tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -34,21 +34,22 @@
...
-      '
+      tmp_file=$(mktemp);
+      code=$(curl -s -w "%{http_code}" -o "$tmp_file" "{{ run_curl_address }}");
+      response=$(base64 < "$tmp_file" | tr -d "\n");
+      rm -f "$tmp_file";
+      if [ "$code" -eq 200 ]; then
+        echo "BODY_BASE64: $response";
+        exit 0;
+      else
+        echo "BODY_BASE64: $response";
```
hash-d: Picky: only the `exit 0` / `exit 1` needs to be within the `if`/`else` clauses.  Move the `echo` to before the `if`, as it is identical in both cases.

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -34,21 +34,22 @@
...
-        if [ $code -eq 200 ]; then
-          echo "BODY: $response";
-          exit 0;
-        else
-          echo "BODY: $response";
-          exit 1;
-        fi
-      '
+      tmp_file=$(mktemp);
+      code=$(curl -s -w "%{http_code}" -o "$tmp_file" "{{ run_curl_address }}");
```
hash-d: I like the single call to curl.  That's a good improvement.

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -34,21 +34,22 @@
...
+      response=$(base64 < "$tmp_file" | tr -d "\n");
```
hash-d: This is the crux of the PR, but I'm not convinced it is really required.  The PR description mentions nested double quotes, but I see no nesting in the provided example.  

Also, the contents of the example don't seem to match what I'd expect as a response from the backend service (it contanis user information in a json, whereas I'd expect a plain message such as `Hello, stranger.  I am Supreme Cipher (1360c152dd3c).`).

Which task is throwing that error?  Is it really failing on a curl call?

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -34,21 +34,22 @@
...
+      fi'
   register: curl_result
   until: curl_result is success
   retries: "{{ run_curl_retries }}"
   delay: "{{ run_curl_delay }}"
 
 - name: Debug the curl command output
   ansible.builtin.debug:
-    msg: "{{ curl_result.stdout }}"
+    msg: "{{ (curl_result.stdout | regex_search('BODY_BASE64: (.*)', '\\1') | first) | b64decode }}"
```
hash-d: This is the only place where you're decoding it from base64.  Is `run_curl` used elsewhere?  Could other tests depend / parse the actual output.  If so, they'd need to do the decoding, too.

hash-d (commented):
@ramdrvcs, I see the move to single curl call as a clear improvement.  However, I'm not so sure about the base64 encoding.   Please check my comments and let me know what you're seeing.

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
ramdrvcs: I have not changed the if/else block much here. I am leaving the original way as it is.

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
ramdrvcs: Thanks

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
ramdrvcs: It is corrected without base64

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
ramdrvcs: It is corrected without base64

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -32,17 +32,18 @@
     namespace: "{{ curl_namespace }}"
     kubeconfig: "{{ kubeconfig }}"
     pod: curl
-    command: >-
+    command: >
```
hash-d, line 35: Apparently, this is the main thing that fixes this issue, though we're not sure what the mechanism for that is

tests/e2e/collections/ansible_collections/e2e/tests/roles/run_curl/tasks/main.yml
```diff
@@ -32,17 +32,18 @@
...
-        code=$(curl -s -w "%{http_code}" -o /dev/null {{ run_curl_address }});
-        if [ $code -eq 200 ]; then
-          echo "BODY: $response";
-          exit 0;
-        else
-          echo "BODY: $response";
-          exit 1;
-        fi
+      response=$(curl -s -w "\n%{http_code}"  "{{ run_curl_address }}");
+      code=$(echo "$response" | tail -n 1);
```
hash-d, line 38: Having a single call to `curl` is definitely an improvement, and not using a tmpfile for that is good, too.

hash-d (approved):
lgtm; I worked with Ram and the full suite still runs fine after the changes.

PR #2515, opened 2026-07-01
feat: add readiness probe and read only root filesystem

CarolinaRFerreira:
## Summary by CodeRabbit

* **Bug Fixes / Improvements**
  * Hardened runtime security by enabling a read-only root filesystem and reducing privileges for the chart’s main containers.
  * Added a dedicated writable temporary directory (and an NGINX cache volume) to maintain expected runtime behavior.
  * Made health checks more flexible: liveness/readiness/startup probes are now conditionally rendered for both the NGINX/proxy and monitoring containers when enabled via chart settings.

### DIFF
--- a/charts/network-observer/values.yaml
+++ b/charts/network-observer/values.yaml
@@ -229,13 +229,13 @@ securityContext:
 
 skipManagementLabels: false
 
-## Controller containers' readiness probe. Evaluated as a template.
+## Controller containers' readiness probe
 ## ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes
 readinessProbe:
-  enabled: true
+  enabled: false
   # httpGet:
   #   path: /healthz
-  #   port: 9000
+  #   port: 9191
   #   scheme: HTTP
   # failureThreshold: 3
   # initialDelaySeconds: 60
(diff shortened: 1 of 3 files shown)


charts/network-observer/values.yaml
```diff
@@ -218,9 +221,38 @@ podSecurityContext:
...
   allowPrivilegeEscalation: false
   capabilities:
     drop:
       - ALL
 
 skipManagementLabels: false
+
+## Controller containers' readiness probe
+## ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes
+readinessProbe:
```
c-kruse, line 269: @CarolinaRFerreira thank you for the pull request. I am curious if you had a plan for these probes already or have been able to test this change with enabled: true. As it stands, I do not believe the commented example: `GET :9191/healthz` would not work.

CarolinaRFerreira: I've made some changes to the values. I already tested them and they seem to be correct, both the probes and the readOnlyFilesystem.

c-kruse (approved):
Thank you for the contribution! Looks good to me.

JPadovano1483 (approved):
LGTM

PR #2517, opened 2026-07-01
Report Connector Status when host and selector are both missing

### LINKED ISSUE #2512: Report Connector status when host and selector are both missing
A Connector needs spec.host or spec.selector. If both are empty, the controller only logs an error today and the Connector may still show Configured=True. Validate this in CheckConnector and set Configured=False with a clear message.

evanwang9x:
Closes #2512 

## Summary by CodeRabbit

* **Bug Fixes**
  * Connector “Configured” status now accounts for spec validation (missing host/selector) alongside TLS and routing issues.
  * Connectors lacking both `spec.host` and `spec.selector` are marked unconfigured with a clearer, explicit message.

* **Tests**
  * Expanded connector configuration status tests, including a new “missing host/selector” scenario.
  * Improved assertions to verify the exact user-facing condition message and configured state.

### DIFF
--- a/internal/kube/site/site.go
+++ b/internal/kube/site/site.go
@@ -1058,6 +1058,7 @@ func (s *Site) CheckConnector(name string, connector *skupperv2alpha1.Connector)
 		return s.updateConnectorConfiguredStatus(connector, stderrors.New("No active site in namespace"))
 	}
 	var tlsErr error
+	var specErr error
 	if connector != nil {
 		tlsErr = s.missingTlsCredentialsErr(connector.Spec.TlsCredentials)
 		if tlsErr != nil {
@@ -1066,6 +1067,9 @@ func (s *Site) CheckConnector(name string, connector *skupperv2alpha1.Connector)
 				slog.String("connector", connector.Name),
 				slog.String("secret", connector.Spec.TlsCredentials),
 			)
+		}
+		if connector.Spec.Host == "" && connector.Spec.Selector == "" {
+			specErr = stderrors.New("Connector must define either spec.host or spec.selector")
 		}
 	}
 	update := s.bindings.UpdateConnector(name, connector)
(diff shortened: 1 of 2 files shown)


internal/kube/site/site.go
```diff
@@ -1067,6 +1068,9 @@ func (s *Site) CheckConnector(name string, connector *skupperv2alpha1.Connector)
 				slog.String("secret", connector.Spec.TlsCredentials),
 			)
 		}
+		if connector.Spec.Host == "" && connector.Spec.Selector == "" {
+			specErr = stderrors.New("Connector must define either spec.host or spec.selector")
```
c-kruse: Suggest including "non-empty" as a qualifier here: Connector must define a non-empty spec.host or spec.selector. If anyone ever sees this message, they have already defined one of these fields.

As I'm sure you noticed when testing, the kubernetes API refuses an invalid Connector that does not satisfy a oneOf required selector or host.

```
$ cat << EOF | kubectl apply -f -
kind: Connector
metadata:
  name: invalid
spec:
  port: 8888
  routingKey: test

EOF

The Connector "invalid" is invalid:
* <nil>: Invalid value: "": "spec" must validate one and only one schema (oneOf). Found none valid
* spec.selector: Required value

$ cat << EOF | kubectl apply -f -
kind: Connector
metadata:
  name: invalid
spec:
  port: 8888
  routingKey: test
  host: ""
  selector: ""
EOF
The Connector "invalid" is invalid: <nil>: Invalid value: "": "spec" must validate one and only one schema (oneOf). Found 2 valid alternatives

```

To "trick" it you need to provide either host or selector with an empty string (really was a miss on our api spec side.) Because of how our go types were set up, we end up missing that distinction and can't easily plainly say "Invalid value for host|selector. Must not be empty".

c-kruse (commented):
Looks good! Just a nit pick on the wording.

internal/kube/site/site.go
evanwang9x: I see thanks for letting me know, I'll implement that change later today👍.

c-kruse (approved):

PR #2524, opened 2026-07-07
Accept Opaque secrets as TLS credentials for backward compatibility

### LINKED ISSUE #2518: Kube-to-System Site Links Broken: tlsCredentials Secret type Opaque not allowed
**Describe the bug**
* I created and started a podman Site with link access enabled
* I attempted to apply a Link from that Podman site to a kube site
* Link status is Error `TLS credentials secret "..." not found` because the Secret type is Opque

**How To Reproduce**
```
# set up system site site
skupper -p podman -n bug-tls-secret-type site create tls --enable-link-access
skupper -p podman -n bug-tls-secret-type system start
# create kube site
kubectl create ns bug-tls-secret-type
skupper site create -n bug-tls-secret-type test
# link sites
kubectl apply -n bug-tls-secret-type -f $HOME/.local/share/skupper/namespaces/bug-tls-secret-type/runtime/links/link-tls-0.0.0.0.yaml
# view links
kubectl -n bug-tls-secret-type get links
```

Results:
```
$ k -n bug-tls-secret-type get links
NAME       STATUS   REMOTE SITE   MESSAGE
link-tls   Error                  TLS credentials secret "link-tls" not found
$ k -n bug-tls-secret-type get secrets
NAME                   TYPE                DATA   AGE
link-tls               Opaque              4      5m5s
skupper-local-ca       kubernetes.io/tls   3      5m49s
skupper-local-server   kubernetes.io/tls   3      5m49s
skupper-site-ca        kubernetes.io/tls   3      5m49s
```

**Expected behavior**
I'd like a working Link please.

**Environment details**
 - Skupper CLI: [e.g. 1.5.0, 1.4.3]
 - Platform: [e.g kubernetes, openshift]

**Additional context**
At some point the requirement to use a Secret with type of kubernetes.io/tls snuck in. I'm not sure that was a well considered change.

AryanP123:
Fixes #2518

## Summary by CodeRabbit

- **New Features**
  - TLS credentials are now recognized from compatible secrets even when they aren’t explicitly labeled as TLS.
  - Syncing can now satisfy TLS profiles backed by opaque secrets (with TLS material present).

- **Bug Fixes**
  - Improved TLS secret classification to be more permissive for backward compatibility while excluding irrelevant types (e.g., basic-auth).
  - Updated cache and sync logic to consistently use the new TLS predicate.

- **Tests**
  - Added unit tests covering TLS/opaque/basic-auth/nil secret scenarios and an integration-style test for opaque TLS sync.

### DIFF
--- a/internal/kube/secrets/context.go
+++ b/internal/kube/secrets/context.go
@@ -18,22 +18,19 @@ const (
 	annotationKeySslProfileContext = "internal.skupper.io/tls-profile-context"
 )
 
-// IsTlsCredentialSecret reports whether a Secret carries TLS credential material.
-// kubernetes.io/tls secrets are always accepted; Opaque secrets with standard TLS
-// keys are also accepted for backward compatibility with pre-generated link tokens.
+// IsTlsCredentialSecret reports whether a secret should be treated as TLS credentials.
+// Presence checks stay permissive; kube-adaptor determines usable contents per sslProfile.
 func IsTlsCredentialSecret(secret *corev1.Secret) bool {
-	if secret == nil {
+	if secret == nil || secret.Data == nil || secret.Type == corev1.SecretTypeBasicAuth {
 		return false
 	}
 	if secret.Type == corev1.SecretTypeTLS {
 		return true
 	}
-	if secret.Type != corev1.SecretTypeOpaque && secret.Type != "" {
-		return false
+	if secret.Type == corev1.SecretTypeOpaque || secret.Type == "" {
+		return len(secret.Data["ca.crt"]) > 0
 	}
-	return len(secret.Data["tls.crt"]) > 0 &&
-		len(secret.Data["tls.key"]) > 0 &&
-		len(secret.Data["ca.crt"]) > 0
+	return false
 }
 
 type profileContext struct {
--- a/internal/kube/secrets/context.go
+++ b/internal/kube/secrets/context.go
@@ -18,6 +18,24 @@ const (
 	annotationKeySslProfileContext = "internal.skupper.io/tls-profile-context"
 )
 
+// IsTlsCredentialSecret reports whether a Secret carries TLS credential material.
+// kubernetes.io/tls secrets are always accepted; Opaque secrets with standard TLS
+// keys are also accepted for backward compatibility with pre-generated link tokens.
+func IsTlsCredentialSecret(secret *corev1.Secret) bool {
+	if secret == nil {
+		return false
+	}
+	if secret.Type == corev1.SecretTypeTLS {
+		return true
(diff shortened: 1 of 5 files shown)


internal/kube/secrets/context.go
```diff
@@ -18,6 +18,24 @@ const (
...
+	if secret == nil {
+		return false
+	}
+	if secret.Type == corev1.SecretTypeTLS {
+		return true
+	}
+	if secret.Type != corev1.SecretTypeOpaque && secret.Type != "" {
+		return false
+	}
+	return len(secret.Data["tls.crt"]) > 0 &&
```
c-kruse: This last assertion would break at least one (probably untested) thing. Until this regression and the subsequent changes to tls credentials dependencies the controller was designed to mostly turn a blind eye to tlsCredentials secret contents. It mostly tried to sort the contents out in the router/kube-adaptor.

It's probably more work than just this to iron out all of the corner cases, but to tell whether or not a secret is suitable to be used with an sslProfile configuration we need more context. Example: a Connector with spec.useClientCert=false (default) only configures a sslProfile with a CA - no cert or key file.


All that to say - let's either properly solve for this with a path with concrete "reasons" a secret is unacceptable for a particular tlsCredential or return to ye olden days with something wide open like 'secret.Type == "" || opaque || tls'

c-kruse: Not to complicate this more, but just ran into another edge case we've never handled well: Listeners and MultiKeyListeners with requireClientCert=false have no real purpose for a CA certificate. We currently configure the sslProfile with one unconditionally and the router crashes when it is empty/absent. Looks like the kubernetes.io/tls type only requires tls.crt and tls.key (not ca.crt.)

Maybe the least controversial, most backwards compatible `IsTlsCredentialSecret` would be something like this? Would really count on your judgement here: A cryptic "Secret not found" error message is a big improvement over a crashing router deployment, but a major downgrade from a working link/service generated by regular skupper tooling.
```
func IsTlsCredentialSecret(secret *corev1.Secret) bool {
	if secret == nil || secret.Data == nil || secret.Type == corev1.SecretTypeBasicAuth {
		return false
	}
	return len(secret.Data["ca.crt"]) > 0
}
```

AryanP123: Went with your approach. Opaque/empty type + `ca.crt` present, basic-auth excluded. I also kept `kubernetes.io/tls` unconditionally true so listener certs without `ca.crt` don't regress. Still treating this as presence-only, happy to follow up with per-use-case validation separately if useful.

c-kruse (approved):
Thanks @AryanP123!

PR #2530, opened 2026-07-09
Connection Sweeper CLI Command

evanwang9x:
Added a CLI command skupper debug sweep to scan through TCP connections and see how long they've been idle and if it exceeds a user set time kill it

## Summary by CodeRabbit

* **New Features**
  * Added a hidden `skupper debug sweep` command to identify idle/orphaned router TCP-adaptor connections and optionally delete them.
  * Works in both Kubernetes mode (executes per ready router pod with per-pod and overall totals) and non-Kubernetes mode (targets a local router management endpoint).
  * Introduces `--idle-threshold` (default 4 hours) and `--execute` (default off) to switch between reporting and deletion.
  * Correlates router connections with kernel TCP timing to generate “idle for …” reasons and killed/skipped/failed outcomes.


internal/cmd/skupper/debug/debug.go
```diff
@@ -18,6 +19,42 @@ func NewCmdDebug() *cobra.Command {
...
+		Use:   "sweep",
+		Short: "Detect and kill idle TCP adaptor connections",
+		Long: `Queries the router management API for TCP adaptor connections, identifies
+connections that have been idle beyond the threshold, and force-closes them
+via adminStatus=deleted.`,
+		Example: "skupper debug sweep --url amqp://127.0.0.1:5672 --idle-threshold 14400",
+	}
+
+	cmd := common.ConfigureCobraCommand(configuredPlatform, cmdDesc, kubeCommand, nonKubeCommand)
+	cmd.Hidden = true
```
nluaces, line 50: why is it hidden?

internal/cmd/skupper/debug/debug.go
```diff
@@ -18,6 +19,42 @@ func NewCmdDebug() *cobra.Command {
...
+
+	cmdFlags := common.CommandConnSweeperFlags{
+		URL:           sweeper.DefaultURL,
+		IdleThreshold: sweeper.DefaultIdleThreshold,
+		Skmanage:      sweeper.DefaultSkmanage,
+	}
```
nluaces: I think setting the default values here is redundant, it has been done already in lines 49-52

internal/cmd/skupper/debug/kube/conn_sweeper.go
```diff
@@ -0,0 +1,192 @@
...
+
+	execConfig := *restconfig
+	execConfig.APIPath = "/api"
+	execConfig.GroupVersion = &corev1.SchemeGroupVersion
+	execConfig.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
+	cmd.Rest = &execConfig
+}
+
+func (cmd *CmdConnSweeper) ValidateInput(args []string) error {
+	if cmd.Flags.IdleThreshold <= 0 {
```
nluaces: For validating numbers, we use in the CLI `validator.NewNumberValidator()`.

internal/cmd/skupper/debug/kube/conn_sweeper.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("--idle-threshold must be a positive number of seconds")
+	}
+	if cmd.clientErr != nil {
+		return cmd.clientErr
+	}
+	if cmd.KubeClient == nil || cmd.Rest == nil {
+		return fmt.Errorf("could not initialize kubernetes client")
+	}
+	return nil
```
nluaces: These validations are not part of the users input, in this method the value of the flags and arguments provided by the user is what should be validated:
```
	cmd.Flags().StringVar(&cmdFlags.URL, "url", sweeper.DefaultURL, "Router management URL")
	cmd.Flags().IntVar(&cmdFlags.IdleThreshold, "idle-threshold", sweeper.DefaultIdleThreshold, "Seconds with no data received before a connection is flagged as orphaned")
	cmd.Flags().BoolVar(&cmdFlags.DryRun, "dry-run", false, "List idle connections without killing them")
	cmd.Flags().StringVar(&cmdFlags.Skmanage, "skmanage", sweeper.DefaultSkmanage, "Path to the skmanage binary")
```

nluaces: Also, all validation errors are returned at once, with 

```
return errors.Join(validationErrors...)
```

You need to append all the errors in a slice array like `var validationErrors []error`

You can follow the implementation of the site create command as an example to follow.

internal/cmd/skupper/debug/kube/conn_sweeper.go
```diff
@@ -0,0 +1,192 @@
...
+	if len(failedPods) > 0 {
+		return fmt.Errorf("sweep failed on %d of %d router pod(s): %v", len(failedPods), len(podNames), failedPods)
+	}
+	return nil
+}
+
+func (cmd *CmdConnSweeper) WaitUntil() error { return nil }
+
+// findRouterPods returns every router pod whose router container is ready,
+// with HA there are multiple replicas and each holds its own connections.
```
nluaces: are these generated comments? if so, I do not think they are necessary. The reasoning behind is that they are not explaining things that are happening in the current code, that have some complexity.

internal/cmd/skupper/debug/kube/conn_sweeper.go
```diff
@@ -0,0 +1,192 @@
...
+
+// podExecer returns a sweeper.Execer that runs argv inside the router
+// container, where skmanage, python3 and the router's own network namespace
+// are all available.
+func (cmd *CmdConnSweeper) podExecer(podName string) sweeper.Execer {
+	return func(argv []string) ([]byte, error) {
+		// ExecCommandInContainer uses the deprecated context-less Stream, so
+		// it can't be cancelled directly; run it in a goroutine and give up
+		// on the sweep's behalf after podExecTimeout. The buffered channel
+		// lets the goroutine finish its send if Stream ever returns.
```
nluaces: I think this can be deleted as well. In any case, part of this explanation should be in the file where the ExecCommandInContainer function belongs.

nluaces:
Thank you for all your work! I left some comments regarding the kube implementation of the command. I suggest to fix also the same issues in the non-kube implementation.

I also suggest to review all the generated comments: if they explain things in the current code that are not easy to understand, they should be kept; if not, I would suggest to delete them.

Thanks!

internal/cmd/skupper/debug/debug.go
evanwang9x: Andy said to have this command hidden for now

internal/cmd/skupper/common/flags.go
```diff
@@ -252,6 +252,13 @@ type CommandVersionFlags struct {
 type CommandDebugFlags struct {
 }
 
+type CommandConnSweeperFlags struct {
```
ajssmith, line 255: The URL and skmanage should not be flags for the sub-command

Suggest following the pattern that the dump subcommand uses for getting skstat as it accounts for the three scenarios

- k8s 
- container (e.g.)
- systemd

We can assume skmanage is present in the router container and for systemd it can be assumed user already had to install the skupper-router binary

internal/cmd/skupper/debug/sweeper/gather.go
```diff
@@ -0,0 +1,175 @@
...
+
+func isTCPAdaptorConn(c connInfo) bool {
+	return c.Container == tcpContainer && c.Host != egressDispatch
+}
+
+// gatherSockets reads kernel socket state, preferring `ss -tin` and falling
+// back to the python netlink script when ss isn't available (e.g. inside the
+// router container, which ships python3 but not iproute).
+func gatherSockets(execFn Execer) (byPeer, byLocal map[string]socketInfo) {
+	if out, err := execFn([]string{"ss", "-tin"}); err == nil {
```
ajssmith: iproute is lightweight and has some other useful utilities...we should look to add to the router container

c-kruse: @ajssmith do i remember us somewhat recently stripping out non essential system dependencies to minimize churn from security scanners? Could be worth raising with the group to decide where our priorities are there.

EDIT: I do see that we've got a python script embedded to work around that - not ideal. I think I'd prefer anything that exercises router system dependencies (python or iproute included) live with the router codebase, get tested there, and bear some obvious `sk-part-of-our-public-api` name so that we don't accidentally break this debug option down the line.

internal/cmd/skupper/debug/sweeper/sweeper.go
```diff
@@ -0,0 +1,95 @@
...
+	DefaultURL           = "amqp://127.0.0.1:5672"
+	DefaultIdleThreshold = 4 * 3600
+	DefaultSkmanage      = "skmanage"
+)
+
+type Config struct {
+	URL               string
+	Skmanage          string
+	IdleThresholdSecs int
+	DryRun            bool
```
fgiorgetti: Should we use `dryRun`, which is false by default, causing the sweeper command to kill connections on execution.
Or change it with something like: `execute bool`, to prevent accidentally killling idle connections?

evanwang9x: I think execute bool is more intuitive than dryrun so I'll change it to be that.

evanwang9x:
@coderabbitai full review

pwright:
@coderabbitai explain the difference between how the primary and fallback logic helps wrt sweeping half closed connections

pwright:
@evanwang9x should we consider dropping the fallback? see https://github.com/skupperproject/skupper/pull/2530#issuecomment-5194132291

evanwang9x:
> @evanwang9x should we consider dropping the fallback? see [#2530 (comment)](https://github.com/skupperproject/skupper/pull/2530#issuecomment-5194132291)

Hey Paul, I think right now the better move is to fix it rather than drop it.
The router image doesn't ship `ss`, `skupper-router:3.5.0` has `skmanage` and
`python3`, but not the iproute2 package that provides `ss`. So on Kubernetes the
netlink fallback isn't really a fallback, it's the only path that works today.
Dropping it would make `debug sweep` non-functional until iproute2 lands in the
router image.

evanwang9x:
@coderabbitai I've widened the netlink state mask in `inetdiag.go`.

Re-check the half-closed question from your earlier comment: can the fallback
path now match and sweep a half-closed router connection, and does any
behavioural gap with the `ss -tin` primary path remain?

nluaces (approved):

internal/cmd/skupper/debug/sweeper/gather.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil, nil, fmt.Errorf("could not read socket state, so no connection could be matched to its socket: ss unavailable (%v) and python3 fallback failed (%v)", ssErr, pyErr)
+	}
+	byPeer, byLocal := socketsFromDiagOutput(out)
+	return byPeer, byLocal, nil
+}
+
+// socketsFromSS builds two {lastrcv, lastsnd} maps — one keyed by peer
+// address, one by local address — by pairing each socket's header line in
+// `ss -tin` output with its following detail line.
+func socketsFromSS(out []byte) (byPeer, byLocal map[string]socketInfo) {
```
fgiorgetti, line 157: It is important to add a unit test to validate `socketsFromSS`.

internal/cmd/skupper/debug/sweeper/inetdiag.go
```diff
@@ -0,0 +1,95 @@
...
+                lastsnd = struct.unpack_from("=I", data, aoff + 4 + 44)[0]
+                lastrcv = struct.unpack_from("=I", data, aoff + 4 + 52)[0]
+            aoff += (alen + 3) & ~3
+        print("%s:%d %s:%d %d %d" % (src, sport, dst, dport, lastrcv, lastsnd))
+        off += (ln + 3) & ~3
+`
+
+// socketsFromDiagOutput parses inetDiagScript's output into the same two
+// maps socketsFromSS produces.
+func socketsFromDiagOutput(out []byte) (byPeer, byLocal map[string]socketInfo) {
```
fgiorgetti, line 77: Similarly, it would be nice to have a unit test to validate this one as well, so it'd be clearer on what is the expected input.

fgiorgetti (commented):
Minor suggestions only, but it seems to work properly.

fgiorgetti (approved):

PR #2534, opened 2026-07-14
Fixed truncated output in ExecCommandInContainer

### LINKED ISSUE #2533: ExecCommandInContainer silently truncates command output
`ExecCommandInContainer` in `internal/kube/client/exec_util.go` wraps its output buffer in a `bufio.Writer` that is never flushed. That means outputs under 4096 bytes is lost entirely and the caller gets an empty buffer and outputs over 4096 bytes is missing its final partial chunk.

evanwang9x:
Fixes #2533 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of command output from container executions by ensuring buffered output is flushed before results are returned.
  * Flush failures are now reported instead of being silently ignored.

### DIFF
--- a/internal/kube/client/exec_util.go
+++ b/internal/kube/client/exec_util.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"io"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


internal/kube/client/exec_util.go
```diff
@@ -26,10 +26,10 @@ func ExecCommandInContainer(command []string, podName string, containerName stri
 		targetContainer = containerName
 	}
 
-	var stdout io.Writer
-
 	buffer := bytes.Buffer{}
-	stdout = bufio.NewWriter(&buffer)
+	bufferedStdout := bufio.NewWriter(&buffer)
+
+	var stdout io.Writer = bufferedStdout
```
c-kruse: *bufio.Writer satisfies the io.Writer interface - no need to over complicate things with different vars.
suggest just `stdout := bufio.NewWriter(&buffer);` and stdout.Flush()

c-kruse (approved):
Nice catch here! Looks good to me

Issue #2536, opened 2026-07-16
Reject Site with both edge and HA enabled


AryanP123:
The Site CRD documents that edge sites cannot also have HA enabled, but this is not enforced. verifySiteSpec does not check for spec.edge && spec.ha, so a Site can be created with both set.

Reject this combination in verifySiteSpec (and optionally via CRD validation / CLI), with a clear status error, and add unit tests.

PR #2540, opened 2026-07-21
Added check so edge sites can't have HA enabled

evanwang9x:
Addresses #2536 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject site configurations where both Edge and high availability are enabled at the same time.
  * Improved configuration feedback with a clear error message naming the invalid combination.

* **Tests**
  * Expanded test coverage to verify acceptance of valid Edge-only and HA-only configurations.
  * Added a failing-case test for the invalid Edge+HA combination.


AryanP123:
Nice lgtm, also just noticed we can add the same edge+HA rejection in non-kube `SiteStateValidator.validateSite` . I think you could add that to this PR if that makes sense. @coderabbitai what do you think

c-kruse (approved):

PR #2541, opened 2026-07-22
Add Network, NetworkLink, InterNetworkIngress, NetworkAccess CRD types and manifests

fgiorgetti:
This is the initial PR (split from multi-van effort) that just introduces the new CRDs and a small addition to RouterAccessStatus. Here are the new custom resource types:

- Network: identifies a Skupper network by its NetworkId
- NetworkLink: establishes a link to another network via TLS hostname/port
- InterNetworkIngress: routes traffic from a NetworkLink or NetworkAccess to a local routing key
- NetworkAccess: exposes a local endpoint to another network with optional dynamic port allocation and TLS

Also adds:
- CRD YAML manifests for all four new types under config/crd/bases/
- Generated client, informer, and lister code for all four new types
- Updates config/crd/kustomization.yaml to reference the new manifests
- Updates pkg/apis/skupper/v2alpha1/register.go to register the new types
- Updates zz_generated.deepcopy.go with DeepCopy methods for new types
- RouterAccess no longer requires a port assigned to a role (dynamic port allocation)

## Summary by CodeRabbit

* **New Features**
  * Added new Kubernetes custom resources: `Network`, `NetworkLink`, `NetworkAccess`, and `InterNetworkIngress`, with validated `spec`/`status` schemas and improved `kubectl` columns.
  * Tightened `InterNetworkIngress` input validation to require `networkLink` and/or `networkAccess`.
* **Bug Fixes**
  * Enhanced `RouterAccess` by exposing allocated ports via `status.roles`, improving role port/endpoint tracking.
* **Documentation**
  * Updated CRD status condition descriptions and added an ADR for multi-VAN network resources.
* **Tests**
  * Added extensive unit tests for the new and updated behaviors.


config/crd/kustomization.yaml
```diff
@@ -7,9 +7,13 @@ resources:
 - bases/skupper_attached_connector_crd.yaml
 - bases/skupper_certificate_crd.yaml
 - bases/skupper_connector_crd.yaml
+- bases/skupper_inter_network_ingress_crd.yaml
 - bases/skupper_link_crd.yaml
 - bases/skupper_listener_crd.yaml
 - bases/skupper_multikeylistener_crd.yaml
+- bases/skupper_network_crd.yaml
+- bases/skupper_network_link_crd.yaml
+- bases/skupper_network_access_crd.yaml
```
c-kruse: Waiting to see your ADR, I'm sure you'll have considered this. I had imagined we'd have the net-new resources not installed by default - enabled/installed in some alternate controller configuration so that "experimental" features require opt in until we feel ready to commit to them.

fgiorgetti: The current implementation (which should come next, once we settle the CRD discussion) does not yet handle the new CRDs as optionals, but I agree with you, that they must be considered optionals.
Along with the new ones, there is also a change to RouterAccess (which should not affect exisintg installations). But that is a great point.

doc/adr/0011-multi-van-resources.md
```diff
@@ -0,0 +1,68 @@
...
+
+Date: 2026-07-28
+
+## Status
+
+Proposed
+
+## Context
+
+New CRDs have been added to support multi-van management through VMS.
```
ajssmith, line 11: We should review the decision to add new CRDs versus extending the existing ones for multi-van support.

I recall that RBAC for the resources was a consideration but we should revisit.

Were there other factors to support adding new CRDs?

nluaces: I'm thinking that depending on the type of modification of the CRD, it could provoke an incompatibility with existent custom resources that were created with old CRDs (ie fields reconsidered as required). Adding a new CRD instead of modifying the current ones gives more capacity for change without having deal with the migration of older CRs.

fgiorgetti: @ajssmith NetworkAccess could be accommodated as part of RouterAccess, as RouterAccess is going to be used for "normal" roles as part of the multi-van work. The main reason why it is defined as a separate resource now, is RBAC.

For NetworkLinks, though, I believe it should be kept separate from Links, because the purpose is different. While Links are used to form a VAN (let's say it as sites that will be able to access all available routing keys on the connected VAN), a NetworkLink has the purpose of connecting distinct VANs, whereas only specific routingKeys will be made available to the other.

Network should be its own IMO. The motivation for it is written as part of the Decision section.

InterNetworkIngress is also something that I see as a new Resource, as we could eventually have multiple of them to expose the same routingKey to different VANs.

fgiorgetti: > I'm thinking that depending on the type of modification of the CRD, it could provoke an incompatibility with existent custom resources that were created with old CRDs (ie fields reconsidered as required). Adding a new CRD instead of modifying the current ones gives more capacity for change without having deal with the migration of older CRs.

@nluaces this is also something to be careful about. Right now, the change we have here for RouterAccess, is not supposed to break existing resources as a new field was added as part of the Status field. But if we modify something to accommodate new needs, it is something we have to keep in mind.

pkg/apis/skupper/v2alpha1/types.go
```diff
@@ -1104,3 +1166,299 @@ type AttachedConnectorBindingSpec struct {
...
+// InterNetworkIngressSpec
+// +kubebuilder:validation:XValidation:rule="(has(self.networkLink) && self.networkLink != ”) || (has(self.networkAccess) && self.networkAccess != ”)",message="At least one of 'networkLink' or 'networkAccess' must be provided and non-empty."
+type InterNetworkIngressSpec struct {
+	RoutingKey string `json:"routingKey"`
+	// +optional
+	NetworkLink string `json:"networkLink,omitempty"`
+	// +optional
+	NetworkAccess string            `json:"networkAccess,omitempty"`
+	Settings      map[string]string `json:"settings,omitempty"`
+}
```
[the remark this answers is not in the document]
fgiorgetti: Updated version seems to work fine here, but I can update it to satisfy coderabbitai.

config/crd/kustomization.yaml
fgiorgetti: @c-kruse I have added a new chart skupper-crds, which installs only the default crds unless you specify `--set multiVan=true`. The Makefile target generate-crd-manifest now generates a new yaml called: `skupper-multi-van-crds.yaml`.
Let me know if you agree with this updated mechanism.

charts/skupper-crds/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+```yaml
+base: true
+multiVan: true
+```
+
+> **Note:** Once CRDs are installed, removing them (by running `helm upgrade`
+> with a flag set to `false`) will **not** delete the CRDs from the cluster.
+> Helm never deletes CRDs during an upgrade or uninstall. Remove them manually
+> with `kubectl delete crd` if needed.
```
c-kruse: Is this true?

IIRC in my now abandoned earlier attempt at this I think  I needed a special annotation to tell helm to not delete CRDs under templates. (I know this is true for /crds.)

See https://github.com/skupperproject/skupper/pull/2376/changes#diff-7381ef09014020f008d032fd1f1bbb5573c8f14f6ebcabe6e00b0ed2e15c2ea5R6

doc/adr/0011-multi-van-resources.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Update to RouterAccess
+
+The `.status.roles[]` field has been introduced so that the controller can persist the allocated port, in case a `.spec.roles[]` entry is defined without a port (set to 0).
+
+When a role is defined without a port, the controller should allocate a port and store it into the `.status.roles[].port`.
+
+### Make them optional
+
+The new CRDs as well as the change to the RouterAccess CRD must be handled as optional by the Controller.
```
c-kruse, line 59: @fgiorgetti this stands out to me as a potentially tricky change. Have you proofed this out yet? Primarily concerned about how the controller will test the RouterAccess status addition.

fgiorgetti: @c-kruse In Skupper V1 we had a similar behavior with Skupper Policies.

We might need something similar here that runs at the controller startup and checks if the multi-van CRDs are available and inspect the routeraccesses CRD using `crdClient.ApiextensionsV1().CustomResourceDefinitions()`.

I would say, we can make it, but to keep it simple, I believe the best would be to run this check only at the controller startup, therefore if CRDs gets installed later, we will need a controller restart.

charts/skupper-crds/README.md
fgiorgetti: This is wrong. Possibly copy/paste issue. I will fix it.

fgiorgetti: @c-kruse please, check it out one more time.

charts/skupper-crds/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: skupper-crds
+description: Helm chart for installing and upgrading Skupper CRDs.
+version: 0.1.0-dev
```
nluaces, line 4: not a blocker: what is happening each release with the other helm charts is that the version matches the appVersion.

charts/skupper-crds/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+```bash
+helm install skupper-crds oci://quay.io/skupper/helm/skupper-crds \
+    --set base=false \
+    --set multiVan=true
+```
+
+### Multi-VAN CRDs
+
+The chart includes 4 additional CRDs that enable multi-VAN (Virtual Application
```
nluaces, line 43: i would not recommend to make reference to the number of additional CRDs and their names, seems like this section is going to get obsolete after changes or the creation of new crds, how about make reference to the actual folder instead?

nluaces (approved):

fgiorgetti:
Closing this one, in favor of #2564 which introduces support to the same features without new CRDs.
Users will still need to update their CRDs to use the new features, but not doing so, won't impact anything.

Issue #2542, opened 2026-07-23
Listener / MultiKeyListener report Ready when Service create fails


AryanP123:
If `spec.host` is not a valid Kubernetes Service name, the controller fails to create the Service but still sets the resource to Ready (`Configured/Operational` True, message `OK`). The error only appears in controller logs.

Affects both `Listener` and `MultiKeyListener` (extended_bindings.go expose path).

To reproduce:

1. Site Ready in a namespace.
2. Apply an invalid host:
```
apiVersion: skupper.io/v2alpha1
kind: MultiKeyListener
metadata:
  name: backend-tcp
spec:
  host: "{name}"
  port: 8080
  strategy:
    weighted:
      routingKeys:
        east-backend: 80
```
(or a Listener with `host: "{badhost}"`)

3. Check status and Services.

Expected: Not Ready
Actual: Ready / OK

PR #2544, opened 2026-07-27
Changed code to report invalid host

evanwang9x:
Addresses #2542 

Changed code to return an error instead of smoothing over it and having it only appearing in the control log

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when listeners are created, updated, deleted, or exposed.
  * Invalid listener hostnames are now rejected with clear status errors.
  * Prevented services from being created for invalid host values.
  * Initialization and configuration failures are now surfaced instead of being silently ignored.

* **Tests**
  * Added coverage for invalid hosts on standard and multi-key listeners.

### DIFF
--- a/internal/kube/site/extended_bindings.go
+++ b/internal/kube/site/extended_bindings.go
@@ -45,7 +45,7 @@ func NewExtendedBindings(controller *watchers.EventProcessor, profilePath string
 	return eb
 }
 
-func (a *ExtendedBindings) init(context BindingContext, config *qdr.RouterConfig) {
+func (a *ExtendedBindings) init(context BindingContext, config *qdr.RouterConfig) error {
 	a.context = context
 	if a.mapping == nil {
 		a.mapping = qdr.RecoverPortMapping(config)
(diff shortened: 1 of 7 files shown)


fgiorgetti (commented):
It works great with both Listeners and MultiKeyListeners.
My only suggestion is to add the same host validation at the CLI level, when running `skupper listener create` or `skupper listener generate`.

fgiorgetti:
> It works great with both Listeners and MultiKeyListeners. My only suggestion is to add the same host validation at the CLI level, when running `skupper listener create` or `skupper listener generate`.

But one thing to be cautions about it. On system sites (nonkube) the host can represent either a FQDN or an IP Address. So if we add such validation, it must be done exclusively to kubernetes sites. @nluaces @c-kruse thoughts?

nluaces:
> But one thing to be cautions about it. On system sites (nonkube) the host can represent either a FQDN or an IP Address. So if we add such validation, it must be done exclusively to kubernetes sites. 

It makes sense, currently the listener commands for create and generate in kubernetes platforms don't validate the hostname. 

But there is already a host validation for the non-kube commands listener create/generate: https://github.com/skupperproject/skupper/blob/8fcff848d2ee00b96ef79a5856b55544549f431f/internal/cmd/skupper/listener/nonkube/listener_create.go#L108

hostname: `hostStringValidator.Evaluate(cmd.Flags.Host)`
IP: `net.ParseIP(cmd.Flags.Host)`

(for the listener generate command the host validation is identical)

c-kruse (commented):
@evanwang9x want to give rebasing a try? Looks like we're on each other's toes. Doesn't look too bad, though.

For ExtendedBindings.init it is important that `exposed` and `selectors` don't get stomped on subsequent init calls. I think this should be fine:
```
	if a.exposed == nil {
		a.exposed = ExposedPorts{}
	}
	if a.selectors == nil {
		a.selectors = map[string]TargetSelection{}
	}
	err := a.bindings.SetBindingEventHandler(a)
```

For Site.setBindingsConfiguredStatus it looks to me like we've got yet another error to jam into that `stderrors.Join`. Now that I look at it, that's probably a bug in one of our implementations - I included the passed `err` and you ignore it (it was unused before my change.) I think we want to include it.

evanwang9x:
Thanks Christian, I rebased it with your suggestion.

PR #2545, opened 2026-07-28
Add optional includePods for label-template ConfigMaps

### LINKED ISSUE #2480: Add an optional way to pass additional labels and annotations to pods as well
Today when a configmap is labeled with `skupper.io/label-template: ""`, the controller will get all of its labels and annotations and pass them along to its owned resources, like, Deployment, Services, ConfigMaps.

But the Deployment's pod(s) do not get annotated/labeled (possibly to avoid a router restart).

We could evaluate an optional way to tell the Controller to pass the labels/annotations to the Deployment's pod(s) as well.

As an example, an extra label could be added to signal that pods should also be annotated? Or eventually depending on the value set to `skupper.io/label-template`?

Not sure what the best approach is, but we could experiment something.

AryanP123:
Fixes #2480

### DIFF
--- a/internal/kube/site/labels/registry.go
+++ b/internal/kube/site/labels/registry.go
@@ -44,11 +44,11 @@ func (l *LabelsAndAnnotations) Update(key string, cm *corev1.ConfigMap) error {
 func (l *LabelsAndAnnotations) SetLabels(namespace string, name string, kind string, labels map[string]string) bool {
 	desired := map[string]string{}
 	if registry, ok := l.namespaces[namespace]; ok {
-		registry.setLabels(name, kind, labels, desired)
+		registry.setLabels(name, kind, labels, desired, false)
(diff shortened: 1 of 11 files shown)


fgiorgetti (approved):
It looks good to me.

internal/kube/site/labels/registry.go
```diff
@@ -192,7 +203,7 @@ func (r *Registry) filter(name string, kind string, target map[string]string, la
 				if isExcluded(k, excludes) {
 					continue
 				}
-				if v2, ok := labels[k]; !ok || v != v2 {
+				if v2, ok := annotations[k]; !ok || v != v2 {
```
c-kruse, line 206: nice catch!

c-kruse (approved):
LGTM!

PR #2558, opened 2026-08-13
Second attempt at two listener integration test.

mgoulish:
## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for sites with multiple listeners.
  * Verifies that each listener creates the correct Service and router configuration, including ports and listener labels.

### DIFF
--- a/tests/integration/kube/controller/listener_test.go
+++ b/tests/integration/kube/controller/listener_test.go
@@ -95,3 +95,67 @@ func TestListenerWithoutSite(t *testing.T) {
 		actual.Status.Status,
 	)
 }
+
+func TestTwoListeners(t *testing.T) {
+
+	// In this test we wait for the Services corresponding to both Listeners to show up,
+	// and also separately wait for the names of both listeners to show up in the Router configmap.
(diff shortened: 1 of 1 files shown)


nluaces:
I think the test is good; which scenario/use case is addressing? 
I was wandering why is necessary to test the creation of two listeners in particular. 
Could be okay to just create one?

mgoulish:
I don't really have a use-case in mind -- I just thought it might be nice to make sure it works with more than one. But I would be very happy to cut it down to one!
I will do that.

mgoulish:
Oh wait.

The reason for this test was that we already had a good test for a single Listener. 
It's called TestSiteWithListener.

This new test just makes sure that everything basically works right if you try it with multiple Listeners.
Make sure they both show up instead of, for example, just the last one.
Make sure they have the right labels.

If that does not seem useful enough I am happy to withdraw it, but there is no point in changing it to a single Listener while TestSiteWithListener already exists.

nluaces (approved):

fgiorgetti (approved):

PR #2561, opened 2026-08-19
Create and delete a Listener many times rapidly.

mgoulish:
Create and delete the same Listener a zillion times quickly, ending up with a creation .. and make sure nothing breaks, and we end up with a Listener, the corresponding Service, and a reference to it in the Router Config.

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for repeatedly creating and deleting a Listener.
  * Verifies the final Listener, its associated Service, and Router configuration are created correctly and remain available.

### DIFF
--- a/tests/integration/kube/controller/listener_test.go
+++ b/tests/integration/kube/controller/listener_test.go
@@ -159,3 +159,69 @@ func TestTwoListeners(t *testing.T) {
 	assert.Equal(t, svcB.Spec.Ports[0].Port, int32(9090))
 	assert.Equal(t, svcB.Labels["internal.skupper.io/listener"], "listener-b")
 }
+
+func TestListenerCreateDeleteStorm(t *testing.T) {
+	// What will happen if we rapidly create and then delete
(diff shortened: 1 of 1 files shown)


mgoulish:
I don't agree with one of CodeRabbit's comments.  It wanted me to make sure everything stabilized after every non-final delete, before doing the next create.

But that is not what I think we should be testing here.

I don't want this test to be polite, I want it to be rude and disruptive.

That's the way to see if we can get something to break.

So I have implemented the 'storm' loop with no error checking.  It just creates and deletes the Listener 100 times without checking anything .... and then we see what we have at the end.

fgiorgetti (approved):

nluaces:
> I don't agree with one of CodeRabbit's comments.

You can argue with CodeRabbit in the thread of its comment. It might learn preferences for this project. 

And now for something completely different: Codespell is failing for a misspelled word unrelated with this pull request (I wonder why this was not indicated in the corresponding pull request).

nluaces (approved):

PR #2563, opened 2026-08-28
Identical Listeners test

mgoulish:
This test takes a while -- every time I have run it, it has taken very close to 27 seconds.

Create two Listeners on a Site, identical except for their names.

The first will become active -- it will own the Service, and its name will show up in the Router Config.

The second one will be rejected with an error.

But then if you kill the first one, after some time the second one will get adopted.
*It* will own the Service, and its name will show up in the Router Config.

## Summary by CodeRabbit

- **Tests**
  - Added integration coverage for identical listener failover regardless of listener ordering.
  - Verified duplicate listeners are rejected while the active listener owns the shared service and router entry.
  - Confirmed the standby listener takes over after the active listener is removed and its duplicate error clears.
  - Validated service ownership, router configuration, and listener readiness after takeover.
  - Improved documentation for related listener lifecycle and stress tests.
- **Documentation**
  - Corrected a certificate-related comment.

### DIFF
--- a/tests/integration/kube/controller/listener_test.go
+++ b/tests/integration/kube/controller/listener_test.go
@@ -237,6 +237,9 @@ func TestListenerCreateDeleteStorm(t *testing.T) {
 	assert.Assert(t, strings.Contains(routerConfig.Data[types.TransportConfigFile], "listener/"+listenerName))
 }
 
+// TestIdenticalListenerFailover checks that when two Listeners are identical
+// except for their names, only the first one is used, but if the first one
+// then goes away, the second one takes over.
 func TestIdenticalListenerFailover(t *testing.T) {
 	// Make two listeners that are identical except for their names.
 	// The first one ought to get the host and port, while the second
--- a/tests/integration/kube/controller/listener_test.go
+++ b/tests/integration/kube/controller/listener_test.go
@@ -320,8 +320,11 @@ func TestIdenticalListenerFailover(t *testing.T) {
 		if done, err := retryOnNotFound(err); !done {
 			return false, err
 		}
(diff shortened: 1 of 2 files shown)


tests/integration/kube/controller/listener_test.go
```diff
@@ -235,3 +235,144 @@ func TestListenerCreateDeleteStorm(t *testing.T) {
...
+
+	_, err = tc.clients.GetSkupperClient().SkupperV2alpha1().Listeners(namespace).Create(ctx, a, metav1.CreateOptions{})
+	assert.NilError(t, err)
+	_, err = tc.clients.GetSkupperClient().SkupperV2alpha1().Listeners(namespace).Create(ctx, b, metav1.CreateOptions{})
+	assert.NilError(t, err)
+	t.Log("created listener-a and listener-b")
+
+	// Wait until 'B' is rejected because it's a duplicate
+	var lb *skupperv2alpha1.Listener
+	waitFor(t, 30*time.Second, 250*time.Millisecond, func() (bool, error) {
```
ajssmith, line 269: Kubernetes does not necessarily guarantee that resources get "created" in the order submitted. While it is likely that the second listener will be rejected the test would be improved if it anticipates either listener can claim being the initial owner.

mgoulish: Thanks!   I will remember this.

I think I have taken care of it now by just checking to see which listener comes out on top.

ajssmith (approved):

c-kruse (approved):

nluaces (approved):

PR #2567, opened 2026-09-04
Fix listener matched status using aggregated network addresses

### LINKED ISSUE #2428: Listener status shows "No matching connectors" / Pending when remote connector exists across inter-site link
**Describe the bug**
A Listener reports Pending / No matching connectors even when a matching Connector exists on a remote site and the data plane is fully functional. Traffic routes correctly through the link, but the Listener's Matched condition remains False.

**How To Reproduce**
1. Create two sites: a Kubernetes site and a Docker site
2. On the Docker site, create a Connector with routing key my-backend
3. On the k8s site, create a Listener with routing key my-backend
4. Create a link from the k8s site to the Docker site
5. Confirm the link is up: kubectl get link -n <namespace>
6. Check listener status: kubectl get listener -n <namespace>

NAME         ROUTING KEY   PORT   HOST         STATUS    HAS MATCHING CONNECTOR   MESSAGE
my-backend   my-backend    8080   my-backend   Pending                            No matching connectors

**Expected behavior**
The Listener should report Ready with HAS MATCHING CONNECTOR: true, because a Connector with the matching routing key exists on the remote (Docker) site and traffic is routing correctly through the link.

To confirm the data plane is working despite the Pending status:
I ran curl on listener side and it can return the response from connector

curl http://<listener-clusterip>:8080/api/hello
# returns response served by the container on the Docker site

**Environment details**
- Skupper version: 2.1.3 (from siteVersion in skupper-network-status)
- Platform: Kubernetes (listener site) + Docker (connector site)
- Link role: inter-router, direction k8s → Docker

**Additional context**
The skupper-network-status ConfigMap on the k8s site correctly aggregates the remote connector:

"addresses": [{"name":"my-backend","listenerCount":1,"connectorCount":1}]
However, the Docker site does not appear in siteStatus — only the k8s site is present. The kube-adaptor's Matched reconciliation appears to check siteStatus entries (which only includes sites with full vanflow records surfaced locally) rather than the aggregated addresses record, which correctly reflects connectorCount: 1.

The kube-adaptor log confirms the timing: the Listener flipped to Pending at exactly the moment the local (stale) Connector's vanflow records were purged — the reconciler found no local connector and did not fall back to checking the remote one via the address aggregation.

AryanP123:
Fixes #2428

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network status handling so binding status can recognize matching connectors and listeners from network address information.
  * Binding and listener status updates now remain accurate when matching entries are represented through address data rather than local records.

### DIFF
--- a/internal/kube/controller/controller.go
+++ b/internal/kube/controller/controller.go
@@ -688,7 +688,7 @@ func (c *Controller) networkStatusUpdate(key string, cm *corev1.ConfigMap) error
 		return nil
 	}
 	c.log.Debug("Updating network status", slog.String("site", key))
-	return c.getSite(cm.ObjectMeta.Namespace).NetworkStatusUpdated(network.ExtractSiteRecords(status))
+	return c.getSite(cm.ObjectMeta.Namespace).NetworkStatusUpdated(network.ExtractSiteRecords(status), status.Addresses)
 }
 
 func filter[V any](controller *Controller, handler func(string, V) error) func(string, V) error {
--- a/internal/kube/site/binding_status.go
+++ b/internal/kube/site/binding_status.go
@@ -6,22 +6,25 @@ import (
 	"strings"
 
 	internalclient "github.com/skupperproject/skupper/internal/kube/client"
+	"github.com/skupperproject/skupper/internal/network"
 	skupperv2alpha1 "github.com/skupperproject/skupper/pkg/apis/skupper/v2alpha1"
 )
 
 type BindingStatus struct {
 	connectors map[string][]string
 	listeners  map[string][]string
+	addresses  []network.AddressInfo
 	client     internalclient.Clients
 	errors     []string
 	logger     *slog.Logger
 }
 
-func newBindingStatus(client internalclient.Clients, network []skupperv2alpha1.SiteRecord) *BindingStatus {
+func newBindingStatus(client internalclient.Clients, network []skupperv2alpha1.SiteRecord, addresses []network.AddressInfo) *BindingStatus {
 	s := &BindingStatus{
 		client:     client,
 		connectors: map[string][]string{},
 		listeners:  map[string][]string{},
+		addresses:  addresses,
 		logger: slog.New(slog.Default().Handler()).With(
 			slog.String("component", "kube.site.binding_status"),
 		),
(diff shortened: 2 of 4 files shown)


c-kruse:
@AryanP123 Can you explain what's going on here? I'm not sure I can make much sense of the changes here or the reporter's diagnosis. the status addresses and siteStatuses comes from the [same source](https://github.com/skupperproject/skupper/blob/2.1.3/internal/flow/status.go#L186C38-L186C55): if the connector-site's records were purged, the connector _should_ have gone with it.

My best wild guess as to what's actually going on here, based on having flirted with changing this several times, is that this is a problem with the purge logic. Ex: network is up and the router at the docker site `rD` and the kube site are happily connected. A connector gets added to site D's configuration, so router `rD` is killed and replaced with `rD'`. At the kube site the collector sees `rD'` announce itself, and maybe it sees its CONNECTOR and SITE records with it. It also notices `rD` has gone away, so it purges `rD`'s records, including the SITE that aught to now belong to `rD'`, but still belongs to `rD`.

AryanP123:
@c-kruse Yeah, fair point. `siteStatus` and `addresses` are built in the same pass in `StatusSync.build()`, so if remote site records get purged, the connector should disappear from both. Your rD -> rD' theory makes more sense.

This PR doesn't fix purge logic. It just makes kube `BindingStatus` check aggregated `addresses` too, like non-kube already does. The issue was that `addresses` showed `connectorCount: 1` but the Listener still said "No matching connectors" because kube only looked at site records.

I couldn't repro it naturally with kind+podman, but I could trigger the same wrong status by editing the network status configmap to match what the reporter saw.

Happy to look at the purge path instead if you'd rather fix it there. Lmk what you think

c-kruse:
@AryanP123 I think we should focus on the purge path instead: it _should_ fix "both" bugs implicitly reported here (missing Site from network status, and inconsistent listener/connector status), and potentially similar network-observer bugs we haven't found yet with the same pattern.

All of this skupper-network-status + statusSync stuff is something we plan to deprecate in 2.3 and remove soon after (2.4?), if we focus on the purge/sources path instead we at least get one long-term fix (in the network-observer.)

Added https://github.com/skupperproject/skupper/issues/2568 with some more context from my previous attempts.

PR #2572, opened 2026-09-10
Track multiple sources per vanflow record during purge.

### LINKED ISSUE #2568: vanflow purge source affinity
`pkg/vanflow/store` attributes each record to the *first* `SourceRef` that produced it (`Metadata.Source`). When a source stops heartbeating and is forgotten, `StatusSync.purge` (`internal/flow/status.go`) and `Collector.purge` (`cmd/network-observer/internal/collector/collector.go`) delete everything indexed to it. This is incorrect when multiple sources assert a record with the same ID.

See example bug from https://github.com/skupperproject/skupper/issues/2428. System site routers emit their own SITE record with a stable identity `SKUPPER_SITE_ID` but get a fresh random `router_id` on every process start. After a system router restarts, the SITE record is re-asserted by the new source but stays attributed to the old one and is purged with it, orphaning the new router's ROUTER/CONNECTOR records. The site disappears from `siteStatus` while `addresses[]` still counts the new connectors.

## Suggestions

Last-source-wins is probably also not ideal: the old router could re-assert the records (on FLUSH) at any time and stomp the new router's source assertion.

Change the store so a record tracks the *set* of sources that have asserted it, `Metadata.Sources []SourceRef`. Update the Purge logic to pop the SourceRef from Sources and to only delete records with no other sources.

## Reproducing
- Link at least 1 kube, and 1 system site. Make sure the system site is **not** --reload-type auto.
- Change the system site: add a listener/connector. See the router container restart (with new configuration.)
- Wait at least **30s**, or for the kube site's kube-adaptor controller to log the forgotten source.
- check network status for the system site.

AryanP123:
Fixes #2568

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source cleanup so removing one source no longer deletes records still associated with other sources.
  * Records are now removed only when no associated sources remain.
  * Improved handling of records shared by multiple sources, preserving contributions from each source.
  * Source removal and record updates now consistently trigger the appropriate change or deletion notifications.
  * Source cleanup now works reliably across different store configurations.
  * Process reconciliation now correctly recognizes records associated with multiple sources.


pkg/vanflow/store/store.go
```diff
@@ -16,7 +16,60 @@ type Entry struct {
 type Metadata struct {
 	LastUpdate time.Time
 
+	// Source is the first source that asserted this record.
```
c-kruse: Looks like the only place this is used (outside of store) is the network observer process mapping stuff. I think it'd be better to stop carrying around a Source field and to add a `HasSource(source SourceRef) bool` method instead - the process controller thing will work fine that way.

pkg/vanflow/store/store.go
```diff
@@ -16,7 +16,60 @@ type Entry struct {
...
+		m.Source = source
+	}
+	return true
+}
+
+func (m *Metadata) RemoveSource(source SourceRef) bool {
+	m.ensureSources()
+	for i, existing := range m.Sources {
+		if sourceRefEqual(existing, source) {
+			m.Sources = append(m.Sources[:i], m.Sources[i+1:]...)
```
c-kruse: I think we need to be careful to avoid race conditions/consistency problems here and copy the slice instead of mutating/shifting its contents. (maybe in AddSource too?) Store methods return Entry/Metadata by value on purpose so that the caller's copy doesn't get changed out from under it. Now with a slice (instead of a string) and the shared backing array it points to, mutating that slice's contents might break guarantees.

Maybe something like this instead?
```
	i := slices.Index(m.Sources, source)
	if i < 0 {
		return false
	}
	m.Sources = slices.Delete(slices.Clone(m.Sources), i, i+1)
   ...
```

PR #2574, opened 2026-09-14
fix dependency alerts

nluaces:
## Summary by CodeRabbit

* **Chores**
  * Updated underlying software components and supporting libraries.
  * Removed several no-longer-used supporting components.

* **Improvements**
  * Improved API response handling and content-type reporting.
  * Added validation support for supported API enum values.
  * Improved handling of optional router link protocol data for more consistent requests and responses.


c-kruse (commented):
oapi-codegen is what we use to generate the network-observer API server boilerplate. Suggest we rerun the generation code and commit that as well: `go generate ./...`.

Also suggest `go mod tidy`.

fgiorgetti (approved):

c-kruse (approved):
Tested out the regenerated observer API and it looks good! Thank you

PR #2575, opened 2026-09-15
Add skupper debug cert subcommand to inspect X.509 certificates

### LINKED ISSUE #2569: Add debug sub-command to view certs
Add a debug sub-command to inspect (e.g. decode) an X.509 certificate in use by Skupper.

This would help in resolution of support requests and potentially used as part of must-gather operations.

The sub-command should render the key fields to check (subject, issuer, validity period, SAN, public key info)

AryanP123:
Fixes #2569

## Summary by CodeRabbit

* **New Features**
  * Added the `debug cert-inspect` command for inspecting certificates.
  * View certificates from Kubernetes resources, local files, or configured certificate directories.
  * Inspect DNS names, IP addresses, email addresses, and URI subject alternative names.
  * Display certificate details in table, detailed, or structured output formats.
  * Select a certificate by name or inspect all available certificates.
  * Identify certificates designated for signing.
  * Added validation for certificate names, file paths, and output formats.


internal/cmd/skupper/debug/kube/cert.go
```diff
@@ -0,0 +1,157 @@
...
+	if cmd.Flags != nil && cmd.Flags.File != "" {
+		info, err := certdisplay.ParseCertificateFile(cmd.Flags.File, cmd.Flags.File)
+		if err != nil {
+			return fmt.Errorf("failed to parse certificate from %s: %w", cmd.Flags.File, err)
+		}
+		return certdisplay.Display([]certdisplay.Info{*info}, cmd.output, true)
+	}
+
+	if cmd.certName != "" {
+		certificate, err := cmd.Client.Certificates(cmd.Namespace).Get(context.TODO(), cmd.certName, metav1.GetOptions{})
```
fgiorgetti: Should we return an error here, requiring a "Certificate" resource to exist, or should we try to read a Secret with the given name?
The reason why I am asking this, is to validate a Secret resource that might be used by a Link resource, which won't have a corresponding Certificate resource.

internal/cmd/skupper/debug/nonkube/cert.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (cmd *CmdDebugCert) collectCerts() ([]certdisplay.Info, error) {
+	var infos []certdisplay.Info
+	seen := map[string]bool{}
+
+	certPaths := []struct {
+		basePath api.InternalPath
+		prefix   string
+	}{
+		{api.CertificatesPath, ""},
```
fgiorgetti: Also include: `api.IssuersPath`, `api.InputIssuersPath`.
Along with that, I have a similar comment to add.
System site users can also provide custom certificates through Secrets.
Maybe we should also include Secrets provided through: `api.InputSiteStatePath`.

ajssmith:
Suggest sub-command name convey the scope e.g. "cert-inspect" or similar

fgiorgetti (commented):
Along with my other comments, I have two other suggestions that could be useful here:

1. Flag to export the public certificate as a file

Something like `--dump-certificate=tls.crt`.

2. Flag to validate-cert-authority

If the provided certificate is a CA we could also verify if it has authority over the provided `--validate-cert-authority=tls.crt`.

@ajssmith @AryanP123 let me know your thoughts on this. And if you guys believe it is useful, feel free to use better names for the suggested flags.

internal/cmd/skupper/debug/nonkube/cert.go
AryanP123: Thanks @fgiorgetti makes sense.

For kube, I'll try the Certificate CR first, then fall back to a TLS Secret with the same name, and include standalone TLS secrets in list output.

For non-kube, I'll add runtime/issuers, input/issuers, and TLS secrets from input/resources.

Happy to do --dump-certificate / --validate-cert-authority in a follow-up unless you want them here in this PR.

fgiorgetti: @AryanP123 it can be done separately for sure, thanks!

pwright:
Given that some users bring their own certs, does it make sense to also report 'signing capability', or as LLM says:

Since `CertificateSpec.Signing` is already available on the kube-side CR (used to distinguish CA-generation vs leaf-signing in `mgr.go`'s `generateSecret`: mgr.go:363-371 ), consider surfacing it as an explicit `Is Signing Cert` field in `cert-inspect` output for kube, distinct from the x509-derived `IsCA`. For nonkube, note that this would currently need to be inferred from directory (`issuers/` vs `certs/`) since no `Certificate` spec is loaded there.

nluaces (approved):

PR #2580, opened 2026-09-21
Separately store the latest expose error for each Listener, so their …

mgoulish:
…errors can't overwrite each other. Return the error from UpdateListener so CheckListener can see it. Also add a unit test for this small new functionality.

I want to use this tiny new functionality in an integration test that I have already finished, but I thought it might be best to first do a separate PR for just this part.

## Summary by CodeRabbit

* **Bug Fixes**
  * Listener exposure failures are now correctly reported when updating listeners.
  * Exposure errors are tracked independently for each listener, preventing one listener’s failure from affecting another.
  * Errors are cleared after successful handling or retrieval.
  * Removing a listener also removes its associated exposure error.

* **Tests**
  * Added coverage for exposure error propagation and listener-specific error isolation.
  * Updated integration test configuration handling to improve validation of listener behavior.


mgoulish:
This is for issue #2581

internal/kube/site/extended_bindings_expose_test.go
```diff
@@ -0,0 +1,118 @@
...
+	// Updating Listener A should return A's error,
+	// and should *consume* A's error,
+	// while the error in B should not be changed.
+	_, gotA := eb.UpdateListener("listener-a", testListener("listener-a", "svc-a"))
+	assert.ErrorContains(t, gotA, "fail-a")
+	assert.ErrorContains(t, eb.lastListenerExposeErr["listener-b"], "fail-b")
+	_, stillA := eb.lastListenerExposeErr["listener-a"]
+	assert.Assert(t, !stillA, "listener-a error should have been consumed")
+
+	// Now updating Listener B should return-and-consume C's error.
```
AryanP123: typo: “return-and-consume C's error” but should say B's error?

